PR #136, opened 2026-06-13
feat(diag): alloc/syscall gate counters (std::diag)

rileyr:
## What

The test-time gate harness (#7 of the fast-protocol-I/O substrate plan, `notes/fast-protocol-io-substrate-2026-06-13.md`). The zero-alloc / one-syscall wins the rest of that plan targets regress **silently** without gates — this is the discipline that keeps them.

- `std::diag::heap_alloc_count() -> Int` — cumulative heap allocations (malloc / realloc / calloc / mmap).
- `std::diag::syscall_count(name: String) -> Int` — cumulative count of a wrapped I/O syscall: `"recv"`, `"recvmsg"`, `"read"`, `"write"`, `"send"`, `"sendto"`.

Read a counter before and after a steady-state region and assert the delta: "this loop did zero heap allocs", "exactly one read per poll". The runtime/test-time complement to compile-time `--warn-unbounded-alloc`.

Both return **-1** when the counting shim is absent (sanitizer builds, where TSan/ASan interceptors collide with `-Wl,--wrap`), so a caller distinguishes "gate unavailable in this build" from a real `0`.

## How

Reuses the existing `__wrap_malloc/realloc/calloc/mmap` shim (already linked into default builds for `LOTUS_ARENA_LOG_BIG_CHUNKS`):
- a relaxed-atomic counter added to each alloc wrapper;
- new `__wrap_{recv,recvmsg,read,write,send,sendto}` count-then-forward shims + the matching `-Wl,--wrap` link flags;
- `lotus_diag_{heap_alloc_count,syscall_count}` readers, exposed via a new `std::diag` dispatch.

Only the runtime's own syscalls are routed; libc- and libssl-internal I/O is untouched. Cost on a default `-O2` build: one relaxed-atomic increment per allocation / wrapped syscall. Sanitizer builds omit the shim entirely (unchanged from today).

## Testing

`crates/hale-codegen/tests/gate_counters.rs`:
- **heap gate** — a pure arithmetic loop reads delta `0` (the steady-state assertion); a byte-building loop moves the counter (proving it counts, not just always-0); the gate reports available (not -1) on a default build.
- **syscall gate** — one `recv_into` (recv timeout, no data → `EAGAIN`, no retry loop) is exactly **one** `read()` syscall.

The **corpus oracle** still runs every fixture clean with the global syscall wraps in place — the key regression check, since the wraps touch every default-build binary (including the `println` → `write` path).

## Note on stacking

Based on **#135** (TCP_NODELAY) because it shares `lotus_arena.c` / `codegen.rs` / `builtins.rs` / `purity.rs` / `spec/stdlib.md`. Retarget to `main` once #135 merges.


rileyr:
Superseded by #137 — GitHub auto-closed this when the base branch `feat/tcp-nodelay` was deleted on #135's merge. Same commit, rebased onto `main`.

PR #149, opened 2026-06-15
perf(arena): pool the subregion struct — zero-alloc locus method calls

rileyr:
## What

Every locus method call opens + closes a per-method scratch subregion. The data *chunk* was already amortized by the per-thread chunk pool, but `lotus_arena_create_subregion` still **malloc'd a fresh `lotus_arena_t` struct per call** (freed on close) — so each method call did one malloc + one free, even a trivial body. Measured via `std::diag`: exactly **1 alloc/call**.

That quietly undercuts the zero-alloc hot loops the fast-protocol-I/O work targets — a parser/handler calling `self.method(...)` per tick allocated on every call.

## Fix

A per-thread free-list of arena structs, mirroring the chunk pool: `lotus_arena_destroy` pushes the struct instead of freeing; `lotus_arena_alloc_struct` pops instead of mallocing. The `subregion_lock` mutex is initialized once on the malloc path and **kept valid (unlocked) across reuse** (not destroyed on the pool path); every other field is reset on take. Pool capped at 256, freed on thread exit by the existing pool dtor, pool-full falls back to free.

## Validation

- **1000-call method loop: 1000 → 0 heap allocs** ( pins it; no-arg + arg methods).
- **Corpus oracle clean — normal AND under ASan/LSan** (no leak, no use-after-free from the reuse).
- **Concurrency suites pass** (/) — the pool is per-thread and a struct is fully dead at destroy, so cross-thread reuse is safe.

### DIFF
--- a/crates/hale-codegen/runtime/lotus_arena.c
+++ b/crates/hale-codegen/runtime/lotus_arena.c
@@ -363,6 +363,19 @@ static __thread lotus_arena_chunk_t *
     g_chunk_pool[LOTUS_CHUNK_POOL_CAP];
 static __thread int g_chunk_pool_count = 0;
 
+/* 2026-06-15: per-thread free-list of arena STRUCTS, mirroring the chunk
+ * pool above. lotus_arena_create_subregion mallocs a fresh lotus_arena_t
+ * per call; the data chunk was already pooled, but the struct was the
+ * remaining allocation on the hot path — every locus method call opens +
+ * closes a scratch subregion, so that's one malloc + free per method call
+ * (measured: 1 alloc/call). Pooling the struct removes it. The
+ * subregion_lock mutex is created once (on the malloc path) and kept valid
+ * across reuse; every other field is reset on take (lotus_arena_alloc_struct).
+ * Freed on thread exit by the pool dtor. */
+#define LOTUS_STRUCT_POOL_CAP 256
+static __thread lotus_arena_t *g_struct_pool[LOTUS_STRUCT_POOL_CAP];
+static __thread int g_struct_pool_count = 0;
+
 /* LOTUS_CHUNK_POOL_STATS — when set, every chunk
  * acquire / release tallies into per-thread counters that
  * dump to stderr at process exit. Useful for diagnosing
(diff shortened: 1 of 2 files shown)


rileyr:
Abandoning. The struct pool achieves zero-alloc method calls and passes corpus (normal + ASan/LSan) and the concurrency suites, but it introduces a SIGSEGV in the WS1 locus-reassignment path (ws1_cross_seed_locus_reassign): on an empty pool, the reclaim-then-reinstantiate sequence makes the new child reuse the *just-reclaimed* struct, and the resulting aliasing crashes a half-initialized reassigned locus. The corpus ASan gate missed it (that cross-seed scenario isn't a fixture; nextest's process-per-test exposed it). Not worth shipping a segfault in a real feature for the alloc win — the optimization needs the reclaim/reinstantiate aliasing resolved first (e.g. instantiate-then-reclaim so the new child can't reuse the old struct).

rileyr:
## Follow-up closed: the "instantiate-then-reclaim" reorder is *not* a free fix

Verified the hypothesis from the close comment against the actual lowering (`codegen.rs::lower_locus_field_reassign`, ~21439-21496). The order today is confirmed **reclaim-then-instantiate**:

1. `__reclaim_<L>(old_ptr)` → `lotus_arena_destroy(old_arena)` → returns old's slot to the parent freelist **and** (with #149's pool) pushes old's struct onto the per-thread struct pool;
2. `lower_locus_instantiation` into self's arena → `create_subregion` pops old's just-freed slot **and** pops old's just-pooled struct → **new child === old child's struct pointer** → the aliasing SIGSEGV.

Reordering to construct-new-then-reclaim-old *would* break the reuse (the new child can't pop a struct that hasn't been pooled yet). But it's **not** a clean swap — it changes resource lifecycle from break-before-make to **make-before-make**: the new instance is constructed (and opens its `@ffi` resources — fd, socket, the `WsClient` connect this path exists for) while the old instance is still live, torn down only afterward. For a reconnect (`self.conn = WsClient{…}`, the original motivating case) that's a behavioral change with real implications (two live connections briefly; or a new instance whose construction needs the old's exclusive resource released first would now fail). It needs its own design call, not a silent reorder.

**Verdict for a future struct-pool retry:** don't reorder the lifecycle. The lower-risk path is to keep break-before-make and instead exclude the reclaim-then-reinstantiate window from struct reuse — e.g. have the reassign path hand the new child a fresh (non-pooled) struct, or have `lotus_arena_destroy` skip pooling when invoked from `__reclaim` in a reassignment. That preserves the zero-alloc win on the common method-call path (the actual target) while leaving the rare reassignment on the malloc path where it's already correct.

No code change here — #149 stays abandoned. This just closes the open follow-up thread with a grounded answer instead of the untested "(e.g. instantiate-then-reclaim)" aside.

Issue #18, opened 2026-05-27
Roadmap: formal verification surface for the language pipeline


rileyr:
This issue scopes the formal-verification candidates that belong **at the language / substrate layer**, distinguished from analyses better handled at lib/app level. The line we're drawing:

> The language provides primitives + type identity. What downstream consumers do with that across deployments (wire compat, schema evolution, transport-specific retry semantics) is library territory and shouldn't drive language design.

Five candidates that meet the language-level bar:

## 1. Memory-bound proofs

Prove at compile time that a locus's arena (or a hot path through it) is bounded by a static constant, independent of runtime input.

The motivation is a recurring leak class: factory-style returns pinning for program lifetime, hashmap cells with array fields growing per-update, hot-path String concat in observe loops. Each was invisible until OOM. A compile-time "this locus's arena cannot grow beyond N KB per epoch" pass would catch them at build time.

Shape: per-method allocation annotations + a dataflow pass that sums allocations × invocation counts on `run()` and bus-handler entry points. Loops and recursive calls need ranking. Where bounds can't be proven, the pass surfaces "unbounded — annotate explicitly with `@unbounded` or rework."

Industry precedent: SPARK Ada (used in flight-critical avionics); AbsInt's aiT for WCET analysis.

## 2. Race-completeness for substrate primitives (COMPLETE — all primitives modeled, #150)

[...]

PR #161, opened 2026-06-25
feat(check): memory-bound proof warnings are opt-in (GH #18 item 1)

rileyr:
## Summary

The unbounded-allocation warning (memory-bound proofs, GH #18 item 1) shipped **default-on** in #122. This reverts it to **opt-in**, behind `--warn-unbounded-alloc` — mirroring how `--warn-resource-leak` and the `@locality` cache-tier budgets are already flag/annotation-gated.

### Why

A bound *per epoch* only means something for a long-lived process — a daemon, a bus handler, a persistent locus. A script that allocates and exits owes the proof nothing, so by default it pays nothing: no warning, and no opt-out to remember. This is Hale's **descent-curve** stance — the proof obligation appears when you reach for the machinery that can violate it, not before. It also realigns item 1 with its siblings: `@locality` budgets and `--warn-resource-leak` were already opt-in, leaving the memory-bound warning the odd one out.

A future `@bounded` locus/module annotation will opt a scope into the proof as a hard **error** contract, with `@unbounded` as the in-scope carve-out (Phase B).

## Behavior

| Invocation | Result |
|---|---|
| `hale check app.hl` | silent (scripts pay nothing) |
| `hale check app.hl --warn-unbounded-alloc` | emits advisory warnings; build still succeeds |
| `hale check app.hl --no-warn-unbounded-alloc` | accepted-and-ignored (back-compat with the former default-on flag) |

## Changes

- `crates/hale-cli/src/main.rs` — flip emission to opt-in.
- `crates/hale-cli/tests/unbounded_alloc_opt_in.rs` (+ fixture) — pins all three behaviors so default-on can't silently return.
- `spec/verification.md` — item 1 is now opt-in; descent-curve rationale + forward pointer to `@bounded`/`@unbounded`.
- `docs/src/systems/operations.md`, `docs/src/systems/performance.md` — flag table + prose; the latter also resolves a contradiction with its own "none of these run by default" footer.
- `notes/memory-bound-proofs.md` — records the revert.

## Test

`cargo test --release -p hale-cli --test unbounded_alloc_opt_in -- --test-threads=1` → 3 passed. `hale-types` + `hale-cli` suites green (the only failure is a pre-existing `rustdoc`-can't-load-`libLLVM` env issue, unrelated).

### DIFF
--- a/crates/hale-cli/src/main.rs
+++ b/crates/hale-cli/src/main.rs
@@ -918,11 +918,15 @@ fn run_check(target: &Path) -> ExitCode {
     let allow_unowned =
         std::env::args().any(|a| a == "--allow-unowned-subscriber");
     let mut diags = hale_types::check_bundle_opts(&bundle, allow_unowned);
-    // GH #18 item 1: unbounded-allocation warnings, ON BY DEFAULT.
-    // `--no-warn-unbounded-alloc` opts out; `--warn-unbounded-alloc` is
-    // still accepted (now a no-op) for back-compat. The warnings are
-    // advisory — they print but don't fail the build (only errors do).
-    if !std::env::args().any(|a| a == "--no-warn-unbounded-alloc") {
+    // GH #18 item 1: unbounded-allocation warnings — OPT-IN, like the
+    // `--warn-resource-leak` / `@locality` budget surfaces. A memory-bound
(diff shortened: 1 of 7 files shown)


rileyr:

PR #162, opened 2026-06-25
feat(check): @bounded / @unbounded annotations for memory-bound proofs (GH #18 item 1, Phase B)

rileyr:
> **Stacked on #161** (Phase A — opt-in flag). Base is `feat/unbounded-alloc-opt-in` so this diff shows only Phase B; it'll retarget to `main` when #161 merges.

## Summary

Phase B of the memory-bound proof (GH #18 item 1). Phase A (#161) made the proof opt-in behind `--warn-unbounded-alloc` — scripts pay nothing by default. This adds the **in-source opt-in** so a long-lived locus can request the proof on itself: the descent-curve dual of the whole-program survey flag.

- **`@bounded locus L { … }`** — opts that locus into the proof on every `hale check`, no flag. The locus that took on long-lived state asks for the check on itself.
- **`@unbounded`** on a `fn` or a lifecycle hook (`@unbounded run { … }`) — the greppable carve-out. Silences that body's sites in `@bounded` scope **and** under the survey flag.
- A program with **no** `@bounded` locus stays silent by default (Phase A preserved); `--warn-unbounded-alloc` still surveys the whole program.

```hale
@bounded locus Aggregator {
    fn on_sample(s: Sample) {
        self.latest = Holder { … };   // ← flagged: accumulates in `self`
    }
    @unbounded fn on_snapshot(s: Snapshot) {
        self.cache = Holder { … };    // ← carved out: acknowledged on purpose
    }
}
```

Severity stays **warning** — the hard `@bounded`-as-*error* contract waits on the precision phases (store-latest vs. append, `@form(cap)` composition) reaching zero in-scope false positives (Phase E).

## Implementation

- **AST**: `bounded: bool` on `LocusDecl`; `unbounded: bool` on `FnDecl` + `LifecycleDecl`. Parser handles both as bare `@`-flags (contextual words after `@`). `@unbounded` is fn/lifecycle-only; a locus opts in with `@bounded` instead, and the rejection diagnostic points there.
- **Analysis** (`alloc_summary.rs`): `AllocSummary.{bounded_loci, unbounded_fns}`; `leak_sites()` drops `@unbounded`-owned sites; `unbounded_alloc_diags(progs, include_all)` gates scope (`false` = `@bounded`-only, `true` = survey).
- **CLI**: always emits `@bounded`-scope warnings; `--warn-unbounded-alloc` widens to all. Codegen treats both annotations as no-ops (propagated through monomorphization).

## Tests

[...]

PR #163, opened 2026-06-25
test(alloc): lock store-latest-is-unbounded with RSS; reframe Phase D to the capacity axis (GH #18 item 1)

rileyr:
> **Stacked on #162** (Phase B). Base is `feat/bounded-annotation`; retargets to `main` as the stack merges.

## Summary

Phase C of the memory-bound proof — **rescoped by measurement**. The design note had hypothesized a "replace-vs-append refinement" treating a whole-value field store (`self.f = X{}`) as *bounded*. RSS **falsified** it: in Hale's bump arena a replaced value isn't reclaimed until dissolve, so a per-iteration whole-value replace accumulates — even for a fixed-size `[Int; 4]`.

| variant (3M-trip `run` loop) | RSS | grows? |
|---|---|---|
| `self.recent[i] = v` (in-place indexed) | 55 MB | no — baseline |
| `self.recent = [i,i,i,i]` (`[Int;4]` replace) | 188 MB | **+133 MB** |
| `self.h = Holder{…}` (flat struct replace) | 96 MB | **+41 MB** |

The model already flags exactly the growing cases and leaves in-place writes alone — so **it's sound and there's no false positive to fix.** The refinement is dropped as unsound.

## Changes

- **`alloc_model_rss.rs`** — new `store_latest_field_replace_grows_inplace_stays_flat`: asserts the whole-value replace is model-flagged **and** adds >60 MB over the in-place baseline, while the in-place write allocates nothing and isn't flagged. A future "optimization" that wrongly suppresses store-latest now gets a red test.
- **`notes/memory-bound-proofs.md`** — records the measurement and **reframes the roadmap**:
  - Bounds live on the **capacity/projection axis**, NOT `@form`. `@form` is an access discipline (`spec/forms.md`: *"discipline goes on the capacity slot, not in the annotation"*); the `cap=N` form-arg is a size *hint*, not a bound.
  - The proof is **per storage slot**: slot 0 (the locus bump Arena) is the store-latest story; slots 1..N (capacity `pool`/`heap` + the `recognition(cap=N, fixed_cell)` projection's hard cap) are the reframed **Phase D — capacity/projection-aware slot bounding** (was mislabeled "@form(cap) composition").

`spec/verification.md` already states the whole-value replace leaks (Phase A), so no spec change — the measurement confirms it.

## Test

`cargo test --release -p hale-codegen --test alloc_model_rss -- --test-threads=1` → 2 passed.

## Next

**Phase D** — capacity/projection-aware slot bounding: route each escaping allocation to the storage slot it lands in and apply that slot's bound model (`recognition` cap = hard bound; `pool` = bounded-by-balanced-release; `heap`/`vec`/`hashmap` = grows).

### DIFF
--- a/crates/hale-codegen/tests/alloc_model_rss.rs
+++ b/crates/hale-codegen/tests/alloc_model_rss.rs
@@ -145,3 +145,93 @@ fn model_unbounded_verdict_matches_growing_rss() {
         flat_rss
     );
 }
+
+// === Phase C (2026-06-25): store-latest into a locus field is UNBOUNDED ===
+//
+// The note had hypothesized a "replace-vs-append refinement" treating a
+// whole-value field store (`self.f = X{}`) as *bounded*. RSS falsified it:
+// in Hale's bump arena a replaced value is not reclaimed until dissolve, so
+// a per-iteration whole-value replace accumulates — even for a fixed-size
+// `[Int; 4]`. These tests LOCK that finding so a future "optimization" that
+// wrongly suppresses store-latest gets a red test (and so the model's
+// flagging of it stays tied to measured RSS). See
+// `notes/memory-bound-proofs.md` §"Phase C finding".
+//
+// All three loops share the runtime bound `self.n` (not a const literal),
+// so loop-ranking can't prove them bounded — the genuine unbounded case.
+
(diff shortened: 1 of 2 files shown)


rileyr:

PR #164, opened 2026-06-25
feat(alloc): D1 slot-shape scaffold for capacity-aware bounding (GH #18 item 1, Phase D)

rileyr:
> **Stacked on #163** (Phase C). Base is `feat/alloc-store-latest-soundness-lock`; retargets up the stack as it merges.

## Summary

**D1 of Phase D** — the reusable infra for capacity/projection-aware slot bounding, with **no verdict change**. The bound solver (D2) will route each escaping allocation to the storage slot it lands in and apply that slot's bound model; D1 captures the slot map and the per-site/per-call slot identity it needs.

## Changes (all in `crates/hale-types/src/alloc_summary.rs`)

- **`LocusShape`** — capacity slots `[(name, kind)]`, `@form` + literal `cap` arg, recognition cap — distilled from each `LocusDecl` onto `AllocSummary.locus_shapes`. Pure AST read, no type inference.
- **`AllocSite.target_field`** — the `self.<field>` of a whole-value replace (`self.f = X{}`); `None` for indexed in-place writes (`self.f[i] = …`) and non-self escapes.
- **`CallEdge.receiver_slot`** — the `self.<slot>` of a slot method call (`self.log.alloc()`); `None` for free fns, self-methods, and form methods (`self.push`, slot implicit from `@form`).
- **`--dump-alloc-summary`** surfaces all three: a per-locus `[shape]` block, `->self.<field>` on sites, `recv=self.<slot>` on calls.

Example dump:
```
locus Acc [shape]
    slot  log (heap)
locus Reg [shape]
    slot  entries (pool)
    form  @form(hashmap, cap=1024)

fn Acc::run   [entry: run (one-shot)]
    alloc struct Entry  escaping=self-store  ACCUMULATES-UNBOUNDED  … ->self.latest  <-- LEAK
    call  <unresolved: alloc> loop_depth=1 result=local recv=self.log
```

## No behavior change

No `SiteVerdict` / `leak_sites` change — confirmed by the unchanged `alloc_summary` unit tests and the `alloc_model_rss` regression. Five new D1 unit tests pin the capture (slots + form cap, recognition cap, store-latest field, indexed-write-has-no-field, slot-insert receiver). `hale-types` lib: 204 passed.

## Discovery findings (recorded in `notes/memory-bound-proofs.md`)

- The walker tags slot-insert **arguments** `Local` (`alloc_summary.rs:888`), so collection growth (`self.items.push(X{})`) is **currently undetected** — that's D2's job.
[...]

PR #165, opened 2026-06-25
feat(alloc): D2 flag growing @form(vec|hashmap) inserts (GH #18 item 1, Phase D)

rileyr:
> **Stacked on #164** (D1). Base is `feat/alloc-slot-shape`; retargets up the stack as it merges.

## Summary

**D2 of Phase D** — detect unbounded growth into collection slots, the leak class the type-free walker missed (it tagged insert *args* `Local`). A discovery pass found the corpus inserts via `v.push(x)` on a **typed** receiver (`v: IntVec`, `self.buf: IntVec`), not direct `self.<slot>.alloc()` — so detection is inherently type-aware, the same class as the deferred String-concat. Chosen approach: **D2-lite** — a lightweight *declared*-type map, no resolver coupling.

## How it works

- A var→declared-type map seeded from fn params + typed `let`s, plus a locus param-field→type map, matched against D1's `locus_shapes[T].form`.
- A `push`/`set`/`insert`/`add` whose receiver type is a `@form(vec | hashmap)` locus becomes an `AllocKind::CollectionInsert` site (escape = self-store, reclaim @ dissolve), so it flows through the **existing** verdict path:
  - flagged in an unbounded context (per-message handler / runtime loop)
  - bounded by a const loop (`accumulates×const`)
  - suppressed by `@unbounded`, gated by `@bounded`/survey
- `@form(ring_buffer | lru_cache)` is cap-bounded → **not** flagged. A user-defined `push` method on a non-form locus → **not** flagged (the `54-geom-leading-edge` `Segment` shape).

## Validation

- **Zero corpus false positives** — surveyed all 68 example fixtures.
- **RSS-validated** (`alloc_model_rss`): a 3M `@form(vec)` push of a 32-byte struct hits **~210 MB** vs a **~55 MB** no-push control. Confirmed `lotus_vec_push` grows a geometric `cap*2` buffer with the element memcpy'd in — the vec genuinely accumulates.
- **6 D2 unit tests**: vec field push in a handler flagged; param push called once bounded; ring_buffer not flagged; bounded loop bounded; user `push` on a non-form locus not flagged; typed-`let` receiver resolves. `hale-types` lib: 210 passed.

## Deferred (to a type-aware stage, with String-concat)

Untyped `let`, reassignment, return-of-form receivers, and the clear/reset balance case (a vec cleared each iteration is bounded-by-peak). The opt-in gating + `@unbounded` carve-out absorb the false-positive edge until then.

## Next

**D3** (per-cell field bounds — a bounded cell with a growing `String` field) — optional; or fold the remaining type-aware detections (untyped receivers + String-concat) into one type-aware stage. **Phase E** promotes `@bounded` to a hard error once in-scope FPs hold at zero.

### DIFF
--- a/crates/hale-codegen/tests/alloc_model_rss.rs
+++ b/crates/hale-codegen/tests/alloc_model_rss.rs
@@ -235,3 +235,78 @@ fn store_latest_field_replace_grows_inplace_stays_flat() {
         inplace_rss
     );
 }
+
+// === Phase D / D2 (2026-06-25): a growing @form(vec) insert accumulates ===
+//
+// D2 flags `v.push(x)` where `v`'s declared type is a growing `@form(vec |
+// hashmap)` locus, in an unbounded context. `lotus_vec_push` grows a
+// geometric doubling buffer (cap*2) with the element memcpy'd in, so the
+// vec genuinely accumulates with the push count. This ties that verdict to
+// measured RSS so a future change that wrongly treats a vec insert as
+// bounded gets a red test. (Element is a 32-byte struct for a clean signal
+// above the test build's ~50 MB baseline; the loop bound `self.n` is
+// runtime, so loop-ranking can't prove it bounded.)
+
+const VEC_PUSH_GROWS: &str = r#"
+    type Cell { a: Int; b: Int; c: Int; d: Int; }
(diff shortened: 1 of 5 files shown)


rileyr:

PR #203, opened 2026-07-15
fix(runtime): timed park for recv_into on async_io pools

rileyr:
**1/6 of the downstream-handoff substrate fixes (2026-07-14).** Merge this stack in order; each later PR is based on the previous branch and will retarget to main automatically.

On a `where async_io` pool the fds are nonblocking, so `recv_into` surfaced EAGAIN as an instant `-2` instead of parking the coroutine. Anything built on the documented "`-2` = recv deadline expired" contract — pond/websocket's ping/pong liveness in particular — saw two instant `-2`s and declared healthy idle peers dead within microseconds, making the contrib WS server unusable on exactly the pool type it's documented for.

## The fix: a timed park

- New `lotus_coop_park_on_fd_deadline(fd, events, deadline_ns)`: parks on epoll until the fd is ready (`0`) or an absolute CLOCK_MONOTONIC deadline expires (`1`); `-1` = park unavailable (falls back to today's behavior). Deadline `0` parks indefinitely; a pre-expired deadline returns without touching epoll.
- The drain loop bounds `epoll_wait` by the earliest parked deadline (rounded up — no 0ms busy-spin) and, after the readiness pass, sweeps expired coros back in with a timed-out flag. The wake_fd cross-pool machinery is untouched; bus cells keep draining while coros are parked.
- The deadline mirrors the socket's `SO_RCVTIMEO`, read back via `getsockopt` once per recv call (the socket stays the single source of truth — survives fd transfer across the bus, no lotus-side table). So `-2` means *deadline expired* on every pool type, and pond's machinery composes unchanged.
- Applied to: tcp `recv_into` / `recv_stamped`, udp `recv_into`, tls `recv_into` / `recv_stamped_into` (WANT_WRITE renegotiation parks on EPOLLOUT), and `recv_bytes`.

## Contract alignments (spec'd in `spec/stdlib.md` rows tcp/udp/tls)

- `recv_bytes` now honors `set_recv_timeout` on async pools — it parked indefinitely before, silently ignoring the timeout.
- udp `recv_into` returns `-2` retryable on timeout — it previously fell into `-1` fatal.

## Tests

[...]

PR #204, opened 2026-07-15
fix(stdlib): http Server reassembles split-written request bodies

rileyr:
**2/6 of the downstream-handoff substrate fixes (2026-07-14).** Stacked on #203; retargets to main when it merges.

`std::http::Server` read each request with a single `recv(8192)`, so a client that writes headers and body in separate TCP segments — python urllib does exactly this — got its body truncated: a valid POST failed while curl (single write) worked, nondeterministically by segmentation. Any HTTP surface consumed by third-party clients hits this.

## The fix

`__http_handle_one_conn` now accumulates to the `\r\n\r\n` header terminator, then to `Content-Length` body bytes, reusing the existing `parse_request` / `header` / `write_response` pieces (a GET with no Content-Length exits as soon as the terminator arrives — one recv, same syscall count as before). Guards:

- **1 MiB request cap** — a declared `Content-Length` over the cap answers `413 Payload Too Large` before buffering; overflow without a complete header block closes without a response (the framing can't be trusted).
- **5s recv timeout** (`__set_recv_timeout_ns`) — bounds a slowloris-style stalled client on classic/pinned pools; inert on `async_io`, where a stalled conn parks only its coroutine.

Keep-alive stays unsupported (`Connection: close`); `Transfer-Encoding: chunked` is not parsed. Spec row + docs book updated in the same change.

## Tests

New `http_split_write.rs`: the urllib head/sleep/body split, a three-way split, single-write GET answers immediately, oversized declared Content-Length → 413, silent-after-headers bounded by the timeout. All existing http suites stay green.


rileyr:
Landed on main as fecabff via the stack merge in #207 (single main CI run for the whole stack).

PR #205, opened 2026-07-15
feat(typecheck): warn when non-returning run() bodies starve a cooperative pool

rileyr:
**3/6 of the downstream-handoff substrate fixes (2026-07-14).** Stacked on #204.

A cooperative pool runs each `run()` cell to completion in birth order, so two loci with statically non-returning `run()` bodies on one pool means the second never starts — *silently*: bus handlers still fire at sleep/yield drains, so the hang looks like a healthy idle. The existing diagnostics only catch blocking *calls* (stdlib allowlist), and the dead-receiver error requires the starved locus to be a bus subscriber. The reported shape slipped both: a publish-only 1s metronome (`while !self.draining { sleep(1s); Tick <- …; }`) placed on pool `main` starved the main locus's own `run()` — boot "completed" with all threads idle and no ports bound.

## The check

- New conservative predicate `run_statically_nonreturning`: the terminal `while` has no `break`/`return`/`terminate`/`fail`/`violate` at any statement depth, and its condition provably never flips — `while true`, `while !self.draining` (the synthetic drain flag only flips at shutdown), or a Bool params flag no member body ever assigns whose declared default keeps the loop live. Handler-flipped flags, bounded counters, and loops with exits never fire it.
- `check_cooperative_pool_blocking` gains a grouping phase over **all** main params fields — including fields with **no placement entry** (they default to pool `main`, the gap that bit) and the main locus itself, birth-ordered last (its `run()` starts only after params-init). `pinned` and `where async_io` entries are exempt; `std::http::Server`-class stdlib loci count via the known-long-running list.
- One warning per pool naming every offender, with the established "use `pinned` (its own thread + a mailbox drained at sleep/yield)" guidance; suppressed where the dead-receiver *error* already fired for that pool.

Spec (`spec/verification.md`) + docs book updated. Ten new cases in `placement.rs` (positive pairs, the metronome-vs-main shape both implicit and explicit, and the full negative matrix); all existing placement tests green.


rileyr:
Landed on main as 9d3bdc2 via the stack merge in #207 (single main CI run for the whole stack).

PR #206, opened 2026-07-15
fix(codegen): resolve self.<scalar> in param defaults lexically

rileyr:
**4/6 of the downstream-handoff substrate fixes (2026-07-14).** Stacked on #205.

```hale
locus Conn {
    params {
        fd: Int = -1;
        conn: Ws = Ws { conn_fd: self.fd };
        // codegen error: no field `fd` on locus self
    }
}
```

Typecheck passed (it skips default exprs — the Milestone-2 cut), then codegen died — but only when `Conn` was instantiated from inside another locus's method body (a per-connection child born in an accept/dispatch handler). Root cause: the enclosing method's `current_self` took precedence over `params_init_self` in the `self.X` read, so the default's `self.fd` resolved against the wrong locus. The F.4 override-context swap itself was correct all along (`sibling_field_forward_ref.rs` proves it).

## Semantics, now pinned in `spec/semantics.md`

`self` resolves to the locus whose source text **lexically contains** the expression: default text (including nested-literal field inits written inside a default) → the declaring locus; a locus-literal field init written in a method body → that method's locus (the F.4 call-site rule); recursive defaults → each to its own innermost declaring locus.

## Mechanism

A `Cx.in_params_default` flag set while lowering a locus's own default text inverts the `current_self`/`params_init_self` precedence. Override expressions restore the flag saved at frame entry (mirroring the F.4 `prev_params_init_self` swap), and the three instantiation sites that lower *member* text with a deliberately-installed `current_self` — bus `where key == self.X` clauses (cooperative + pinned) and `birth_check` conds — clear it, so keyed subscriptions on loci born in defaults keep resolving against the new locus (`bus_routing_keys` pins this).

Also new: a declaration-order guard — a default reading a **later-declared** sibling is now a compile error ("reads `self.b` before it is initialized") instead of a silent GEP+load of an uninitialized slot, tracked by an initialized-fields set that travels with the params-init context.

## Tests

New `params_default_scalar_self_ref.rs`: the four instantiation quadrants (fn main / main params default / method body / field reassign), call-site-override-still-resolves-to-caller (F.4 pin), two-level recursion, forward-ref rejection. Must-stay-green all green: `sibling_field_forward_ref`, `ws1_ffi_handle_reassign`, `locus_let_lifecycle`, `bus_routing_keys`, `topic_phase2*`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,13 @@ Eight substrate findings from a downstream service built on hale
   (including fields with no placement entry and the main locus's
   own `run()`) warn naming every offender — the second-born
   `run()` never starts, and the failure was silent.
+- **`self.<scalar>` in nested-literal param defaults works.**
+  `conn: Ws = Ws { conn_fd: self.fd }` now resolves `self`
+  lexically (the declaring locus) even when the instantiation
+  happens inside another locus's method body; call-site overrides
+  keep resolving to the caller (F.4). A default reading a
+  later-declared sibling is now a compile error instead of an
+  uninitialized read.
 - Filed as issues: migrating `Stream.send`/`recv` to
   `fallible(IoError)` (finding 5) and implicit error propagation
   on tail-position `return` (finding 8).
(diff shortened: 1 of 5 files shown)


rileyr:
Landed on main as bcb65f2 via the stack merge in #207 (the whole stack was merged in one push so main CI runs once).

PR #208, opened 2026-07-15
fix(parser): name the reserved keyword when it lands in a binding position

rileyr:
**6/6 of the downstream-handoff substrate fixes (2026-07-14).** Stacked on #207.

`let accept = …;` errored with `expected variable name, got Accept` — technically true, but the Debug variant name sends users on a grammar hunt instead of saying the word is reserved.

Now: ``expected variable name, but `accept` is a reserved lifecycle keyword in Hale — pick another name``. The lifecycle wording covers `birth`/`accept`/`run`/`drain`/`dissolve`/`on_failure`; every other hard keyword gets "a reserved keyword"; the Debug fallback stays for non-keyword tokens. Implemented in shared `expect_ident`, so all 62 call sites (param names, field names, …) inherit it. The keyword-name mapping is a new `TokenKind::keyword_lexeme()` — the reverse of the lexer's keyword match — kept honest by a round-trip test against `keywords::HARD_KEYWORDS` in `keyword_sync.rs` (the same sync discipline the highlighter lists use).

No spec change (diagnostic texts aren't pinned in spec); CHANGELOG line included.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,9 @@ Eight substrate findings from a downstream service built on hale
   longer flag.** Both diverge — the payload allocates at most once
   per invocation. Removes the false-positive class on strict
   parsers (`fail E { … }` inside `while`).
+- **Parser: reserved keywords in binding position are named.**
+  `let accept = …` now says ``expected variable name, but `accept`
+  is a reserved lifecycle keyword in Hale — pick another name``.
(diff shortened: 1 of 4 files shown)


rileyr:
For the record: this landed on main as db23b50 via the stack merge in #207 (it had rebase-merged into its stack parent `alloc-lint-divergence` first).

PR #210, opened 2026-07-15
feat(stdlib)!: Stream send/recv migrate to fallible(IoError)

### LINKED ISSUE #209: Migrate Stream.send/recv to fallible(IoError) — `send(...) or ...` currently dies in codegen
From a downstream handoff (2026-07-14), finding 5.

## Symptom

`Stream.send(msg) or <disposition>` fails codegen:

```
unsupported in codegen v0: fallible method `__StdIoTcpStream.send` — not a synthesized @form ... or a user-declared `fn` member with `fallible(E)`
```

`send_bytes` "works" only because callers don't use it in an `or` context.

## Root cause

Neither `send` nor `send_bytes` is actually declared `fallible` — they're legacy `-1/0` sentinel methods (`crates/hale-codegen/runtime/stdlib/io_tcp.hl`). An `or`-context call routes into `lower_fallible_method_call` (`crates/hale-codegen/src/channels/mod.rs`, the `Unsupported` arm), which finds no `fallible(E)` declaration and errors.

`spec/stdlib.md` (§ Stream methods, "legacy sentinel shape") already marks migrating these to `fallible(IoError)` as follow-on work — they predate open-question #24 (2026-05-25) lifting the "no fallible on locus methods" restriction.

## Suggested fix

Migrate `Stream.send` / `send_bytes` / `recv` / `recv_bytes` to `fallible(IoError)` (declaration in `io_tcp.hl` + whatever `try_lower_user_locus_fallible_method` needs). Watch out: `locus_member_fallible_rejects.rs` notes v0.1 rejected String returns from fallible methods — check that constraint for `recv`. Spec + docs update in the same change; existing sentinel-shape callers (pond) need a migration note.

## Workaround

Don't address `send` with `or`; check the returned Int sentinel (`n < 0`).

rileyr:
Closes #209.

The four `std::io::tcp::Stream` methods carried a legacy `-1/0` sentinel shape — `s.send(x) or …` died in codegen (`unsupported in codegen v0: fallible method`), and `recv` returned empty on both genuine errors and clean EOF with no way to tell them apart.

## The new contract

| method | success | fails on |
|---|---|---|
| `send(msg)` / `send_bytes(b)` | Unit (`or discard` for fire-and-forget) | any write error |
| `recv(max) -> String` / `recv_bytes(max) -> Bytes` | the data; **empty on EOF *and* recv-timeout** | genuine I/O errors only (bad fd, `connection_reset`, `broken_pipe`, …) |

Timeout stays a liveness signal, not an error — pond-style ping/pong loops and the http server's "serve what we have" shape keep working unchanged in structure.

## Mechanism

- The C send/recv primitives record their outcome in an errno-style thread-local (`0` = ok / EOF / timeout, errno on genuine failure); the `.hl` method bodies read it via new `__last_io_status` / `__io_error_kind` path-calls and `fail IoError { … }` with the existing errno→kind taxonomy — the same tags the fallible path-calls (`connect`, `listen_socket`, …) already produce.
- `IoError` is now **declared in the stdlib seed** (`io_tcp.hl`), field-identical to the types-layer injection (which guards itself off) and codegen's `emit_io_error_alloc` — so user code can construct and `fail` it directly.
- No compiler changes were needed for the method surface itself: the fallible locus-method machinery already supports Unit/String/Bytes success, and typecheck derives fallibility straight from the seed `FnDecl`.
- Bonus: `Stream.recv` joins the async_io **timed park** its `recv_bytes`/`recv_into` siblings got in ed24923 (park deadline mirrors `SO_RCVTIMEO`), so `set_recv_timeout` now bounds it on async pools too.

## Breaking change

Every call site must address the error. All 34 in-tree sites migrated: the stdlib http server (`or discard` on response writes, `or ""` folding read errors into the EOF path), both fixture examples, nine test suites (`or raise`), and the spec/docs examples. Sentinel-checking callers migrate as: `let n = s.send(x); if n < 0 {…}` → `s.send(x) or handler(err);`. CHANGELOG carries the migration note; the spec's "migration is follow-on work" paragraph is replaced by the shipped contract.

## Tests

New `stream_fallible_io.rs` pins: bad-fd send/recv fail with kind+errno; EOF is empty success (`or` never fires); a 200ms recv-timeout returns empty without raising; send-to-closed-peer fails `broken_pipe`/`connection_reset`; `or discard` on Unit sends. Full workspace suite green (1,674 tests, serial run).


rileyr:
Landed on main as 927e2f2 via the #212 stack merge (single main-CI run for the two-PR stack). Thanks!

PR #211, opened 2026-07-15
feat(stdlib): add std::io::tls::upgrade socket-upgrade primitive + verify toggle

adisney:
## What

Adds `std::io::tls::upgrade(fd: Int, host: String, verify: Bool) -> Int fallible(IoError)` — wraps an **already-connected** TCP fd in a client TLS session, enabling STARTTLS-style protocols. First consumer: pond/pq's Postgres `SSLRequest` flow, needed for daydream to reach AWS RDS (which enforces TLS).

- **SNI always sent.** `verify=true`: chain + hostname verification against the system trust store (same as `connect`). `verify=false`: encrypt without authenticating the peer — `sslmode=require` semantics, needed because RDS certs are not in system trust stores.
- **fd ownership**: on failure, `upgrade` does NOT close the fd — the caller owns it (asymmetric with `connect`, documented in spec/stdlib.md along with hostname-vs-IP and double-upgrade caveats).
- `lotus_tls_connect` refactored to dial + `upgrade(fd, host, verify=1)` internally with its external contract preserved (including errno ordering across the failure-path `close`), so existing TLS tests double as regression coverage.
- Returned handle is fully interchangeable with `connect`'s across `send_bytes`/`recv_bytes`/`recv_into`/`close`/`set_nodelay` (same handle table).

## Testing

- Default suite (network-free): bad-fd arg-guard test + a pin-down test that drives the **real handshake-failure path** (upgrade a listening socket's fd) and proves the fd is left open by closing it successfully afterward.
- `#[ignore]` network tests: dial+upgrade(verify=true) reproduces `connect` behavior end-to-end; upgrade(verify=false) handshakes; both pre-existing `connect` network tests pass unchanged (refactor regression net). 4/4 pass.
- Full `cargo test -p hale-codegen` compared against a clean-baseline run: no new failures.
- E2E: a .hl program doing `tcp::connect` + `tls::upgrade` against a live HTTPS host succeeds under both verify modes.

## Notes for reviewers

- Touches `builtins.rs`, `stdlib_surface.rs`, `spec/stdlib.md` — overlaps files with #210 (fallible Stream migration); whichever lands second needs a small rebase.
- Known pre-existing issue kept out of scope to preserve `connect` behavior exactly: a failed `SSL_set1_host` is non-fatal on the verify path (logged; tracked separately).


adisney (commented):
## PR Review: feat(stdlib): add `std::io::tls::upgrade` + verify toggle

### Summary
Clean, well-documented addition. The `connect → dial + upgrade(verify=1)` refactor is a faithful extraction — I diffed it against the pre-refactor `connect` and checked the locking model, errno handling, fd-ownership on every error path, and the codegen/surface/sig wiring. No correctness or safety blockers. Findings below are all minor (observability / test-hygiene / docs).

### Verified as correct (no action needed)
- **Locking model preserved.** Both the old `connect` and the new `upgrade` release `g_tls_mutex` immediately after `SSL_new` and re-acquire it only for the handle-table mutation; `lotus_tls__recvmsg_bio()` self-guards its lazy singleton init with the same mutex. So doing the handshake outside the lock is not a new race — it matches prior behavior and the shared `SSL_CTX` is safe for concurrent `SSL_new`/handshake.
- **All five `upgrade` error paths** (bad-arg, ctx, `SSL_new`, handshake, realloc) balance the mutex and free the `SSL`; none close `raw_fd`, matching the documented caller-owns contract. `SSL_free` / `bio_destroy` don't close the fd either, so there's no hidden double-close via the freed BIO.
- **`connect`'s `saved_errno`** is captured before *both* `fprintf` and `close()`, so `upgrade`'s real errno survives to the Hale `IoError.kind` derivation (the comment only mentions `close`, but the placement covers `fprintf` too).
- **Wiring matches end-to-end:** builtins decl `i32(i32, ptr, i32)` == C `int(int, char*, int)`; the lowering's arg order/types (fd truncate→i32, verify zext→i32, host ptr) match; `stdlib_surface` SIG `[Int, Str, Bool] -> Int fallible(IoError)` and the (still-alphabetical) closed-list entry are correct.

### Findings — all minor

1. **`runtime/lotus_tls.c`, `lotus_tls_connect` (~L418-420) — the "handshake failed" log now fires on non-handshake failures and double-logs on the real one.** Pre-refactor, `connect` printed "handshake failed" only on an actual `SSL_connect` failure. Now it prints for *every* `upgrade` error: a ctx / `SSL_new` / realloc `ENOMEM` failure prints `lotus_tls_connect: handshake failed (...)` though no handshake was attempted, and a genuine handshake failure prints *two* lines (`lotus_tls_upgrade: handshake failed ...` then `lotus_tls_connect: handshake failed ...`). The port-bearing line is worth keeping, but consider gating/rewording so "handshake failed" isn't attached to alloc failures — e.g. emit it only when `errno == ECONNREFUSED`, or word it generically (`connect failed (host=%s port=%u)`). Minor, but this codebase leans hard on log accuracy.

2. **`tests/io_tls.rs`, `upgrade_failure_leaves_fd_open` — hardcoded port `47847` is a needless flakiness risk.** The test never connects to the port; it only needs a bound-but-unconnected fd to drive the `send→ENOTCONN` handshake-failure path. Bind to port `0` (ephemeral) instead of a fixed port to remove any CI collision / `TIME_WAIT`-reuse hazard.

3. **`spec/stdlib.md` (tls row) — add the STARTTLS over-read caveat.** The double-upgrade and hostname caveats are documented; the classic STARTTLS footgun isn't: the caller must read *exactly* the plaintext prologue (e.g. the single `'S'` byte) and must not buffer/over-read past it before calling `upgrade`. Structurally the risk is low here because `upgrade`'s recvmsg BIO and `tcp::recv` are both unbuffered (no shared read buffer to smuggle plaintext through), but a one-line caveat would harden the first consumer (pond/pq) and future ones against the plaintext-injection class.

4. **`upgrade(fd, ...)` takes a raw fd, not a TLS handle — worth stronger emphasis (and maybe a debug guard).** TLS handles from `tls::connect`/`upgrade` are small table indices (0, 1, 2, …) that alias real fd numbers, and both are typed `Int`. Passing a TLS handle where a raw fd is expected would silently `SSL_connect` on fd 0/1/2. The doc already anchors "the fd came from `tcp::connect`" — given the failure is silent rather than a type error, consider making that even more prominent, or a cheap sanity check. Pre-existing to the `Int`-as-fd/handle design across the tcp/tls/udp surface, so not a blocker.

### Verdict
Looks good with minor suggestions. No correctness or safety blockers; the refactor is behavior-preserving as claimed, and the two adversarial-review fixes (errno-across-close, fd-left-open pin-down test) hold up.

adisney:
Thanks for the thorough review. All four findings addressed in d470b72 (no behavior change; errno + return codes preserved).

**1. `lotus_tls_connect` "handshake failed" log mislabels alloc failures / double-logs.** Fixed. Reworded the connect-level failure log from `handshake failed (host, port)` to `upgrade failed (host, port)`. It's now accurate for *all* upgrade error paths (ctx / `SSL_new` / realloc ENOMEM as well as a genuine handshake failure), no longer duplicates upgrade's own `handshake failed` line on the real-handshake case, and still carries the connect-level host/port context that upgrade's line lacks. Went with the generic-reword option rather than gating on `errno == ECONNREFUSED`, because gating would keep the alloc paths silent at the connect level *and* still double-log on the handshake case — the reword satisfies both concerns at once. Added a comment explaining why.

**2. `upgrade_failure_leaves_fd_open` hardcoded port 47847.** Fixed. Now binds port `0` (kernel-assigned ephemeral) via `listen_socket("127.0.0.1", 0)`. Confirmed it still exercises the real handshake-failure path, not the arg guard: the fd is valid (`>= 0`, so no `raw_fd < 0` short-circuit), `SSL_connect`'s ClientHello `send(2)` hits ENOTCONN on the never-connected socket, and the test still asserts `upgrade_err` + `close_ret=0` (fd left open) with no `unexpected ok`. Passes.

**3. STARTTLS over-read caveat missing from spec.** Fixed in `spec/stdlib.md`. Added a caveat that the caller must read *exactly* the plaintext prologue before `upgrade` (any over-read consumes the peer's TLS `ServerHello` out of the kernel buffer, corrupting the handshake — the classic plaintext-injection footgun), plus a note that the current design is safe *because* `upgrade`'s recvmsg BIO and `tcp::recv_*` share no read buffer, and that a future read-ahead/buffered `tcp` reader would break this property.

**4. fd-vs-handle `Int` aliasing warning.** Fixed in `spec/stdlib.md`. Strengthened to an explicit warning that raw fds and TLS handles both alias numerically as `Int` (handles are small table indices 0,1,2,… that collide with real fd numbers), so passing a handle to `upgrade` (→ `SSL_connect` on stdio) or an fd to `tls::send_bytes`/`recv_bytes` (→ garbage handle-table index) is silent misuse — corrupt I/O, no type error — with the failure mode spelled out.

Verification: `cargo build` OK; default `io_tls` suite passes incl. the modified `upgrade_failure_leaves_fd_open`; all 4 `#[ignore]` network `io_tls` tests pass (incl. `handshake_failure_returns_fallible_error`, which exercises exactly the connect failure path reworded in finding 1); `tls_fast_io` passes.

adisney:
Rebased onto current main (`db23b50`, post-v0.10.0) — 45 upstream commits. Conflicts were confined to `spec/stdlib.md` (the docs sweep rewrote the io table rows this PR documents): resolved by taking main's updated rows — including the new TLS `recv_into` async-io parking semantics — and re-grafting this PR's `upgrade` documentation, caveats, and footgun warnings into them. No code conflicts.

Re-verified post-rebase: `cargo build` clean, io_tls default suite 2/2, network suite 4/4 (`--ignored`), and downstream sanity — pond's `pq` (the first consumer, hale-lang/pond#14) typechecks and passes its offline suites (122/122 via the new `hale test`) against a toolchain built from this rebased branch.

adisney:
CI note: the `test 2/4` failure (`ownership_bubble_crosspool::world_collects_crosspool_bubbled_ships`, SIGSEGV/139) is a pre-existing flake on `main`, not introduced by this PR. The same test fails identically (same assertion, same segv) on main's own CI runs from before this PR existed: [29295409200](https://github.com/hale-lang/hale/actions/runs/29295409200) (Jul 14), [29292040438](https://github.com/hale-lang/hale/actions/runs/29292040438) (Jul 13), [28894541847](https://github.com/hale-lang/hale/actions/runs/28894541847) (Jul 7) — 3 of the last 15 main runs. It passes locally on this branch, and this PR's diff doesn't touch the cross-pool bubble machinery (changes are confined to `lotus_tls.c`, the tls stdlib surface, docs, and tls tests). Looks like the same class #181/#195 have been hardening. Re-pushed (amend, content unchanged) to retrigger CI since fork PRs can't rerun failed jobs on this repo.

rileyr (approved):
Nice, thanks!

PR #257, opened 2026-07-27
Deliver GH #253 (teardown delivery contract) + GH #254 (std::compress + std::tar)

### LINKED ISSUE #253: Teardown order can dissolve a live subscriber before sibling publishers finish (silent drop at scope exit)
From the hale-bun handoff (UPSTREAM.md item 2) — hit for real in hale-bun's install fan-out.

## Shape

- A `main locus` holds `tally: Tally` (cooperative, subscribes `install.done`) declared before `f0..f3: Fetcher` (pinned, publish `install.done`).
- Main's `run()` publishes jobs and returns immediately.
- At scope exit, Tally is drained + dissolved **before** the pinned fetchers drain their job queues. Every `Done` the fetchers then publish has no live receiver and is silently dropped. The work completed (files on disk); the messages vanished.

The compile-time dead-receiver check passes — a subscriber exists statically. This is its runtime-lifetime cousin: the receiver was real but already dead.

hale-bun fixed it structurally (parent polls `tally.is_done()` in sliced sleeps before exiting — "exit when done" beats "hope the drain order is friendly"), which is arguably the right app shape anyway. But three language-side options are worth a decision:

a. **Dependency-ordered teardown** — drain/dissolve children in reverse bus-dependency order (publishers before their subscribers). The bus graph is already known to the compiler. Has a cycle problem (mutual pub/sub).
b. **A diagnostic** — warn when a locus's teardown can outlive a sibling publisher on a topic it subscribes to. The bus-graph verification track (#18, item 4) already has the edges. Likely the honest ceiling given (a)'s cycles.
c. **Document the drop** — if silent-drop-at-teardown is the contract, spec it; today it reads as a declaration-order accident.

This is exactly the class of system-shaped hazard the verification track exists for; tagging it against #18.

### LINKED ISSUE #254: stdlib: compression (gzip / tar / zstd)
From the hale-bun handoff (UPSTREAM.md item 5): compression was the entire distance between hale-bun's parody registry and speaking the real npm protocol. Also blocks HTTP `Content-Encoding` and any archive tooling. Probably the highest-leverage stdlib addition for "write real infra tools in Hale."

Likely shape: system-library bindings in the lotus runtime (zlib is effectively universal; libzstd already ships in the macOS artifact vendoring sweep) surfaced as `std::compress::*` free fns over `Bytes`, plus a tar reader/writer in pure Hale on top. Streaming vs one-shot API is the main design question — one-shot `Bytes -> Bytes fallible` covers the npm/registry/`Content-Encoding` cases and can ship first.

rileyr:
Closes #253. Closes #254.

## #253 — teardown order dissolved live subscribers (options decided: a targeted slice of (a), plus (c))

Investigating the hale-bun shape showed the hazard was **worse than declaration order**: reordering didn't help at all. The parent (`main locus` with no subscriptions of its own) dissolves *eagerly* right after `run()` returns — cascading its subscriber fields' teardown and destroying the arena that holds the pinned children's self structs — while the pinned worker threads were still running (their joins were deferred to fn-scope exit). Every final publish was dropped in ANY order, and the arena-lifetime overlap was a latent UAF.

Shipped:
- **Eager path**: a dissolving parent joins its own pinned children (mailbox shutdown → `pthread_join` → bus drain) *before* the field cascade — the pinned mirror of the existing 2026-06-01 pool-worker join rule.
- **Fn-exit flush**: subscription-less pinned entries join before any cooperative entry's teardown (pinned entries *with* mailboxes keep their slot so dissolve-publishes can still reach them).
- **The contract, spec'd (option c)**: spec/runtime.md "Teardown delivery contract" — what's now guaranteed, what still drops by design (publish after the last subscriber dissolved, e.g. pool-queued work outliving teardown), and the "exit when done" guidance. docs bus chapter gets the user-facing version.
- Self-checking corpus fixture `72-teardown-publish-delivery` (subscriber declared first — the historically worst order; exits 1 on any missed event).

Option (b) — a bus-graph diagnostic for the *residual* shape (pool-placed publisher with queued work at teardown) — stays with the #18 verification track: after this fix the remaining hazard needs run()-blocking analysis to warn without over-firing, which is that track's machinery.

Repro behavior before/after: `tally saw 0 of 2` (any order) → `tally saw all 2` (any order), with fetchers joined before the subscriber dissolves.

## #254 — std::compress + std::tar

- `std::compress::{gzip, gunzip, zstd, unzstd}(b: Bytes) -> Bytes fallible(IoError)` — one-shot. gzip rides zlib (`-lz`, universal); **zstd is dlopen'd at first use** so neither `hale build` nor emitted binaries gain a link-time libzstd dep — no-lib machines get a clean `not_found`. Corrupt input → `invalid`; 1 GiB one-shot guard (zip-bomb = error, not OOM). gunzip auto-detects gzip vs bare zlib.
- `std::tar` — ustar subset over Bytes. Indexed read (`entries` / `entry_name` / `entry_size` / `entry_type` / `entry_data`, prefix field honored), append-style write (`pack` / `pack_dir` / `finish`, long-name prefix split, mtime 0 for reproducible archives).
- `std::io::fs::write_bytes(path, b)` — the binary-safe file write the pipeline needed (`write_file` is strlen-based and truncates at the first NUL; there was no way to ship an in-memory archive to disk).
- New TU `runtime/lotus_compress.c` (same isolation rationale as lotus_tls.c). CI already installs zlib1g-dev + libzstd-dev.
- `compress_tar.rs` pins in-Hale round-trips, magic bytes, corrupt-input kinds, AND **system `tar -tzf` acceptance** of a Hale-built `.tar.gz` — wire-format compatibility, not just self-consistency. zstd assertions degrade to a documented skip where libzstd is absent.
- Spec rows for both namespaces + the files.md book chapter (snippet compile-verified).

Deferred, noted in spec: compression level parameters, streaming (the >1 GiB path), zstd frames without a declared content size.

Local verification: corpus oracle (incl. both new fixtures), compress_tar, tcp_raw_fd_freefns, 11-suite bus/teardown batch, hale-types, hale-syntax (docs snippet gate) — all green; `hale fmt --check` clean.


rileyr:
The first CI run's `test 2/4` failure was the static/dynamic differential catching a **real pre-existing devirt soundness bug** via the new fixture: `72-teardown-publish-delivery` diverged (`saw 1 of 2` static vs `saw all 2` dynamic, load-dependent).

Root cause: `direct_call_eligible` constrained *subscriber* placement (same-thread) and handler quietness, but never looked at **publisher** placement. A direct call executes the handler on the *publishing* thread — so a pinned publisher ran the same-thread subscriber's quiet handler off-main, and the fixture's two pinned fetchers ran it **concurrently**, occasionally losing a `self.seen + 1` read-modify-write. Reachable by any pinned/pooled publisher into a quiet same-thread subscriber; the fixture's shape (two synchronized pinned publishers + a quiet counter handler) is what finally surfaced it.

Fix in d091192: direct-call eligibility now also requires every publisher same-thread; off-thread publishers stay on the enqueue path, which serializes dispatch on the draining thread. All devirt suites (`bus_devirt_direct`, `bus_devirt_no_pinned`, the differential) green locally.

Issue #354, opened 2026-08-03
@effects has no closed allow-list: none: is open-ended and fails open as classes are added


rileyr:
`@effects(none: {…})` is a blacklist and there is no dual. Every other clause in the system is a *complete* declaration; the categorical one is the odd one out, and its open-endedness fails open as the language grows.

## The gap

`none:` forbids a listed set and permits everything else — verified:

```hale
@effects(none: {syscall})
fn f(n: Int) -> String {
    let t = std::time::monotonic_ns();   // time
    return "a" + "b";                    // alloc
}
```

`ok: 1 file(s) typechecked`. That is the correct and intended behaviour of `none:`.

What cannot be said is the inverse: **"this function may do these things and nothing else."** Expressing "may allocate, nothing more" today means enumerating every other class:

```hale
@effects(none: {syscall, block, publish, time, entropy, env, ffi, spawn, recursion})
```

## Why that is a fail-open, not just a nuisance

A `none:` list is fixed at the moment it is written. Add a class to the language and every such contract silently permits it — the annotation still reads "only alloc" and no longer means it. Nothing fails; the certificate just quietly weakens.

**This is already happening.** `@deterministic` desugars to a hardcoded blacklist (`parser.rs:1354`):

```rust
"deterministic" => vec![EffectClass::Time, EffectClass::Entropy, EffectClass::Env],
```

Three classes, fixed at parse time. So `@deterministic` cannot see a **user-declared effect class**. A program that declares `effect wallclock;` and classifies a leaf as carrying it may reach that leaf freely from a `@deterministic` function. Live since #345 shipped, and it is the same shape as the holes closed in #348 and #350: a certificate that reads stronger than it is.

With user effect classes the class set is now **open by construction**, so a `none:` list can never be exhaustive even in principle. That is what makes the missing dual structural rather than cosmetic.

## The asymmetry

| clause | form |
|---|---|
| `publish: {…}` | complete declaration — publishing outside the set is a violation |
| `causes: {…}` | complete declaration — "omitting one is the error" |
| `depends: {…}` | complete declaration |
| `none: {…}` | open-ended blacklist |
| **missing** | **closed allow-list over classes** |

The three complete forms are the future-proof ones precisely because they are closed.

## Proposal

```hale
@effects(only: {alloc})
fn decode(dg: Datagram) -> Msg { … }
```

[...]

Issue #353, opened 2026-08-03
Missing language and stdlib features: closures, str::split, time formatting, regex, and an unvalidated stdlib namespace in hale check


rileyr:
Survey of language and stdlib features Hale does not have. Everything below was verified against `HEAD` (4f260d9) with `hale build` as the oracle — **not** `hale check`, for a reason that turns out to be item 9.

The pattern, stated up front: Hale has gone very deep on the *architecture* layer — effects, placement, bus causality, supervision, verification — and comparatively shallow on the *ordinary program* layer. The gaps below are unglamorous, but they are what you hit in the first hour of writing a real handler: split a string, format a timestamp, map over a list.

---

## 1. Closures / lambdas as values — the one that is upstream of the rest

`Expr` has no closure variant. `let f = |x: Int| -> Int { return x + 1; };` is a parse error (`expected expression, got Pipe`).

Named function pointers *do* work end to end:

```hale
fn inc(x: Int) -> Int { return x + 1; }
fn apply(f: fn(Int) -> Int, v: Int) -> Int { return f(v); }
fn main() { println(apply(inc, 1)); }   // builds and runs
```

So higher-order functions exist, but every function must be named at the top level.

This is upstream of several items below. `map` / `filter` / `reduce` / `sort_by` / `retain` / `find` and every callback-style API are unwritable without anonymous functions — the stdlib cannot offer them even if someone wrote the bodies. Ranking this first is not about elegance; it is about unblocking a category.

## 2. `std::str::split` — and the rest of the everyday string surface

The complete `std::str` namespace today: `parse_int`, `parse_float`, `parse_decimal`, `can_parse_int`, `can_parse_float`, `range_parse_int`, `range_eq`, `byte_at_unchecked`, `index_of`, `lower`, `upper`, `trim`, `substring`, `replace`, `repeat`, `pad_left`, `pad_right`, `from_bytes`, `clone`.

Missing: `split`, `join`, `contains`, `starts_with`, `ends_with`.

Worth noting that `lotus_str_contains`, `lotus_str_starts_with` and `lotus_str_index_of` **already exist in the runtime C** — they carry `memory(read)` attributes from the pure-read work. They are simply not exposed through the stdlib registry. That part is plumbing, not implementation.

Splitting a string is arguably the most common single operation in service code.

## 3. Date/time formatting and parsing

The complete `std::time` namespace: `monotonic`, `monotonic_ns`, `sleep`, `now`, `time_from_unix`.

[...]

Issue #334, opened 2026-08-01
Design: canonical identity for topics and locus instances


rileyr:
Five separate problems found over the last two days share one cause:
the compiler has no canonical identity for a **topic** or for a **locus
instance**, so every analysis that must decide *"are these two
references the same thing?"* rolls its own comparison, and each one is
wrong differently.

This issue is the root cause and the plan. #332 and #333 stay open as
the concrete reproducers; they are symptoms, and both are independently
testable.

## The five

| where | what broke | missing identity |
|---|---|---|
| #331 (fixed) | a library's `publish:` contract was unsatisfiable once imported | topic: `Recalled` / `relay::Recalled` / `__lib_lib_relay_main_Recalled` |
| **#332** | cross-seed delivery silently doesn't happen | topic: one source topic, two disconnected bus-graph nodes |
| **#333** | an aliased locus races across pools, no diagnostic | instance: two field declarations, one object |
| `depends:` | shared-state influence is not in the closure | instance: no node to hang the edge on |
| v0.12.0 hole 3 | contracts through interface slots were vacuous | the implementation behind a slot |

Not five bugs that rhyme. One concept surfacing wherever an identity
decision is required.

## Functions already have this, and it is why they work

`FnKey` is a canonical function identity, introduced with the callgraph
witness engine. The effect system works — `@no_syscall`, `causes:`, the
witness paths — precisely because every reference to a function
normalizes to one key before anything is compared.

Topics and instances have no analogue, so each analysis improvises:

- `topic_ref_matches` / `topic_tail` (added in #331) normalizes three
  spellings by comparing trailing names, and ships with a documented
  limitation: **two topics with the same trailing name from different
  seeds are indistinguishable.** That caveat exists only because there
  is no `TopicId`. With one, the matcher is `==` and the limitation
  does not need documenting because it does not arise.
- F.31 (`spec/types.md`) resolves *"the pool of the field instance"* by
  reasoning per **field declaration**, which is why #333 races: each
  worker correctly concludes it owns its own field, and nothing relates
  the two declarations to the single object they both name.
- The bus graph keys subjects by string, which is why #332 has the same
  topic present twice under spellings that never unify.

## Why this is load-bearing, not cosmetic

Every guarantee in the verification story derives from the closed world
— closed call graph, closed topic set, finite classified frontier.

But "closed" buys nothing unless the elements can be enumerated **and
told apart**. A closed set whose members have no canonical identity is
[...]

PR #386, opened 2026-08-04
claims phase 1: groups, named forbid-reaches claims, witnesses (#382)

rileyr:
**⚠️ Landing procedure (unitary — updated per follow-up review):** the previously stated "merge #386 first, then #387 immediately after" was operational discipline, not atomicity — #386 alone would still put a known false-certifying phase-1 implementation on `main`. The genuinely unitary order is:

1. **First merge #387 into this PR's branch** (`gh382-claims-phase1`) — this PR then carries the complete stack, including the projection-vacuity fix and the unresolved-receiver backstop;
2. **then merge this PR into `main`.**

(Equivalently: squash the stack into one landing. Either way, no known-unsound commit ever enters `main`.)

Phase 1 of #382: **groups + named `forbid reaches` claims + witnesses**, per the build order ("falsification first") and the grounding audit / proposed rulings on the issue thread.

## What lands

**Surface** (all contextual keywords; zero lexer changes):

```hale
group delta_wing = { delta::*, DeltaStore };
group gamma_wing = { gamma::Research };
group probes     = { } may_be_empty;

main locus Org {
    claims {
        iso_dg:   forbid reaches(delta_wing, gamma_wing);
        iso_call: forbid reaches(delta_wing, gamma_wing) via { calls };
        no_spend: forbid reaches(gamma_wing, effects(money));
    }
}
```

**Semantics** — the rulings from the issue thread, implemented:

- **Groups name declarations; claims project them; evaluation is fn-grained.** A locus member projects to all of its methods/lifecycle/modes — projection only ever *adds* sources and sinks (the conservative direction). Vacuity is judged at the decl grain.
- **Declaration-required everywhere.** Unknown group member, unknown group in a claim, undeclared effect class in `effects(...)` → errors with did-you-mean, never empty sets. Empty group → vacuity error unless `may_be_empty`. Duplicate claim names → error (the name is the contract-of-record).
- **Cross-seed members resolve at the mangle stage** (the #334 path, never name-suffix matching): `alias::Name` canonicalizes in `QualifiedRenameApplier`, `alias::*` enumerates the imported seed's decls via the same rename table codegen uses, and witnesses demangle back to author spelling. `ClaimSet::Effects` user-class indices ride `remap_user_effects` so merged tables can't alias.
- **Unknown ⇒ violation.** Indirect calls (#353) and computed publish subjects on a path from a `forbid` source fail closed, exactly as `@no_syscall` treats the same shapes.
- **Errors gating `check`, day one. No advisory tier.** An advisory claim reads as law and doesn't bind (the #354 fail-open shape). Errors also pass `retain_owned_advisories` unconditionally, so future library-tier claims survive app builds.
- **`claims { }` is main-only** (parse error elsewhere, like `bindings`/`placement`): main is the closed-world gate, so bundle-wide claims cannot be evaluated anywhere earlier.

**Witness** (one minimal countermodel per claim):

```
claim `iso_dg` violated: `delta_wing` reaches `gamma_wing` —
witness: `delta::Triage::on_task` -(publishes "org.metrics")-> `gamma::Research::on_metric`
```

## Implementation

- `hale-syntax`: `TopDecl::Group`, `LocusMember::Claims`, parser productions (named-entry block copied from the `bindings` shape), keyword listing + regenerated highlighter blocks. The formatter needed **zero work** (token-stream engine).
- `hale-types/src/claims.rs` (new): group resolution + validation + BFS evaluator over the composed call ∘ bus graph, reusing the same substrate as every shipped certificate (`summarize_with_stdlib_and_renames`, the bus graph's subscriber lists incl. wildcard coverage, `class_mask_with` for composed classes). Wired into `check_bundle_opts` beside `causes_diags`/`depends_diags`.
- Codegen: claims lower to **zero code** (same story as `placement`); mangle/codegen matches get explicit no-op or rewrite arms.

## Tests (negative controls per soundness law 4)

- `hale-types/tests/claims.rs` — 16 tests: bus-path violation with full witness (the canary), isolated-wings control, call-path violation through a handle method, `via {calls}` ignoring a bus path, `effects(...)` sink violation + avoidance control, vacuity/may_be_empty, unknown member/group/class with did-you-mean, duplicate names, main-only parse error, trailing-only glob, indirect-call fail-closed.
- `hale-cli/tests/xseed_claims.rs` + `fixtures/xseed-claims/` — the flagship two-wing shape across seed boundaries: `delta::*` glob + qualified member resolve cleanly, the isolation claim fires across the boundary, and the witness is fully demangled (no `__lib_` leaks).

## Spec & docs (same-change per repo convention)

`spec/grammar.ebnf` (`group_decl`, `claims_block` + `top_decl`/`locus_member` arms), `spec/verification.md § Claims`, `spec/tokens.md`, `docs/src/verification.md` (new section with a parse-gated snippet), `CHANGELOG.md`.

## Deliberately deferred (tracked on #382)

`only edges` grants + the topology artifact (phase 2 — blocked on the shape_hash ruling), indexed effect families + `bound` over user classes (phases 3–4), `require`/`cover`/`during`/`avoiding` (phase 5), library-tier claims that travel with imports, the `model.rs` schema layer (this PR evaluates directly over the existing summaries; the schema layer lands with the artifact so it serializes what evaluation consumed). `tree-sitter-hale` is a separate repo and already lags `target`; `group`/`claims` go in its next sync.

Closes nothing; first deliverable of #382.


rileyr:
Responding to the external QA review (phase-1 half). Disposition of the three follow-up concerns:

**1. Projection vacuity — accepted, fixed on the phases-2–5 branch.** The reviewer is right that decl-grain vacuity and relation-projection vacuity are different properties, and the gap was real: a group of fn-less loci (pure-data stores) passed the decl-grain guard but projected to zero executable vertices, so a `forbid` over it held while proving nothing — a fail-open wearing the vocabulary's clothes. Now an **error** at either endpoint of `forbid` / `only edges` and at `bound`'s `from`, with the suggested phrasing: *"group `x` projects to no executable source vertices — its declarations have no fns, so the claim proves nothing about them."* The diagnostic also says why (the fn-grained walk cannot see pure-data access) and what to do (name the loci that hold the behavior). Tested at all three positions.

**2. Witness provenance — agreed; it's the normalized-model milestone.** Per-edge spans (callsite, publish site, subscription, destination decl) are exactly what the provenance-bearing model layer is for, and "which source decision introduced it and where to edit next" is the right bar for the witness to aim at. Deliberately not bolted onto the current path-of-names rendering: provenance added ad hoc per evaluator is how three subtly different span conventions happen.

**3. Convergence onto one model — agreed, with sequencing on purpose.** Phase 1 evaluates directly over `AllocSummary` + `BusGraph` by design (falsification first, per the promotion doctrine on #382: prove the claim IR against real use before freezing the schema it implies). The review's recommendation — normalized, provenance-bearing model consumed equally by annotations, claims, diagnostics, and the artifact as the *next* architectural milestone — matches the plan of record exactly, and the concern about evaluators accumulating subtly different meanings of edges/unknowns is the strongest argument for doing it before more verbs land. Tracked on #382.

rileyr:
**⚠️ Merge only together with #387.** The follow-up review correctly notes this branch, taken alone, contains neither the projection-vacuity fix nor the unresolved-receiver fail-closed backstop — both of which affect phase-1 `forbid reaches`. Those fixes are implemented against the phases-2–5 rewrite of `claims.rs` (the shared evaluator context, outcome plumbing, and artifact hooks), so backporting them here would mean maintaining two implementations of each and force-rebasing the stacked PR mid-review.

The resolution: this PR and #387 land as a unit — merge this base first, then #387 immediately after, so no commit on `main` is ever a known false-certifying implementation. If review concludes otherwise (e.g. #386 should merge alone), say so and the two soundness fixes get backported here first.

PR #387, opened 2026-08-04
claims phases 2-5: only-edges grants, topology artifact, indexed families, bound, require/cover/count/during/avoiding (#382)

rileyr:
Phases 2–5 of #382, stacked on #386 (phase 1). The full verb set from the issue's build order — plus the two forms promoted by the surface-completeness audit (`avoiding`, `count`) — with per-phase negative controls throughout.

## Phase 2 — `only edges` + the topology artifact

- **`only edges A -> B { publish T; subscribe T; }`** — every DIRECT A→B edge must match a granted line; every un-granted edge is reported (the grant list is the review surface, so the full diff matters). `publish`/`subscribe` admit the same edge — the verb names which end's declaration is the reviewable line. Call edges are never grantable. A third-party sink outside B (the `log.**` shape) is not an A→B edge — the infra-wildcard question resolves to "no carve-out machinery needed at the boundary form".
- **The topology artifact** — `hale check <t> --dump-topology` emits the serialized model (sorts, relations in author spelling) + every named claim's normalized form and result, under `"schema": "1.0"` and a `shape_hash` (FNV-1a/64 over the canonical model half; claims excluded so one topology under different law keeps one shape). `--check-topology <baseline>` diffs and fails with a regenerate hint — the `.hale.effects` precedent. Compiler-side `shape_hash` is deliberately namespaced away from the runtime per-topic hash; reconciliation with iris's fusion key stays tracked on #382.

## Phase 3 — indexed effect families + `@budget(<user class> = N)`

`domain wing = { delta, gamma };` + `effect knowledge(wing);` — every instantiation `knowledge(delta)` interns as an ordinary declared class, `knowledge(*)` as an auto-populated composed class. The whole feature is the reduction the issue describes: masks, `only:` complements, cross-seed remap, and did-you-mean all work unchanged, so a misspelt index is an undeclared-class error (`Did you mean \`knowledge(delta)\`?`) and a domain member added later lands outside every existing contract. One v1 rule: the domain must be declared earlier in the same file as the family (what lets the parser expand eagerly and keep everything inside shipped machinery). Companion: `@budget(llm = 1)` — a new `QuantDim::UserClass` counting calls to declared carriers, with the standard loop/indirect unboundedness rules and declaredness checked with a hint.

## Phase 4 — `bound` claims

`bound llm <= 1 on paths from planners;` — `@budget`'s per-call semiring behind the claims surface: total carrier sites per invocation is a call-tree **sum** (two calls to a carrier are two sites — deliberately `@budget`'s semantics, not a longest-path max, which undercounts). Cycle / loop-nested carrier / indirect call / computed subject ⇒ unbounded ⇒ violated. The witness carries the count and a representative chain. Built-ins keep their `@budget` spellings (`bound alloc` is an error pointing there).

## Phase 5 — `require` / `cover` / `count` / `during` / `avoiding`

- `require subscribes|publishes(some G, topic T)` — existence over the declared bus ends.
- `cover topic in seed(a): subscribed_by(some G)` — bounded universal over an imported seed's declared topics; every uncovered topic named; an alias with no topics is an error (empty coverage domain = vacuity).
- `count publishers|subscribers(topic T) ==|<=|>= N` — distinct loci; `== 1` is single-writer, and a violation names the competing writers.
- `during <phase>` — source roots restricted to the named lifecycle phase/method; a phase naming nothing in the group is an error, not a vacuous pass.
- `avoiding <group>` — vertex mask; the interposition form ("every path passes the permission gate").

Qualified topic references in claims (`topic t::Tasks`, grants) canonicalize at the mangle stage alongside bus subjects — the applier gained a Claims arm — and everything demangles in witnesses and the artifact.

## Tests (canary + control per form, per the user's standing ask)

- `hale-types/tests/claims_phases.rs` — 27 tests: un-granted edge canary + granted/subscribe-spelled controls + call-edge + third-party-sink + unknown-grant-topic (phase 2); family propagation/aliasing/star/misspelt-index/family-without-domain/claim-sink + budget canary/control/loop/undeclared (phase 3); bound canary/control/loop/built-in-rejection (phase 4); require both verbs both ways, count single-writer canary + control, during birth canary + run-phase control + empty-phase error, avoiding interposition hold + bypass canary (phase 5).
- `hale-cli/tests/xseed_claims_verbs.rs` + fixture — qualified refs and `cover` across a seed boundary (uncovered topic named in author spelling, no `__lib_` leaks), artifact round-trip (`--check-topology` passes on a fresh dump, fails with the regenerate hint on a stale one), shape_hash stability.
- All 19 phase-1 tests unchanged and green; full workspace suite green.

## Spec & docs (same-change)

`spec/grammar.ebnf` (all six verb productions + `domain_decl` / `effect_family_decl`), `spec/verification.md § Claims` (the full verb set + families + artifact), `spec/tokens.md`, `docs/src/verification.md` (extended with a fragment showing the whole family), CHANGELOG, regenerated highlighter keyword blocks.

## Deferred (tracked on #382)

Library-tier claims that travel with imports; annotations lowering onto the claim IR (§8 — after the IR survives real use, per the promotion doctrine); `during` over user classes in `@phase_effects`; the iris shape_hash reconciliation; grants for call edges (deliberately never).


rileyr:
Responding to the external QA review (phases-2–5 half), point by point:

**1. Artifact insufficiency — accepted; partially fixed now, remainder scoped honestly.** The artifact now additionally serializes **groups** (the claim vocabulary), **labels** (declared effect carriers, i.e. the `is:` frontier), and **unknowns** (fns with indirect calls or computed publish subjects — exactly the places evaluation failed closed). With those, the reachability-class claims (`forbid reaches`, `only edges`, `require`/`cover`/`count`) are independently re-evaluable from the file, and an auditor can see where certification stopped. What remains missing — per-edge spans, weights, the phase relation, a seed sort — is now stated in the module header and spec as the v1 contract: *a topology + claims report whose passing rows are certified by the derivation*, with the complete normalized-model export as the tracked milestone. The reviewer's either/or (narrow the contract or serialize everything) is answered with "narrow honestly now, serialize completely at the model milestone."

**2. shape_hash two-identities — accepted and resolved by widening the one hash.** Groups, labels, and unknowns live *inside* the hashed model half, so a carrier added, a group widened, or a new fail-closed site changes the shape identity. Claim *results* stay excluded (one topology under different law keeps one shape — that separation is load-bearing for the diff semantics). The residual exclusions are the same items missing from the model, documented in the same place.

**3. `only edges` grant synonymy — defended as-is, flagged for the fmt ruling.** The two spellings are deliberate documentation, not semantics: the verb names which end's declaration a reviewer should go read (`publish T` says "read the source wing's publish"; `subscribe T` says "read the target wing's subscription"), and the issue's own reference examples use both. The review is right that vocabulary which *looks* semantic but isn't carries risk — the natural place to resolve it is `hale fmt` canonicalization (one spelling normalized, the other accepted), which can land without breaking either form. Deferred to that ruling rather than collapsed now.

**4. `bound` wording — the semantics is the intended one; the spelling is the issue's.** It is per-invocation aggregate cost (`@budget`'s call-tree sum — two calls to a carrier are two sites), stated explicitly in the spec and pinned by the diamond test; the longest-path reading undercounts exactly the case the claim exists for. "on paths from" is #382's surface grammar. If it misleads in practice, the surface can change cheaply before promotion freezes it — noted on the issue.

**5. `during` as name-matching — agreed, and documented as the first implementation.** The phase relation belongs in the normalized model; the current reduction (phase = the named lifecycle hook / method per source locus) is the version that ships with a canary and a control today. The spec flags it as such rather than as the definition.

**6. Claim failures as a guide — agreed as the north star.** "Capability introduction point, boundary crossed, best completion site, smallest authority-preserving edit" all require provenance-rich edges — one more consumer of the same model milestone, which is now the clearly-indicated next step from three independent directions (witnesses, artifact, DX).

On the overall recommendation (merge the phase-1 proof, keep indexed effects, make the normalized model the next milestone): that matches the plan of record on #382.

rileyr:
**Adversarial soundness audit result + fix.** A targeted audit of the one path no review had independently verified — `Callee::Unresolved` handling in the claims evaluators — found a **confirmed false-certificate class**. Four receiver shapes land in the call graph as unresolved edges with `recv_ty: None`, and the walk silently dropped them:

1. struct-literal receiver — `B { }.work(n)`
2. chained field — `self.mid.inner.work(n)`
3. call-result receiver — `let b = make_b(); b.work(n)`
4. branch-valued receiver — `let b = if … { B { } } else { B { } }; b.work(n)`

For each, `forbid reaches(a_side, b_side)` certified while `hale run` executed the forbidden path. The audit also established this is a **shared pre-existing gap**, not a claims regression: `@effects(none: {money})` misses the identical shapes, because claims inherited the effects predicate (fail closed only on `indirect`/fn-param) verbatim, and the root cause is the summarizer's `receiver_type_name`, which types exactly two receiver shapes (bare ident via `var_types`, `self.<field>` via `locus_param_field_types`).

**Fix in this PR — the claims-side backstop**: an unresolved, untyped-receiver call whose *name* matches a method declared in the claim's target set now fails closed, in all three walks (`forbid` targets, `only edges` targets, `bound` carriers), with an actionable diagnostic ("bind the receiver to a typed field or local so the call resolves"). Design points:

- Keyed on the target set's method names, not on every unresolved edge — an unresolved call to a name outside the target set still certifies (pinned by a control test), so `println`/builtin edges can't trip it.
- `recv_ty: Some` edges (synthesized form/builtin methods — `counts.set(x)`) are exempt: they're known non-locus receivers, so existing certificates are unaffected.
- Over-approximates in the safe direction only: it can demand a receiver be typed when the real target was elsewhere; it cannot false-certify.

Six regression tests: the four audit shapes, the non-matching-name control, and `only edges`/`bound` coverage.

**Explicitly out of scope here**: the summarizer root fix (typing the four receiver shapes, which repairs the effect system's identical miss — `@no_syscall` et al. on the same shapes). That changes shipped certificate behavior across the corpus and the committed `.hale.effects` baselines, so it belongs in its own change; flagged on #382 as the follow-up, with the audit's suggested approaches (syntactic typing of the four shapes, or feeding the walker the typechecker's inferred receiver types).

rileyr:
Responding to the follow-up review. The blocking finding is correct and is now fixed at head, using the review's own "simpler and safer" option.

**1. The transitive hole — confirmed and fixed.** The wrapper counterexample reproduced exactly as predicted (`Bridge { }.hop(n)` → `Bridge::hop` → `B::work`; `hale check` printed `ok`). The review's diagnosis is the right one: no name comparison against the target set is sound, because the unresolved edge may be a wrapper whose name matches nothing — the edge itself is the uncertainty. Implemented per the review's first option:

- `CallEdge` gains `receiver_present: bool`, set at the single construction site. A bare-`Ident` receiver that names an import alias (`pay::charge(n)` parses as a qualified call) is excluded; `self` is excluded; the catch-all `<expr>` callee arm is conservatively *included*.
- Every judgment that traverses calls — the `forbid` BFS, `only edges`, `site_count` — now fails closed on `receiver_present && recv_ty.is_none() && unresolved`, with the same actionable diagnostic. The name-keyed sets are deleted, not augmented.
- The backward-slice alternative was rejected for exactly the reason the review gave: it's a parallel evaluator waiting to drift. The receiver-typing root fix in the summarizer remains the real repair, tracked on #382.

All four suggested tests are in: `an_untyped_receiver_wrapper_reaching_the_group_target_fails_closed`, `…_reaching_an_effect_carrier_fails_closed`, `…_reaching_a_bound_carrier_fails_closed` (each from the review's counterexample programs), plus the artifact pair below. The previous name-match control test now asserts the *opposite* — an unresolved name outside the target set fails closed too, with the wrapper rationale in its doc comment.

**2. Artifact misses the new unknown class — fixed.** `unknowns` now records `untyped_receiver_call:<callee>` per fn, inside the hashed model half — so introducing `Bridge { }.hop(n)` adds an unknown row and changes `shape_hash` (pinned by `an_untyped_receiver_edge_changes_shape_hash`, which diffs the hash between the typed and untyped variant of the same program, and `the_artifact_records_the_untyped_receiver_unknown` with a fully-typed negative control). The callee name is included precisely so an outside evaluator can apply the same fail-closed rule.

**3. `topology.rs` header contradiction — fixed.** The opening paragraph no longer claims universal third-party re-evaluation; it points at the v1 SCOPE paragraph, and the two now agree.

**4. #386 stacking.** Acknowledged — see the note on #386. Short version: the soundness fixes are implemented against the phases-2–5 rewrite of `claims.rs` (the shared `Cx`, outcome plumbing, artifact hooks), so backporting them onto the phase-1 module means reimplementing them twice and force-rebasing this PR; instead #386 now carries an explicit **merges-only-with-#387** marker. The review's underlying requirement — that no commit on main is ever a known false-certifying implementation — is preserved: the stack lands as a unit.

rileyr:
Both remaining points from the second pass are addressed at head (`bff1bd8`):

**1. Landing procedure** — the #386 banner now prescribes the genuinely unitary order the review laid out: merge **#387 into `gh382-claims-phase1` first** (so #386 carries the complete stack), **then** #386 into `main`; or squash the stack into one landing. The "merge #386 first, then #387 immediately after" wording is retracted for the reason given — it was discipline, not atomicity.

**2. Artifact wording** — the v1-scope paragraph (module header + spec) now makes the narrower claim verbatim: independent replay covers the serialized user call/bus graph, group boundaries, declared bus-end existence/cardinality, and declared user-effect carrier labels; everything needing the phase relation (`during`), seed membership (`cover`), compiler-derived built-in effects, or the stdlib-expanded summary remains a compiler-certified report row until the normalized model lands.

**3. On the effects-side receiver typing being the immediate next soundness change** — agreed, including the inconsistent-judgments framing (a claim failing closed on a path that `@effects(none:)` still certifies is a state to exit quickly). Elevated on #382 with that example; it's the next change after this stack lands.

Issue #408, opened 2026-08-05
Fleet claims: compose topology artifacts across multiple main-locus applications


rileyr:
## The third horizon

Claims today have two owners, and both are sound because each closes a world:

```
library seed   closes over its own boundary; law travels with the import
main locus     closes one application / executable
```

A system built from several independently compiled binaries has laws that live above both, and today nothing can state them:

- traffic from one domain must not reach a privileged sink except through a designated mediator;
- exactly one deployed process may publish an authoritative channel;
- an event must be routed to *both* of two required consumers;
- an operator surface may command a service but must not acquire that service's authority directly;
- must-arrive traffic must not be configured over a transport whose contract permits loss.

Every binary can be locally correct while the assembled system is wrong. Recursively checking every seed in a repository proves every local claim and says nothing about whether the deployed binaries are connected in the intended arrangement.

The missing feature is not "recursive `hale check`". It is a **deployment closure**.

This is a **downstream handoff**: the proposal came from a multi-binary production system whose shape makes the gap concrete — many independent mains, shared topic declarations imported across binaries, route selection supplied outside the source through bus-config files, explicit connect/listen roles, and a hand-maintained topology document mirroring those route files.

## Proposed ruling

> `main locus` closes one application. A **fleet plan** closes one deployment of applications. Each closure emits a typed graph artifact; the next closure **composes those artifacts rather than reopening their source**. Claims belong to the owner of the horizon they quantify over.

The generalization worth stating explicitly, because it also settles the dynamic-assembly question: **a closed world is not the whole repository or the whole fleet. It is a stage boundary at which a graph becomes exact enough to certify.** Many mains are not competing roots — they are sealed graph fragments that become typed components at the next closure stage.

Corollary: *policy may be dynamic, validity is not.* Configuration and machine state may be dynamic while a plan is being constructed; claims evaluate once the plan is exact. Arbitrary runtime graph mutation outside a checked plan invalidates the certificate.

## Why not source-merge every main

The tempting implementation is a "super-main" that imports every application and runs ordinary claims over the merged source. It should be rejected, and the reasons are soundness reasons rather than taste:

- **It invents edges.** An unbound topic is in-process by default, so merging two binaries makes matching publishers and subscribers appear locally connected when no deployed route exists.
- **It erases edges.** Real routes supplied by deployment config, service discovery, or a broker do not appear merely because source was merged.
[...]

PR #410, opened 2026-08-05
fix(claims): never emit an artifact for a program that fails to typecheck

rileyr:
Found while answering an open question on #408. Verified against shipped behavior before the fix.

### The bug

```
$ hale check both.hl --dump-topology     # both.hl has a type error
exit=1
artifact bytes: 771
claims: [{"name": "wired", "form": "require subscribes(...)", "result": "holds"}]
```

A program that doesn't typecheck emitted a **full** artifact — populated relations, and claims evaluated over a graph derived from source the compiler couldn't understand. The claim reports `"holds"` for a program that cannot compile: a certificate asserting a property of something that will never run.

That's worse for a consumer than emitting nothing, because it fails **open**. An admission step looking for "no violated claims" *passes* it — there are none. Cross-artifact composition (#408) would have trusted it.

### The rule

**The artifact's existence means the model is sound.** A program that doesn't typecheck emits no artifact and says why:

```
refusing to emit a topology artifact: `both.hl` does not typecheck, so its
model is not a truthful description of any program. Fix the type error first.
```

A **violated claim** is the opposite case and still emits, unchanged. The model is well-defined there, the row is a truthful report, and being able to replay a violation independently is the whole point of publishing the model — `"result": "violated"` would be dead vocabulary otherwise. Both directions are pinned by tests.

### `DiagKind::Claim`

Separating the two cases needs a way to ask "does this program typecheck?" when claim violations are themselves errors in the same stream.

The new variant is an error like any other and renders **identically** to `Type` — the message already begins "claim `x` violated", so a distinct prefix would only stutter, and no existing diagnostic output changes. It's applied at the single funnel every claim diagnostic passes through (`claims_report`, which already post-processes the whole batch to demangle cross-seed symbols) rather than at its ~30 construction sites.

It's independently useful: tooling can now categorize claim diagnostics without matching on message text.

### Cost

`check` previously ran `check_bundle_opts` once; a naive gate would have run it twice on every `--dump-topology`. Since nothing mutates the bundle between the two points, the analysis is computed once and shared between the artifact gate and the diagnostic report — so the gate is free.

### Verification

| program | artifact | exit |
|---|---|---|
| type error | **none** | 1 |
| claim violated | emitted, 1028 B, claim rows present | 1 |
| clean | emitted, 1156 B | 0 |

Plain `check` (no dump) is unaffected in all three.

Ran the affected binaries rather than the full suite: `topology_artifact_contract` (8), `xseed_claims_verbs` (5), `check_arg_parsing` (6), `hot_factory_xseed` (2), `artifact_integrity` (6), `claims` (16), `claims_edges` (28), `claims_phases` (27), `library_claims` (2), plus the three that filter on `DiagKind` — `effects_conformance`, `instance_aliasing`, `lsp`. All green; CI has the rest.

Closes the first item under "Resolved since filing" in #408.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,40 @@ the conformer from the group.
 The fact was already in the model (the artifact tags these edges
 `via_interface`); it just never reached the human. Witnesses with no
 interface in them are unchanged.
+### Topology artifact schema 1.4: one verdict vocabulary, and the document's own verdict
+
+Bundle claims and fn-grained certificates (`@effects`, `@budget`,
+`@phase_effects`) are the same kind of statement at different
+granularity — #392 §8 already reports the second as the claim form it
+is pointwise sugar for. They disagreed about how to spell an outcome:
+claim rows carried three states, `lowered` rows carried a bool.
+
+Both now use `Verdict`, and it distinguishes a case that was
+previously folded away:
+
+| verdict | meaning | repair |
+|---|---|---|
+| `holds` | proved | nothing |
+| `violated` | disproved — a counterexample exists | fix the program |
+| `uncertified` | well-formed but not provable — the graph has an unknown | resolve the unknown edge |
+| `invalid` | the statement is malformed | fix the claim |
+
+`violated` and `uncertified` were one value because **unknown ⇒
+violation** — an indirect call fails closed rather than certifying an
+absence nothing established. That rule is unchanged and both still
+fail the build. What changes is that the artifact records which
+happened: the repairs differ, and composing models across binaries
+needs a propagated unknown to read as "not provable" rather than as
+disproved when nothing disproved it.
+
+The artifact also gains a top-level **`verdict`** (`clean` /
+`law_failed`) — every law reduced to one field, so a consumer does
+not reconstruct it by walking two arrays and knowing which strings
+count as passing. Only `holds` passes: a law that could not be
+checked has not been satisfied. It says nothing about whether the
+program typechecks, because an artifact is only emitted for one that
+does.
+
 ### A topology artifact is only emitted for a program that typechecks
 
 `hale check broken.hl --dump-topology` emitted a **full artifact** for
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,34 @@ the conformer from the group.
 The fact was already in the model (the artifact tags these edges
 `via_interface`); it just never reached the human. Witnesses with no
 interface in them are unchanged.
+### A topology artifact is only emitted for a program that typechecks
+
+`hale check broken.hl --dump-topology` emitted a **full artifact** for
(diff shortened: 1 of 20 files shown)


rileyr:
Pushed the two Phase 0 schema items from #408 onto this branch, since they're the same area and the artifact-soundness rule is what makes the second one clean.

### One verdict vocabulary

Claims and lowered certificates disagreed about how to spell an outcome — three states vs a bool. Both now use `Verdict`, which also splits a case that was previously folded away:

| verdict | meaning | repair |
|---|---|---|
| `holds` | proved | nothing |
| `violated` | disproved — a counterexample exists | fix the program |
| `uncertified` | well-formed but not provable — the graph has an unknown | resolve the unknown edge |
| `invalid` | the statement is malformed | fix the claim |

`violated` and `uncertified` were one value because **unknown ⇒ violation** — an indirect call fails closed rather than certifying an absence nothing established. That rule is unchanged and both still fail the build. What changes is that the artifact records *which*, because the repairs differ and because composition needs a propagated unknown to read as "not provable" rather than as disproved when nothing disproved it.

Verified end to end — a claim traversing a call through a fn-typed param:

```
claim `iso` cannot be certified: `Src::go` (reachable from `srcs`) calls through
a function-typed parameter, whose target is not knowable statically.
```
```json
"claims": [{"name": "iso", "result": "uncertified"}], "verdict": "law_failed"
```

Previously that row read `violated`, which said something the compiler had not established.

### The document's own verdict

Top-level `verdict`: `clean` / `law_failed`. Composition needs "did this component pass?" as a precondition, and reconstructing it by walking two arrays *and* knowing which strings count as passing is exactly what a consumer gets subtly wrong. `Verdict::passed()` is the one definition, and only `holds` passes — `uncertified` does not, because a law that could not be checked has not been satisfied.

It deliberately says nothing about whether the program typechecks. It doesn't need to: with the gate in this PR, an artifact is only emitted for a program that does, so its existence already carries that. The two changes compose — which is why they're together.

### Note on the refactor

`ClaimOutcome.result` and `LoweredCertificate.violated` had exactly two readers between them, both in `topology.rs`, so retyping was contained. One regex pass rewrote text inside a `format!` string (`"effect assertion violated: …"`, which `strip_prefix` downstream depends on) — caught by the compiler and restored, and I checked for other stranded rewrites.

`serde_json` is now a dev-dependency of `hale-types` so these tests parse the whole document rather than asserting on substrings — the only shape of test that catches a structural error, which is how the unclosed array shipped in the first place.

Suites run: `artifact_integrity` (9), `claims` (16), `claims_edges` (28), `claims_phases` (27), `library_claims` (2), `topology_artifact_contract` (8), `xseed_claims_verbs` (5), `effects_conformance` (3). Schema references in the spec, the claims chapter, and the tests move to 1.4.

PR #412, opened 2026-08-05
fix(claims): provenance for only-edges and bound (downstream review)

rileyr:
Item 4 from the claims developer-experience review. **Stacked on #411** — both touch `claims.rs`, so this targets that branch and will retarget to `main` once #411 merges.

### `only edges`

The review's phrasing: it "correctly reported every ungranted crossing in one invocation" but "both errors were anchored only at the claim declaration. Since `only edges` is explicitly a reviewable boundary inventory, it should have at least the same provenance quality as `forbid`."

Now:

```
onlyedges.hl:29:9: claim `boundary` violated: un-granted edge `Ops::act`
  -(publishes "Sneaky")-> `Core::on_sneaky`. Granted: `Allowed`. …

onlyedges.hl:11:9: claim `boundary`: the un-granted publish happens here
            Sneaky <- c;
            ^^^^^^^^^^^^
onlyedges.hl:17:44: claim `boundary`: received here. Grant this edge with
                    `publish Sneaky;` if it is intended
        bus { … subscribe Sneaky as on_sneaky; }
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

The receiving diagnostic carries **the exact grant line to add**, since "name it in the grant list" is only actionable if you know the spelling. An un-grantable call gets its own site plus why a grant isn't the fix there.

### `bound`

The review: "The checker knows enough to classify the condition. It should attach the carrier site and the construct that made it unbounded."

It did know — and printed all four possibilities:

> a recursion cycle, loop-nested carrier, indirect call, or computed publish subject makes the count unbounded

Now:

```
claim `one` violated: paths from `planners` carry an unbounded number of
`llm` sites (limit 1) — a carrier is reached from inside a loop in
`Planner::go`, so it repeats per iteration

bnd.hl:10:22: claim `one`: this is the loop-nested carrier
                self.n = model_call(i);
                         ^^^^^^^^^^^^^
```

Verified for all four conditions:

| condition | names | points at |
|---|---|---|
| loop-nested carrier | the fn holding the loop | the carrier call |
| recursion | the fn reachable from itself | — |
| unfollowable call | the fn making it | the call |
| computed publish subject | the publishing fn | the publish |
| step ceiling | — | — |

Recursion and the step ceiling are properties of a *walk* rather than of one expression, so they name what they can and omit the caret rather than pointing somewhere arbitrary.

### Implementation

The classification was always computed — reaching the verdict requires it — and then discarded. `Heaviest` becomes `Result<(u64, Vec<FnKey>), Unbounded>` instead of an `Option`, and the `Unbounded` variants carry the fn and span. That's the whole change; no new analysis.

Secondary spans follow the same bundle-only rule as the `forbid` witness — a stdlib body parses in its own offset space, so a span from there attributed to a bundle file would name the wrong source.

Six tests in `claims_edges`, including a guard that the four-way disjunction string is gone. Ran `claims_edges` (36), `claims` (16), `claims_phases` (27), `library_claims` (2), `budget_contract` (9), `xseed_claims_verbs` (5), `claims_artifact_unknowns` (4).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,46 @@ behavior.
 
 ## Unreleased
 
+### `only edges` and `bound` say where to edit (downstream review)
+
+`forbid` names the crossing call, or the publish and the receiving
+subscription. The other two path claims anchored only at the claim
+line, which for `only edges` defeats the point — it is explicitly a
+*reviewable boundary inventory*, and making the reviewer hand-find
+the crossing is the work it exists to save.
+
+**`only edges`** now points at the un-granted publish, at the
+subscription that receives it (with the exact grant line to add), and
+at an un-grantable call — with why a grant is not the fix there.
+
+**`bound`** knew which of four conditions made a count unbounded and
+printed all four:
+
+> a recursion cycle, loop-nested carrier, indirect call, or computed
+> publish subject makes the count unbounded
+
+Now it names the one that applies and points at it:
+
+```
+claim `one` violated: paths from `planners` carry an unbounded number of
+`llm` sites (limit 1) — a carrier is reached from inside a loop in
+`Planner::go`, so it repeats per iteration
+
+bnd.hl:10:22: claim `one`: this is the loop-nested carrier
+                self.n = model_call(i);
+                         ^^^^^^^^^^^^^
+```
+
+The classification was always computed — reaching the verdict
+requires it — and then discarded. Keeping it costs one enum on the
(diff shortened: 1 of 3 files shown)


rileyr:
Superseded by #414 — this PR's head was stuck on a pre-rebase commit after its original base branch was deleted, so `pull_request` CI stopped firing on it. Same branch, rebased onto current `main`, reopened as #414.

Issue #409, opened 2026-08-05
Constitutions: share one claimset across many entrypoints, composed per environment


rileyr:
## The problem

One codebase, many entrypoints. Depending on where it is being deployed, **any** entrypoint must satisfy the claimset for that destination.

Nothing supports that today:

- `claims { }` is only legal inside `main locus`, so a law that should hold for twenty entrypoints must be copy-pasted into twenty main loci — and the copy that gets forgotten fails open, silently.
- `hale check <dir>` is one seed and does not recurse, so there is no way to ask "does every entrypoint in this repository satisfy X".
- There is no notion of a claimset varying by deployment destination, so the dev/staging/production distinction lives in convention or in a shell script.

Note this is **universal quantification over entrypoints, each checked independently**. No routes, no composition, no artifacts. It is deliberately not #408 (laws about how separately-deployed binaries are *connected*), and neither depends on the other.

### A claimset is declarable outside `main`

Today the placement rule is load-bearing for a good reason: claims are closed-world statements, and `main` is the only place the world is closed. That reason constrains *evaluation*, not *authoring*.

So: a **constitution** is a named claimset declared once, outside any main; a `main locus` adopts it; and evaluation still happens in main, against main's closed world. One text, N evaluations, N closed worlds. Nothing about the soundness argument changes.

### Environments compose it additively

The tempting model is "each environment has its own claimset", and it fails in one specific direction. If production has a law that development lacks, then development never exercised production's architecture and the first violation surfaces at the production build — the same fail-open pattern in deployment clothing.

The reverse direction is safe and genuinely useful. A development environment may want *stricter* law than production: "nothing reaches a real external payment provider", satisfied additively by the development entrypoint wiring a stub whose annotations differ from the real provider. Real architecture is still exercised; the extra claim holds because of a membership difference rather than a carve-out.

So the rule is **monotone, not invariant**: an environment may add laws, never drop them. A development constitution is built *from* the core one.

### Union only — no override

This is the rule that makes the rest hold up.

If a derived constitution can override an inherited clause, weakening becomes expressible and reads exactly like ordinary composition at the call site — a "refinement" of `count publishers <= 1` into `<= 3`. Detecting that would require an ordering on claims, i.e. proving one sentence implies another, and getting it wrong fails open.

[...]

PR #415, opened 2026-08-05
feat(claims): constitutions — one claimset across many entrypoints (GH #409)

rileyr:
The language half of #409. The manifest and the entrypoint × environment matrix are deliberately not here — see the bottom.

### The gap

`claims { }` is only legal inside `main locus`, and that rule is load-bearing: claims are closed-world statements and `main` is the only place a world is closed. But it constrains **evaluation**, not **authoring**. A law meant to hold for twenty entrypoints had to be copy-pasted into twenty main loci — and the copy somebody forgets fails open, silently.

```hale
constitution Core {
    tenant_iso: forbid reaches(billing, research);
    one_writer: count publishers(topic Settled) == 1;
}

constitution Dev extends Core {
    no_real_payments: forbid reaches(app, payment_provider);
}

main locus App {
    params { … }
    claims {
        adopt Dev;
        local_rule: require publishes(some billing, topic Settled);
    }
}
```

Every clause is still evaluated **in the adopting main's own closed world**. One text, N evaluations, N worlds — nothing about the soundness argument changes.

### Union only, no override

This is the rule the rest rests on, and it's the one worth arguing about.

If a derived constitution could override an inherited clause, weakening would be expressible and would read exactly like ordinary composition at the adoption site — a "refinement" of `count publishers <= 1` into `<= 3`. Detecting that means proving one sentence implies another, which fails open when it gets it wrong.

With union only, a stricter bound is simply **a second named claim** coexisting with the inherited one. Both are checked; the stricter one does the work. Weakening isn't rejected — it's *unexpressible*, which is the version that needs no maintenance.

So: two constitutions declaring one name is an error naming **both** origins, and a local clause shadowing an adopted one is too.

### Verified semantics

| case | behavior |
|---|---|
| `adopt Dev` where `Dev extends Core` | pulls Core's clauses transitively, all evaluated here |
| an adopted clause that fails | gates the build (adoption isn't decorative) |
| `B extends A`, both declare `rule` | error naming `A` and `B` |
| local clause named like an adopted one | error: "cannot replace an adopted one" |
| diamond (`Both extends L, R`, both extend `Base`) | `Base` contributes **once** |
| `A extends B`, `B extends A` | cycle error naming the path |
| `adopt Coer` near `Core` | "Did you mean `Core`?" |
| two constitutions sharing a name | error |
| `adopt` in a library-tier `claims { }` | error — adoption is the closing world's act |
| `adopt` inside a `constitution` | parse error pointing at `extends` |

Diamond dedup is by **origin**, not by claim name. Deduping by name would swallow the genuine two-origin collision the override rule depends on.

### Provenance instead of annotation

Once environments may add laws, two kinds of sentence share one block with opposite lifetimes: product laws true everywhere, and environment rails like "nothing reaches the real payment provider" — deliberately false in production. They read identically.

Schema **1.5** records which constitution each clause came from:

```json
{"name": "tenant_iso", "form": "…", "result": "holds", "source": "Core"}
```

A clause written in the main has no `source`. The distinction is structural and can't drift the way a hand-applied marker can. It's also what a future workspace check reads to ask whether every entrypoint adopted the shared claimset.

### Groups need no new mechanism

A constitution applied to every entrypoint can't name one application's internals — and doesn't need to. It names **shared library vocabulary**, which each entrypoint either imports or doesn't; the ones that don't get an empty group, and `may_be_empty` is the existing opt-out. That's why there are no abstract roles or binding machinery here.

### Not in this change

The manifest binding entrypoints to environments, and the workspace command that checks the (entrypoint × environment) matrix. Those sit on top of `--workspace` (#413) and need a manifest format decision; the language half is independently useful and testable, and building the tooling seam before this settled would have meant building it twice.

Ten tests in `constitutions.rs`. Also ran `claims` (16), `claims_edges` (36), `claims_phases` (27), `library_claims` (2), `artifact_integrity` (9), `topology_v2` (5), `topology_artifact_contract` (9), `xseed_claims_verbs` (5), `claims_artifact_unknowns` (4), `cross_seed_imports` (7), and the parse corpus.


rileyr:
Pushed the **tooling half** of #409 onto this branch — the manifest and the (entrypoint × environment) matrix. Both halves ship together now, so the issue closes in one merge.

### Where the constitution is bound, and why it isn't in source

`adopt Core;` in source means *always, everywhere*. That is the right meaning for a law an entrypoint always carries — but it cannot express the actual monorepo case, because an entrypoint deployed to two environments must satisfy **both** claimsets and cannot write two conflicting `adopt` lines.

So the environment binding lives in `hale.toml`, where the deployment facts already are:

```toml
[environments.dev]
constitution = "Dev"
entrypoints  = ["apps/prober", "apps/dashboard"]

[environments.prod]
constitution = "Prod"
entrypoints  = ["apps/prober"]
```

```sh
hale check apps/prober --env prod    # adopts Prod for this run
hale check --matrix                  # every (entrypoint, environment) pair
```

`--env` injects the constitution exactly as if the source had written `adopt` — same evaluation, same closed world, union with whatever the source already adopts. The source never learns where it will be deployed; it just gets checked more than once.

### The property, verified

One source, two environments, two verdicts:

```
app-a --env dev   → exit 0        (Core holds)
app-a --env prod  → exit 1        claim `strict` violated
```

and `extends` reaches through the binding — adopting `Prod` brings `Core`, with each clause attributed to **its own** origin rather than to the adopted constitution:

```json
[("tenant_iso", "holds", "Core"), ("strict", "violated", "Prod")]
```

### The hole composition can't close

An entrypoint listed in no environment is checked against no claimset at all. No single compilation can see that a sibling was left out, so `--matrix` makes it an **error**:

```
entrypoint mx/app-b is in no environment. Every entrypoint must say where it
deploys — one that is listed nowhere is checked against no claimset at all
```

A seed with no `main locus` is not an entrypoint and isn't demanded of the manifest — otherwise every library in the tree would need listing. A listed path that doesn't exist, an unknown environment name, and `--matrix` against a manifest with no `[environments]` are all errors rather than quiet successes.

### One test-writing note

The provenance test initially failed with `trailing characters` — my helper folds stderr into stdout so message assertions need not care which stream a diagnostic took, and that test's program deliberately violates a clause, so its witnesses landed in the JSON. Same trap `xseed_claims_verbs` hit when dump mode stopped suppressing diagnostics. It now captures stdout alone.

Nine tests in `env_matrix.rs`. Also re-ran `constitutions` (10), `workspace_check` (7), `check_arg_parsing` (6), `topology_artifact_contract` (9) and the docs-snippet gate.

PR #421, opened 2026-08-05
feat(fleet): claims over the composed model (GH #408 Phase 2)

rileyr:
Phase 2: the fleet claim verbs over the model Phase 1 composes. **Stacked on #420** — retarget to `main` once that merges.

## The flagship result

An interposition claim catching a route that skips its mediator, with a witness that crosses artifacts:

```
fleet claim `orders_pass_oms` violated — witness:
  prober-0::Probe::submit  [rogue/main.hl]
  -(route `bypass`)->
  gw-0::Gateway::on_order  [gw/main.hl]

fleet claim `one_order_authority` violated — witness:
  counted 2 deployed publisher endpoint(s) of `svc.order.request`: prober-0, oms-0
```

**Both components check clean on their own.** Only the deployment is wrong — which is the entire reason this tier exists, and something no per-application check could ever see.

The witness names the instance-qualified vertices, the **route** carrying the hop (a route, because there is no cross-process call to name), and the **source file** each vertex lives in. That last part is what Phase 0 was for: a bundle-global offset could not have been turned into a location by anything outside the process that produced it.

## The verbs

Carried in the plan as normalized rows, not source grammar — a generator can produce a plan without Hale syntax committing to any deployment format.

- **`forbid_reaches`** walks the composed edge set (interior calls, interior bus hops, explicit routes). `avoiding` masks a group out, making it the interposition form: any surviving path is a bypass.
- **`only_edges`** grants by wire **subject**, not transport address — there are no cross-process calls to grant.
- **`require_subscribes` / `require_publishes`** are structural: some instance in the group exposes the endpoint. Deliberately worded so nobody reads "the ledger listens for fills" as "every fill is durably booked."
- **`count_publisher_instances` / `count_subscriber_instances`** count instance-qualified **endpoints**. That is a different sort from the application tier's declaration count, so it gets a different spelling rather than silently changing what the noun means.

Groups quantify over instances by id or label; their vertices are every vertex of those instances — the same projection an application group makes from a locus to its methods, one altitude up. An unknown or empty group is an error, never a vacuously satisfied claim.

## Acceptance criteria

With #420 this covers **all fourteen** of #408's criteria for the Phase 0–2 slice. The ones this PR adds: **3** (a `forbid reaches` witness crossing A → route → B), **4** (the witness names source paths and the route), **6** (a second deployed publisher fails cardinality), **7** (removing the required subscriber fails existence), **13** (every judgment has a canary that must fail).

Six new tests, fourteen in the file total.

## Note on one test

`removing_the_required_subscriber_violates_the_existence_claim` had to drop the route as well as the subscriber — once the gateway subscribes nothing, the route to it can no longer be formed, so the composition fails before any claim runs. That is correct behaviour (an unformable route is a plan error, not a claim result), but it means the existence claim has to be tested with the route removed rather than by deleting the subscriber alone.

## Not in this slice

Transport-policy labels (Phase 4), workspace/deployment matrices (Phase 5), the meta-scheduler (Phase 6), signed plans and runtime attestation (Phase 7). External (non-Hale) nodes are also still out — every acceptance criterion here is Hale-to-Hale, and dropping `ExternalNode` removed the trust-boundary and adapter-contract questions wholesale.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,38 @@ behavior.
 
 ## Unreleased
 
+### Fleet claims over the composed model (GH #408 Phase 2)
+
+`forbid_reaches` (with `avoiding`), `only_edges`, `require_subscribes`
+/ `require_publishes`, and instance cardinality — evaluated over the
+fleet model composed in Phase 1, carried in the plan as normalized
+rows rather than source grammar.
+
+The flagship result is an interposition claim catching a route that
+skips its mediator, with a witness that crosses artifacts:
+
+```
+fleet claim `orders_pass_oms` violated — witness:
+  prober-0::Probe::submit  [rogue/main.hl]
+  -(route `bypass`)->
+  gw-0::Gateway::on_order  [gw/main.hl]
+```
+
+Both components are individually legal; only the deployment is wrong.
+The witness names the instance-qualified vertices, the route carrying
+the hop (a route, because there is no cross-process call to name), and
+the source file each vertex lives in — the last of which is exactly
+what Phase 0's source maps were built for.
+
+Fleet cardinality counts instance-qualified **endpoints**, a different
+sort from the application tier's declaration count, so it gets a
+different spelling: a second deployed publisher fails
+`count_publisher_instances` while each component still checks clean on
+its own.
+
+Groups quantify over instances by id or label, and an unknown or
+empty group is an error rather than a vacuously satisfied claim.
+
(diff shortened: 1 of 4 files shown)


rileyr:
Superseded by #422 — GitHub auto-closed this PR when its base branch (#420's) was deleted on merge, and a closed PR cannot be retargeted to `main`. Same branch, rebased onto current `main`.

PR #430, opened 2026-08-06
refactor(claims): the application checker walks the shared engine (GH #408)

rileyr:
`evaluate_forbid_reaches` had its own breadth-first walk, written before the fleet tier existed — and that second walk is where every fleet defect fixed this week came from. It had to remember each rule the first one had learned, and it forgot four: through-stdlib edges, unknown propagation, the zero-length overlap case, and the label on each hop of a witness.

Both prohibition checks now go through `model_graph::search`.

> **This description has been rewritten.** Its first version described an API that no longer exists and made two claims that were false — see *Corrections* at the end. Two rounds of external review are folded in.

## What the engine owns

`search` is generic over the vertex and edge-label types, and owns the parts that are identical everywhere and easy to get subtly wrong: the queue, the visited set, the parent tree, root seeding, masking, the step ceiling, and the propagation of incomplete vertices.

```rust
pub struct Visit<V, E, H> { pub edges: Vec<(V, E)>, pub hole: Option<H> }
pub enum HolePolicy { Halt, PathWins }
pub enum Search<V, E, H> { NotFound, Found {..}, Uncertified {..}, Saturated {..} }
```

A caller reports that a vertex's edge set is incomplete; the **engine** decides the verdict. Forgetting to consult a hole is not a mistake this API allows — the first version of this PR left that as a caller-side convention, and the review was right that it left one of the cited defect classes structurally open.

Three rules are baked in because they are semantics rather than mechanism, and each was a shipped bug before it lived here: roots are tested (source/target overlap is a zero-length path), masked vertices are neither tested nor traversed, and a reachable hole is never dropped.

Returning the parent **tree** rather than a path is what kept this small: `render_violation` walks that tree and is untouched, as is the fleet witness renderer.

## Two opposite policies, both deliberate

On an edge it cannot follow, the application checker **stops** and names the edge — the repair is to make that edge resolvable. The fleet tier **walks past** and refuses only if no path is found at all — a concrete cross-binary counterexample beats a refusal. Collapsing them during a mechanical refactor would have been the easy mistake.

## Scope, stated precisely

This centralizes unweighted **transitive** reachability, which is what `forbid reaches` asks at both tiers. Two neighbours are deliberately outside it:

- **`only edges`** enumerates direct crossing edges — a cut-edge subset query, with no transitive walk at either tier.
- **`bound`** keeps `site_count`, a recursive **weighted** traversal with its own memoization, cycle handling and step ceiling. A quantitative semiring is a different algorithm over the same edges, not a duplicate of this one.

The engine centralizes traversal and hole propagation. It is not the whole derivation: callers still decide which relations exist, which contracted stdlib edges are included, which vertices are incomplete, and what counts as a target.

## Behavior is unchanged — checked, not asserted

This is the highest-stakes evaluator in the checker, so the differential is the argument. Re-run after every revision:

| | result |
|---|---|
| `hale check` diagnostics + exit codes, 88 corpus programs | identical |
| topology artifacts (claim verdicts included), 88 corpus programs | **byte-identical** |
| topology artifacts, 3 real downstream applications | **byte-identical** |
| test suite | 1010 green |
| downstream fleet plan | composes at the same `fleet_shape_hash` |

Artifacts carry every claim's verdict, so byte-identity is a direct statement that no claim changed its answer.

## Fixed during review

- **zero-length paths were suppressed** in `ModelGraph::reaches` (`dst.contains(v) && !src.contains(v)` disabled the engine's root test) — a false absence reachable through the shared API regardless of the current caller
- **a tripped step ceiling was reported as a certified absence** — `max_steps` is now `Option<u32>`; search *exhaustion* can prove an absence, search *abandonment* cannot
- **hole propagation was a caller-side convention**, now structural
- **the enforcement guard asserted things it did not check** — `search(` was satisfied by the module's own tests; the fleet check proved only that a `ModelGraph` was mentioned; "exactly one queue" counted only `pop_front`

## Corrections to this PR's earlier description

Recorded rather than quietly edited away:

- it described `Visit::Halt`, which no longer exists
- it named the test `neither_claim_evaluator_rolls_its_own_search`, whose name was **false** — `site_count` is a traversal. Now `no_prohibition_evaluator_defines_a_private_bfs`
- it stated **"`bound` never walks the graph"**, which is simply wrong
- it reported 1001 tests; the current figure is 1010

## Deliberate follow-ups, not folded in

Each changes checker verdicts, and this PR's regression argument is byte-identical output:

1. **Application `PathWins`.** Today an unresolvable edge from one root can suppress a definitive violation reachable from another, so the result depends on traversal order. It fails closed, but the evidence is weaker. Prerequisite: the application closure must scan the whole vertex and return `Visit::partial(edges, hole)` — the policy flag alone is not enough, and a note in `claims.rs` says so where someone would act on it.
2. **Testing the target before the ceiling check.**
3. **`Verdict::Uncertified` rather than `Violated` on application saturation** — now the semantically accurate result.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,11 +30,23 @@ fleet tier walks past and refuses only if no path is found at all,
 since a concrete cross-binary counterexample is worth more than a
 refusal. `search` expresses both.
 
-`neither_claim_evaluator_rolls_its_own_search` fails the build if a
-third walk appears. Every fleet defect fixed this week — dropped
-through-stdlib edges, ignored unknowns, the zero-length overlap case,
-mislabelled witness hops — was a rule one walk had and the other
-lacked.
+Hole propagation moved INTO the engine: a caller reports that a
+vertex's edge set is incomplete and picks a policy, and the engine
+decides the verdict. Forgetting to consult a hole is no longer a
+mistake the API allows. Two fail-opens in the fleet wrapper closed
+with it — source/target overlap answered `None` instead of a
+zero-length path, and a tripped step ceiling answered `None` instead
+of refusing. Search *exhaustion* can prove an absence; search
+*abandonment* never can.
+
+Scope, stated precisely: this is BOOLEAN reachability, what `forbid
+reaches` and `only edges` ask. `bound` keeps its own weighted
+traversal (`site_count`), because a quantitative semiring is a
+different algorithm over the same edges — not a duplicate of this
+one. `no_prohibition_evaluator_defines_a_private_bfs` fails the build
+if a third frontier appears, its companions check that both
+evaluators still call the engine and that the engine holds exactly
+one queue.
 
 ### One reachability engine, shared by both tiers (GH #408)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,34 @@ behavior.
 
 ## Unreleased
 
+### The application checker moves onto the shared engine (GH #408)
+
+`forbid reaches` had its own breadth-first walk, written before the
+fleet tier existed. Both are now `model_graph::search`, which owns the
+queue, the visited set, the parent tree, root seeding, masking and the
+step ceiling — the bookkeeping that is identical everywhere and easy
+to get subtly wrong. What stays with each tier is what genuinely
+differs: what a vertex is, which edges exist, what counts as the
+target, and the diagnostic for an edge the walk cannot follow.
+
+Behavior is unchanged, and that is checked rather than asserted:
+`hale check` output and topology artifacts — claim verdicts included —
+are byte-identical across all 88 corpus programs and three real
+downstream applications, before and after.
+
+The two tiers keep their opposite policies on an unfollowable edge,
+because both are deliberate. The application checker stops at it and
+names the edge, since the repair is to make that edge resolvable. The
+fleet tier walks past and refuses only if no path is found at all,
+since a concrete cross-binary counterexample is worth more than a
+refusal. `search` expresses both.
+
+`neither_claim_evaluator_rolls_its_own_search` fails the build if a
+third walk appears. Every fleet defect fixed this week — dropped
+through-stdlib edges, ignored unknowns, the zero-length overlap case,
+mislabelled witness hops — was a rule one walk had and the other
+lacked.
+
 ### One reachability engine, shared by both tiers (GH #408)
 
 The fleet tier read the topology artifact and rebuilt a smaller graph
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,11 +39,13 @@ zero-length path, and a tripped step ceiling answered `None` instead
 of refusing. Search *exhaustion* can prove an absence; search
 *abandonment* never can.
 
-Scope, stated precisely: this is BOOLEAN reachability, what `forbid
-reaches` and `only edges` ask. `bound` keeps its own weighted
-traversal (`site_count`), because a quantitative semiring is a
-different algorithm over the same edges — not a duplicate of this
-one. `no_prohibition_evaluator_defines_a_private_bfs` fails the build
+Scope, stated precisely: this centralizes unweighted TRANSITIVE
+reachability, which is what `forbid reaches` asks at both tiers.
+`only edges` stays direct crossing-edge enumeration — a cut-edge
+subset query with no transitive walk — and `bound` keeps
+`site_count`, a weighted traversal, because a quantitative semiring
+is a different algorithm over the same edges rather than a duplicate
+of this one. `no_prohibition_evaluator_defines_a_private_bfs` fails the build
 if a third frontier appears, its companions check that both
 evaluators still call the engine and that the engine holds exactly
 one queue.
--- a/crates/hale-cli/src/fleet.rs
+++ b/crates/hale-cli/src/fleet.rs
@@ -676,10 +676,12 @@ fn evaluate_claims(
                 continue;
             }
             // An instance in BOTH groups already reaches the
-            // forbidden set by standing still. The BFS cannot see it
-            // (it refuses a destination that is also a source), so
-            // `forbid_reaches(g, g)` reported `holds` while every
-            // path in the fleet ran inside the prohibition.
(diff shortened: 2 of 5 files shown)


rileyr:
Addressed the deep-dive review in `e614644`. All three findings reproduced before fixing, and two claims this PR made about itself were wrong — corrected below.

## Findings 1 and 2 — both real, both fixed

**Zero-length paths were suppressed.** `ModelGraph::reaches` passed `dst.contains(v) && !src.contains(v)`, which disabled the root test the engine documents itself as owning. The reviewer's point that this matters *independently of the current caller* is the right one: the fleet precheck hid it, which meant the rule this PR claims to have centralized was still duplicated, and any future consumer of the "sound" wrapper would reproduce the original false absence. Predicate is now `dst.contains(v)`, with the suggested test.

**A tripped ceiling was a certified absence.** `NotFound | Saturated` were mapped together. Search *exhaustion* can prove an absence; search *abandonment* never can. `max_steps` is now `Option<u32>` — "no ceiling" is `None` rather than `u32::MAX`, which as noted could not have tripped without overflowing first.

The fleet precheck stays, but only because a plan-level message naming both groups reads better than a one-vertex witness. Correctness no longer depends on it, and the comment says so.

## Finding 3 — taken, not deferred

This was the most useful finding, so I did it rather than filing it. `Visit` is now `{ edges, hole }` with an explicit `HolePolicy`: a caller reports that a vertex's edge set is incomplete, and the **engine** decides the verdict. The fleet side-channel is gone.

You were also right that the old shape misdescribed reality — a vertex usually has known edges *and* an incomplete set, which `Halt` could not express. `Visit::partial` now does, with a test that a partial vertex still contributes its known edges under `PathWins`.

`Search` carries the parent tree on every non-empty outcome, so a refusal can render the path from the source to the unfollowable edge instead of naming the vertex alone.

## Finding 4 — the test name was false, and so was the PR body

`site_count` is a recursive weighted traversal, so "neither claim evaluator rolls its own search" was untrue as written, and **"`bound` never walks the graph" in the PR body was simply wrong**. Renamed to `no_prohibition_evaluator_defines_a_private_bfs`, with the scope stated in the module header: this is Boolean reachability; a quantitative semiring is a different algorithm over the same edges, not a duplicate.

Strengthened as suggested — catches `pop_back` and `VecDeque`, asserts both evaluators still *call* the engine, and requires **exactly one** queue in the engine rather than at least one.

## Finding 5 — added

Eight direct tests on the generic engine, covering what the wrapper tests could not reach: root-is-target-without-expansion, masked root, halt, saturation, partial vertices, multiple roots, shortest-path parent tree, plus both fixed fail-opens.

## Finding 6 — agreed, and the wording is corrected

The claim was overstated. The changelog now says what is actually true: one implementation of BFS bookkeeping plus structural hole propagation. Edge derivation and target semantics remain caller responsibilities, and `calls_via_stdlib` is still added explicitly by fleet composition.

## Deliberately not changed

**The application halt policy.** Your analysis is correct — a violation reachable from another root can be suppressed by an unrelated unresolvable edge, and the outcome depends on traversal order. But switching it changes checker verdicts, which is a semantic decision for the maintainer rather than something to fold into a refactor whose whole argument is byte-identical output. The engine now supports `PathWins` for the application tier the moment that call is made.

**Ceiling-versus-target ordering.** Same reason: it changes behavior at the boundary, and preserving the differential is worth more here than the edge case.

**`Verdict::Uncertified` for the application ceiling.** Agreed it is the accurate result; also a verdict change, so it belongs with the item above.

## Re-verified after the redesign

Topology artifacts — claim verdicts included — remain **byte-identical** across all 88 corpus programs and three real downstream applications. Downstream fleet plan composes at the same shape hash. 1010 tests green.

The earlier CI failure was unrelated: every job died at "Set up job" with `Failed to resolve action download info` before compiling anything.

rileyr:
Second-round items addressed in `2b9bfad`, and the PR description has been rewritten.

## The guard asserted things it did not check

All three correct, and the first one has a wrinkle worth recording: `read(ENGINE).contains("search(")` was vacuous — but not via the declaration, which is `pub fn search<V, E, H>(` and does not contain that substring. It was satisfied by **this module's own tests**, which call `search(` six times. Same conclusion, slightly different cause, and if anything a better illustration of the point: a guard can be kept alive by the very tests that were supposed to depend on it.

Now pinned as call sites: `let out = search(` in the engine, `match graph.reaches(` in fleet. The frontier count pins all three of exactly one `VecDeque::new()`, exactly one `pop_front()`, zero `pop_back()`.

## `only edges` is not transitive reachability

Correct, and my wording was wrong in two places. `only edges` enumerates direct crossing edges at both tiers and never walks transitively. The changelog, the guard's module header, and the PR body now name `forbid reaches` alone, with `only edges` and `bound` stated as the two deliberate neighbours.

## The module header overstated the result

Also correct. It described itself as the single derivation of "what reaches what" while callers still decide which relations exist, which contracted stdlib edges are included, which vertices are incomplete, and what counts as a target. It now says what is actually centralized — the traversal and hole propagation — and explicitly disclaims the stronger reading.

## The stale `claims.rs` comment, and a correction to my own response

Fixed, and it carried an error of mine worth naming. I said the application tier could adopt `PathWins` "the moment that call is made". You are right that it cannot: the closure returns `Visit::hole(())` and discards successors already collected for that vertex. Invisible under `Halt` because the engine stops anyway, but flipping the flag alone would make every partial vertex lose its known edges.

The note now sits in `claims.rs` where someone would act on it, and says the prerequisite is scanning the whole vertex and returning `Visit::partial(edges, hole)`.

## Not done here

The structured `ApplicationHole` conversion — you flagged it as a follow-up rather than a blocker, and I agree. It is a prerequisite for the `PathWins` switch, and both belong in the change that actually decides that policy, since that one alters verdicts and cannot be defended by a byte-identical differential.

## Re-verified

Artifacts remain byte-identical across all 88 corpus programs and three real downstream applications; downstream fleet plan composes at the same shape hash; 1010 tests green.

Thanks for both rounds — the two fail-opens in the first review were in code I had shipped as "verified byte-identical", which was true and beside the point. Behavior preservation says nothing about whether the new API is sound, and that is exactly where both bugs were.

PR #454, opened 2026-08-11
feat(chains): second tranche — take/skip/enumerate, seeded sum, sort_into/reverse_into/group_count_into

rileyr:
Stacked on #453 (retargets to `main` automatically when that merges — the only overlap is CHANGELOG/spec adjacency).

The remainder of the chain vocabulary recorded open when #380 shipped, minus stream sources. Everything keeps the zero-allocation contract: the elementwise stages fuse into the one loop, and the whole-set terminals materialize into **caller storage** — exactly the boundary `spec/semantics.md` already promised ("terminals that materialize into caller storage").

## Stages

- **`take(n)` / `skip(n)`** — count elements arriving *at their own position* (`filter(p).skip(2)` skips the first two matches). The two orders of `skip`/`take` are different chains and both are pinned. Limits evaluate once, pre-loop.
- **`enumerate()`** — binds `idx` (0-based count of elements reaching the stage) in every later stage and the terminal. Explicit opt-in, so a chain never captures a user's own `idx` local; a second `enumerate` shadows the first — the scoping falls out of the innermost-out wrapping order. A `min`/`max` key mentioning `idx` is **rejected** rather than silently miscompared (the best element's count is unrecoverable at compare time).

## Terminals

- **`sum(seed)`** — the seed is the accumulator's starting value *and* its typed zero: `sum(0.0)` sums Floats, `sum(100)` offsets an Int sum. Closes the Int-only gap without literal suffixes, pre-typecheck-safe.
- **`sort_into(target, cmp?)` / `reverse_into(target)`** — push survivors into a caller-owned `@form(vec)`, then reorder in place (the vec's own `sort()`/`sort_by(cmp)`, or an end-swap loop). Take-saturation still runs the post-loop reorder.
- **`group_count_into(target, key?)`** — one hashmap `bump(key)` (increment-or-init) per survivor; bare form keys on the element. Accumulates across chains because `bump` does.

The whole-set terminals are recognized stage-less (`xs.sort_into(sorted)`) — the compound `*_into` names belong to this vocabulary, unlike bare `into`. The it-mention recognition walker (and the subst walker) now descend if/match/tuple expressions, so a conditional key like `group_count_into(t, if it % 2 == 0 { "even" } else { "odd" })` recognizes and lowers.

## Test strategy — pairwise over the combinatorial surface

5 orderable stages × 6 terminal families × 2 sources × 3 contexts is not cartesian-testable; coverage is pairwise over the interaction classes plus the dangerous cells: every new stage × every terminal family, both orders of order-sensitive pairs, enumerate before/after filter + double-enumerate shadowing, `idx` reach into `each` blocks and group keys, early-break interactions (take's break vs `any`'s, `continue` past a counted element), hashmap `.entries` × new stages, counter re-init when a chain re-executes in a loop, uid separation, a map sandwich around positional counters, and the negatives (facade methods named `take`/`skip` stay ordinary calls — pond's zero-arg `take()` facades verified unaffected; idx-keyed min rejected; stage-less `sum(seed)` stays conservative).

13 new tests in `element_chains.rs` (28 total), plus `tests/hale/chains_tranche2_test.hl` for the typechecked end-to-end shapes. Spec vocabulary tables, module docs, and the `docs/` collections chapter updated in the same change.

Validation: workspace suite 2522 passed / 0 failed; `hale test tests/hale` 11/11; fmt gate clean; pond `ml/neural` + checks re-run green (pond sqlite/jobs test failures on this machine are a missing `sqlite3.h` system header, unrelated).

Not in this tranche (still open, by design): stream sources ("needs an end" stays a compile-time rejection), general group-into-collections (needs vec-valued hashmap cells the type system doesn't admit), `map` composed with the element-valued fallible terminals (documented projection rule unchanged).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,59 @@ behavior.
 
 ## Unreleased
 
+### Element chains: the second vocabulary tranche
+
+The remainder of the chain vocabulary recorded open when #380
+shipped, minus stream sources (still future work). Everything fuses
+into the same single loop — or, for the whole-set terminals,
+materializes into caller storage — so the zero-allocation contract is
+unchanged.
+
+New **stages**:
+
+- `take(n)` / `skip(n)` — positional selection. Both count elements
+  arriving at their own position in the chain, so
+  `filter(p).skip(2)` skips the first two *matches*; the two orders
+  of `skip`/`take` are different chains and both are pinned by
+  tests. Limits are evaluated once, before the loop.
+- `enumerate()` — binds `idx`, the 0-based count of elements
+  reaching the stage, in every later stage and the terminal. An
+  explicit opt-in stage rather than an always-bound name, so a
+  chain never captures a user's own `idx` local; a second
+  `enumerate` shadows the first for everything after it. A
+  `min`/`max` key mentioning `idx` is rejected rather than silently
+  miscompared (the best element's count is not recoverable from the
+  element at compare time).
+
+New **terminals**:
+
+- `sum(seed)` — the seed is the accumulator's starting value *and*
+  its typed zero, so `sum(0.0)` sums Float elements and `sum(100)`
+  starts an Int sum at 100. Closes the "Int elements only" gap
+  without literal suffixes, and stays pre-typecheck-safe.
+- `sort_into(target, cmp?)` / `reverse_into(target)` — push the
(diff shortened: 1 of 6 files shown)


rileyr:
Superseded by #455 — this PR was auto-closed when its stacked base branch was deleted on #453's merge; #455 is the same branch against `main`.

PR #457, opened 2026-08-11
fix(cli): structured related spans + pre-pass diags through normal reporting (downstream handoff)

rileyr:
An editor-tooling handoff reported two CLI diagnostic defects, both worst on the duplicate-top-level-name error — the easiest error to hit under the per-directory seed model (second file with its own `fn main()`), and the one that reported worst.

## 1. Debug formatting leaked into user-facing diagnostics

``duplicate top-level name `Widget` (previous declaration at Span { start: Pos(5), end: Pos(11) })`` — the raising site has no source text, so it `{:?}`-formatted the previous span into the message, which reached every editor via the LSP payload.

`Diag` now carries `related: Vec<(Span, String)>` (the reporter's suggested design, mirroring LSP `relatedInformation`), and the renderers — which do have the sources and file-base table — resolve each entry:

- text: `note: previous declaration at path:line:col`, correct across files
- `--json`: a `related` array of `{file, line, col, note}` — **absent when empty**, so existing consumers see an unchanged shape
- `hale lsp`: `DiagnosticRelatedInformation`, a clickable second location in clients (resolved from the un-shifted diagnostic, since a related span may live in a different file than the primary)

The sweep found two more `{:?}`-span leaks (duplicate topic binding, duplicate capacity slot) — same treatment — and the fn/const duplicate path, which had *no* previous location at all, now recovers it from the symbol table via `TopSymbol::span()`.

## 2. Pre-pass diagnostics bypassed the reporting path

Resolver diags raised inside `apply_sync_inference` were printed through a bare `render()` + early bail: no filename, `--json` silently ignored (**empty stdout with exit 1** — a CI gate saw a failed build with zero explaining diagnostics), wrong stream, and multi-file positions resolved against the alphabetically-first file's text (coordinates in neither file).

The reporter proved the fix in-tree: `hale lsp` already discards the pre-pass return and lets `check_bundle` re-raise through the properly-routed path — which is why the LSP attributed the same diagnostic correctly while the CLI did not. All three CLI bail sites (`check`/`run`/`build`) now match the LSP; `run` and `build` had additionally been double-reporting.

## Tests

`diag_reporting.rs`: `--json` on a duplicate-`main` seed emits NDJSON on stdout naming the *second* file with a position inside it plus the `related` array, nothing on stderr; plain mode renders `b.hl:1:6` + a cross-file note at `a.hl:1:6`; `"Span {"` never reaches a user; the ordinary-error NDJSON shape is unchanged (no `related` key when empty). `lsp.rs` gains a case asserting `relatedInformation` with the clickable previous-declaration location. Spec's `--json` row documents the `related` array.

Validation: full workspace suite 2528/2528; fmt gate clean; the report's own repro script re-run end to end against the fixed binary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,24 @@ behavior.
 
 ## Unreleased
 
+### `hale lsp`: go-to-definition on `std::` paths
+
+Downstream handoff. `textDocument/definition` on a `std::` path
+(`std::http::Server`, `std::bytes::BytesBuilder`, …) now jumps into
+the stdlib's own source. The rename table maps the user-facing path
+to the mangled declaration in the embedded `AP_SOURCE`; the
+declaration's owning per-domain file (the stdlib is a concatenation
+of `core.hl`, `http.hl`, … — now exported per-file as `AP_FILES`,
+with a test pinning the concatenation layout the span math depends
+on) is materialized into a versioned read-only cache
+(`~/.cache/hale/stdlib-<version>/`), and the location is a plain
+`file://` URI. That answers the handoff's install-story concern —
+an `install.sh` binary has no stdlib checkout — without the
+synthetic-URI schemes rust-analyzer-style setups need client
+support for: a materialized real file works in every editor.
+C-backed path-call primitives (`std::str::*` and friends) have no
+Hale definition and still return nothing.
+
 ### Diagnostics carry secondary locations, and the pre-pass stops eating them
 
 Two defects from a downstream editor-tooling handoff, both worst on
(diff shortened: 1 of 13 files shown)


rileyr:
Second commit on this branch: `textDocument/definition` on `std::` paths — the other half of the editor-tooling handoff. Rename table → embedded `AP_SOURCE` declaration span → owning per-domain file (new `AP_FILES` export, layout-pinned by test) → materialized versioned read-only cache, returned as a plain `file://` URI. Chosen over a synthetic URI scheme deliberately: a real file needs no client-side virtual-document support, so it works in every editor including bare vim, and covers the install.sh no-checkout case. C-backed path-call primitives still return nothing (no Hale source exists). Suite 2530/2530.

Issue #296, opened 2026-07-29
RFC: Deterministic record & replay — `hale replay`


rileyr:
**Ask:** name deterministic record/replay as a product feature and build it on
three pieces we are already paying for — `@deterministic` (#265), the sealed
`DeploymentPlan` (#262), and the `lotus_obs` record stream (iris). None of the
three is individually sufficient; together they are most of a flight recorder,
and today we ship none of the payoff.

The claim to earn:

> Re-run a recorded execution and get the same sequence of deliveries, the
> same state, the same outputs — in a debugger, on a different machine, weeks
> later.

## Why this issue exists now

"Reproduce the incident exactly" is worth more to someone running a system than
any dispatch number we can print, and it is the claim our substrate is closest
to supporting without being able to make. A cooperative pool is a single
consumer thread by construction, messages are copies, loci never reach
sideways, and every delivery already passes a probe — the ingredients are
sitting there uncombined.

Determinism currently appears in the spec only as an *open question*:
`spec/testing.md:358` ("Does the runtime need deterministic scheduling…
Probably yes for some benchmark classes; opt-in") and `spec/runtime.md:2155`
("single-scheduler test mode is fully deterministic by construction") — the
latter framed as a property that fell out of the cooperative scheduler, not one
we offer anyone.

Meanwhile both in-flight issues have already built toward it without either
one owning it:

- **#265** proposes `@deterministic` explicitly as "no time/entropy/env reads
  reachable — the contract deterministic/replicated workloads need, **and
  combined with a message log it buys exact replay debugging**." That message
  log is this issue.
- **#262** open question 10 asks whether an accepted deployment can be
  "serialized, inspected, **replayed**, signed, or compared against policy."
  A recording is meaningless without the arrangement it ran in; `DeploymentPlan`
  (landed as a value in #291 R3) is that arrangement.

## The three legs

| Leg | What it gives replay | Where it stands |
|---|---|---|
| **Deterministic code** | the program computes the same thing from the same inputs | `@deterministic` — #265, blocked on the `EffectSet` classification of the `lotus_*` frontier (`stdlib_surface`'s `FnEntry` column, #291 R2). `@no_ffi` / `@no_block` / `@no_recursion` landed in #294. |
| **Deterministic arrangement** | replay runs into the same graph, placement, and bindings | `DeploymentPlan` on Cx (#291 R3); #262 phase 1 makes it submittable and diffable. |
[...]

Issue #262, opened 2026-07-27
RFC: Upstream deployment elaboration and runtime meta-scheduling for `main locus`


rileyr:
## Summary

Today, `main locus` is a largely static declaration of the application’s root structure, placement, topology, bindings, and deployment policy.

This issue proposes exploring a Hale metaprogramming phase **upstream of `main locus`** in which Hale code can construct a deployment graph from runtime-visible facts—hardware topology, configuration, available resources, cluster state, or policy—and then submit that graph to the compiler/runtime for verification and activation.

The longer-term consequence may be a runtime-changeable deployment model:

> `main locus` becomes the currently committed, verified arrangement of the system rather than an arrangement fixed entirely at build time.

The component selecting those arrangements would function as a **meta-scheduler**.

An ordinary scheduler decides which work runs where.

A meta-scheduler decides:

* which loci should exist;
* how many instances should exist;
* where they should be placed;
* how their bus edges should be realized;
* which implementations should inhabit declared perspectives;
* and when the current deployment should transition to another valid deployment.

## Motivation

Hale already separates:

* logical system structure;
* physical placement;
* message topology;
* transport bindings;
* ownership and lifecycle;
* and replaceable behavior through perspectives.

However, the concrete deployment is currently authored statically.

That leaves several cases outside the language model or pushes them into external deployment tooling:

* selecting placements from the machine’s actual CPU and NUMA topology;
* adapting worker cardinality to available cores;
* selecting transports based on whether peers are local or remote;
* reacting structurally to failed hardware or lost cluster members;
* replacing one deployment strategy with another at runtime;
* specializing capacities and routing for the actual environment;
* moving a locus or graph fragment without embedding deployment policy in domain code.

The language already contains most of the vocabulary needed to describe the result. The missing piece is a programmable process that can produce and revise that description.

## Proposed model

The central idea is to treat a deployment as a typed value or compiler-recognized graph.

Conceptually:

```text
system components
+ deployment policy
+ observed environment
+ current deployment
        ↓
deployment elaborator
        ↓
candidate main graph
        ↓
verify
        ↓
seal / commit
        ↓
running system
```

The elaborator should not directly perform arbitrary deployment side effects.

Instead, it should produce a declarative candidate that Hale can inspect, validate, specialize, and activate.

Bad model:

```text
metaprogram imperatively spawns and connects the system
```

Preferred model:

```text
metaprogram describes the desired system
Hale validates and realizes it
```

A rough conceptual surface might look like:

[...]

PR #463, opened 2026-08-12
feat(replay): deterministic record & replay — `hale replay` (GH #296)

rileyr:
Phases 0–4 of #296 plus the CLI. The shipped claim, scoped per review: **re-run a recorded execution and get the same schedule and the same journaled inputs, with an explicit, checked coverage boundary.** External ingress and non-journaled I/O re-execute live and are gated behind `--allow-live-effects`; the full "same outputs" promise arrives with the staged artifact/executor layers (see the review thread).

Hardened by a full review round (2cf0e1a): recorder events on process-private rings (no iris ekind reuse), admission by executable identity (`exec_digest`), safe-by-default effect gating, atomic capture budgeting with fail-closed write paths, whole-artifact validation, bidirectional diff driven by a machine-readable runtime verdict, per-call journal argument identity, widened `pub_id`, and ABI-snapshot payload flagging.

```sh
LOTUS_OBS_RECORD=run.halerec hale run app.hl   # record
hale replay run.halerec app.hl --diff          # re-execute; report first divergence
hale replay run.halerec app.hl --at 4120       # SIGSTOP at consume #4120, attach a debugger
```

## Phase 0 — the determinism guarantee

Single-pool execution was deterministic by accident; it is now deterministic by contract. `spec/testing.md` § Determinism states it, both former "determinism mode" open questions resolve against it, and `replay_determinism.rs` pins it by comparing the complete ordered BUS record stream across repeated runs — with exact-count vacuity guards, per the negative-control discipline from #298.

## Phase 1 — a recording that never drops

`LOTUS_OBS_RECORD=<path>` flips the observation plane's disposition from *drop rather than stall* to *stall rather than drop*: an in-process drain appends every ring record to the file, a producer whose ring is full **blocks** against the drain cursor, a thread that cannot get a ring **fails the run** (exit 70, names `LOTUS_OBS_RINGS`; recording defaults to the 64-ring max first), and a drain write error fails the run (exit 74). A trailer marks clean finalize — a reader that doesn't find it holds a truncated recording and says so.

## Phase 2 — identity and payloads (format v0.3)

Every queued delivery carries `msg_id = consumer_id:16 | per-publisher-thread seq:48` (full width, loudly range-guarded) — deterministic across runs because a re-executed publisher thread re-derives its own ids in its own order, no global coordination. The delivery identity proper is **(target locus, msg_id)** — the consume record carries the target, so two subscribers of one publish on one consumer are distinguishable and the enforcement gate matches both. Consumer ids are stable where pthread ids are not: main = 1, pool workers = 16 + registration index, pinned threads = 64 + obs instance id (stamped from the synthesized `__pinned_main` prologue; anonymous ingress threads get run-unique ids from a disjoint range). Wire captures store canonical bytes; raw in-process structs store **metadata only** (an ABI snapshot would carry pointers, padding, and secret-derived bytes). Public bus records are aligned across runs by file-side identity maps (topic id → subject, public ring → consumer), and the publish probe returns its assigned sequence as a token every deliver probe receives — deliveries name their exact publish even under same-subject races and nested republish. Synchronous direct dispatch captures no payloads, but its public publish/deliver records are compared by `--diff`.

## Phase 3 — the input journal

`std::time::now`, `std::time::monotonic[_ns]`, `std::rand::next_int`, `std::os::getrandom`, and the `std::env` surface journal per consumer under recording and serve back under `LOTUS_REPLAY`. Two findings from building it: **`monotonic`/`monotonic_ns` lowered as inline `clock_gettime` IR** — no symbol, nothing a journal could interpose — so they now route through a named primitive (`lotus_time_monotonic_ns`), the same rationale that made `now` a symbol; and per-call journaling of `rand::next_int` is the only sound choice, because the global RNG's mutex makes seed-replay diverge under any thread interleaving.

A read past the recorded history falls back live and is counted — **replay degrades, never refuses** (the RFC's Q6 preference) — with a divergence summary on stderr at exit.

## Phase 4 — recorded-order enforcement

Each consumer (cooperative pool worker, pinned mailbox, main queue) re-consumes its queued deliveries in the recorded order: early arrivals are held in a per-consumer buffer and released in order, with a bounded (1s) hold after which the oldest held cell is released and the miss counted — a genuinely divergent replay reports rather than deadlocks. `where async_io` pools refuse replay loudly (their coro interleaving is a later milestone). The end-to-end pin: two pinned publishers racing 40+40 deliveries into one sink, replayed twice, both reproducing the recorded interleaving exactly — a coin flip per delivery without enforcement.

## Admission

`hale replay` compiles through the same pipeline as `hale run` and admits by identity, strongest first: a framed SHA-256 `exec_digest` over the full toolchain source tree (parser, analyses, codegen, every runtime TU, stdlib, the replay CLI — via the build-script digest), rustc version, git commit, build options, and every application source's logical path + length + contents. `shape_hash` is the secondary structural check; unstamped recordings need `--allow-unverified-model`; a truncated or internally corrupt artifact is refused by whole-artifact validation (exact trailer position + entry count) in both the CLI reader and the runtime loader, which snapshots the inherited, validated fd's bytes into anonymous memory before serving. Safety admission runs first and fails closed on `syscall`, `ffi`, `unclassified`, and transport-bound `bindings` — recursing through modules — without `--allow-live-effects`. The journal serves per-consumer with exact argument matching; env values are redacted by default (`LOTUS_OBS_RECORD_ENV=full` opts in).

## Cost contract

Unchanged where it matters: the recording and replay flags resolve in the same pre-main constructor as `lotus_obs_live` (the P19 hoist-soundness pattern) and every new site gates behind the existing branch discipline — env unset stays one predictable branch on paths that already paid one.

## Not in this PR, staged in spec + CHANGELOG

Ingress injection + fleet replay (Phase 5 — captured ingress payloads are in the recording, flagged; injection needs `LOTUS_OBS_WIRE` fleet identity), replay-under-a-different-plan (Phase 6, blocked on #262's plan admission), and async_io replay.

## Tests

2560 passed, 0 failed workspace-wide (+9 new: determinism pin, 5 recording-mode tests including losslessness under forced wrap pressure and the ring-exhaustion failure, 3 CLI end-to-end including the racing-interleaving reproduction and both admission rejections).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,30 @@ behavior.
 
 ### Record & replay, phases 2–4: `hale replay` (GH #296)
 
-Building on the phase-0/1 recording below, the rest of the RFC's
-core promise ships: **re-run a recorded execution and get the same
-deliveries in the same order, with the same inputs.**
+Building on the phase-0/1 recording below: **re-run a recorded
+execution and get the same schedule and the same journaled inputs,
+with an explicit, checked coverage boundary.** (Scoped per review:
+external ingress and non-journaled I/O re-execute live and are
+gated — see the safety flag below.)
+
+Hardened by a full review round before merge: recorder events
+moved off iris protocol ekinds onto process-private rings; replay
+admission is by executable identity (`exec_digest` = compiler
+version + source bytes), not just structural `shape_hash`, and
+unstamped recordings are refused without `--allow-unverified-model`;
+replay is **safe by default** — a program whose effect frontier
+reaches `syscall`/`ffi` is refused without `--allow-live-effects`;
+capture failure, write failure, and finalize failure all fail the
+run (never a silent gap); a recording is clean only if the entire
+artifact validates (exact trailer position + entry count); the
+journal carries per-call argument identity (a changed env name or
+rand bound is a named divergence, not a substituted value);
+`--diff` is bidirectional and fails on ANY runtime divergence via
+a machine-readable verdict; payload topic identity is a stable
+subject hash (manifest ids are registration-order and race);
+raw-struct payload captures are flagged as ABI snapshots and
+compared by size (canonical recording codecs staged); `--at
+consumer:N` is the stable multi-consumer debugger coordinate.
 
 - **`hale replay <recording> <program.hl>`** — compiles through the
   same pipeline as `hale run`, admits the recording by
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,28 @@ with an explicit, checked coverage boundary.** (Scoped per review:
 external ingress and non-journaled I/O re-execute live and are
 gated — see the safety flag below.)
 
-Hardened by a full review round before merge: recorder events
+Hardened by two full review rounds before merge. Round 2 (the
+identity/admission/trust round): the safety gate consumes TYPED
+effect rows and fails closed on `unclassified` and on
+transport-bound `bindings` (a bound publish sends real traffic
+that user-level effects cannot show); `exec_digest` is a framed
+SHA-256 over the toolchain source hash + version + options + full
+source paths/lengths/contents (32 bytes in the header, stamped in
+four parts, re-stamped at finalize); journal entries carry their
+EXACT encoded arguments (replay memcmps them — the 32-bit hashes
+they replace folded adjacent integers); env VALUES are withheld
+from recordings by default (`LOTUS_OBS_RECORD_ENV=full` opts in;
+withheld reads replay as named divergences); raw in-process
+payloads store metadata only (no pointer/padding bytes on disk);
+`msg_id` is full-width (consumer:16|seq:48, loudly guarded) and
+the delivery identity includes the target locus; the runtime
+loader independently validates the whole artifact (one open +
+fstat + private mmap, checked arithmetic, exact trailer position
+and count, per-kind value shapes); `--diff` compares per-consumer
+PUBLIC bus streams via file-side identity maps (direct dispatch
+is visible) and groups journal comparison per consumer.
+
+Round 1: recorder events
 moved off iris protocol ekinds onto process-private rings; replay
 admission is by executable identity (`exec_digest` = compiler
 version + source bytes), not just structural `shape_hash`, and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,88 @@ behavior.
 
 ## Unreleased
 
+### Record & replay, phases 2–4: `hale replay` (GH #296)
+
+Building on the phase-0/1 recording below, the rest of the RFC's
+core promise ships: **re-run a recorded execution and get the same
+deliveries in the same order, with the same inputs.**
+
+- **`hale replay <recording> <program.hl>`** — compiles through the
+  same pipeline as `hale run`, admits the recording by
+  `model_hash` (a recording from a different model is rejected,
+  never misreplayed; a truncated recording is refused), then
+  re-executes under `LOTUS_REPLAY`. `--diff` records the replay and
+  reports the first per-consumer divergence; `--at N` stops
+  (SIGSTOP) at the Nth consume so a debugger can attach.
+- **Delivery identity + payload capture** (format v0.2, tagged
+  entries): every queued delivery carries a deterministic
+  `pub_id` (stable consumer id + per-publisher-thread seq — a
+  re-executed run re-derives the same ids with no global
+  coordination); payload bytes are captured once per queued
+  publish, external wire ingress flagged. Synchronous direct
+  dispatch captures nothing by design: a closed-world same-thread
+  call cannot carry external input.
+- **The input journal**: `std::time::now`, `std::time::monotonic`
+  / `monotonic_ns` (now a named runtime primitive —
+  they lowered as inline `clock_gettime` IR nothing could
+  interpose), `std::rand::next_int`, `std::os::getrandom`, and the
+  `std::env` surface are journaled per consumer under recording
+  and served back under replay. A read past the recorded history
+  falls back live and is counted — **replay degrades, never
+  refuses** — with a divergence summary at exit.
+- **Recorded-order enforcement**: each consumer (cooperative pool
+  worker, pinned mailbox, main queue) re-consumes its queued
+  deliveries in the recorded order, holding early arrivals in a
+  per-consumer buffer with a bounded (1s) degrade path, so two
+  racing pinned publishers replay in exactly the interleaving that
+  was recorded — pinned end-to-end by a CLI test that replays a
+  40+40-delivery race twice. `where async_io` pools refuse replay
+  loudly (their coro interleaving is a later milestone).
+- Remaining, explicitly staged: ingress injection + fleet replay
+  (Phase 5, needs `LOTUS_OBS_WIRE` fleet identity), and
+  replay-under-a-different-plan (Phase 6, blocked on #262's plan
+  admission).
+
+### Record & replay, phases 0–1: the determinism guarantee, and a recording that never drops (GH #296)
+
+- **Single-pool determinism is now a stated guarantee**, not an
+  implementation accident. A program whose loci all run on the main
+  cooperative scheduler produces the same publishes and deliveries
+  in the same order on every run, given the same inputs — one pool
+  is one consumer thread by construction, so there is no scheduling
+  freedom for an order to vary within. Written into
+  `spec/testing.md` § Determinism (both former "determinism mode"
+  open questions resolved against it), pinned by
+  `replay_determinism.rs`, which compares the complete ordered
+  BUS_PUBLISH/BUS_DELIVER stream across repeated runs. A divergence
+  there is a runtime bug, never a flaky test.
+- **`LOTUS_OBS_RECORD=<path>` — lossless recording mode.** The
+  observation plane's sampler disposition (overwrite-oldest,
+  count the drops) becomes a flight recorder's: an in-process
+  drain appends every ring record to the file, a producer whose
+  ring is full **blocks** against the drain cursor instead of
+  overwriting, a thread that cannot get a ring **fails the run**
+  (recording defaults to the 64-ring maximum first), and a write
+  failure fails the run rather than truncating silently. Never
+  drop, by construction. Implies `LOTUS_OBS=1` and counts as an
+  attached observer, so a recording needs no external consumer.
+  Wholly opt-in: unset, the lowering is instruction-identical to
+  the unobserved build, per the established gate discipline.
+- **`BUS_CONSUME` (ekind 8, recording mode only): the per-consumer
+  delivery order.** BUS_DELIVER is enqueue-time and lands on the
+  publisher's ring — it cannot say in what order a consumer ran
+  its handlers. Under recording, every dequeue-driven handler
+  invoke stamps a consume record on the consuming thread
+  (subscriber locus + the thread's gapless count); its ring
+  position is the order a replay will serve back. Synchronous
+  direct dispatch deliberately emits none — its deliver record is
+  its consumption point. Never emitted outside recording, so
+  existing consumers never meet an unknown ekind.
+- The recording file format is **pre-stable** (v0.2 as shipped:
+  tagged header + ring records + payload/journal blobs +
+  clean-finalize trailer); the fully self-describing artifact
+  shape may still evolve with the remaining GH #296 phases.
+
 ### Observability: supervision in the model, backpressure on the wire, identity in the segment
 
 Five items from a downstream observability handoff, resolved as one
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -80,6 +80,7 @@ dependencies = [
  "hale-lsp",
  "hale-syntax",
  "hale-types",
+ "libc",
  "openssl",
  "serde",
  "serde_json",
--- a/crates/hale-cli/Cargo.toml
+++ b/crates/hale-cli/Cargo.toml
@@ -24,3 +24,4 @@ toml = "0.8"
 # lotus_tls.c implements std::crypto's ES256 with it), so this
 # adds no second crypto provider to the system.
 openssl = "0.10"
+libc = "0.2"
--- a/crates/hale-cli/build.rs
+++ b/crates/hale-cli/build.rs
@@ -66,6 +66,164 @@ fn embed_spec() {
     fs::write(dest, out).unwrap();
 }
 
+/// Minimal SHA-256 (no deps; build-script only). FIPS 180-4.
+fn sha256(data: &[u8]) -> [u8; 32] {
+    const K: [u32; 64] = [
+        0x428a2f98, 0x71374491, 0xb5c0fbcf, 0xe9b5dba5, 0x3956c25b,
+        0x59f111f1, 0x923f82a4, 0xab1c5ed5, 0xd807aa98, 0x12835b01,
+        0x243185be, 0x550c7dc3, 0x72be5d74, 0x80deb1fe, 0x9bdc06a7,
+        0xc19bf174, 0xe49b69c1, 0xefbe4786, 0x0fc19dc6, 0x240ca1cc,
+        0x2de92c6f, 0x4a7484aa, 0x5cb0a9dc, 0x76f988da, 0x983e5152,
+        0xa831c66d, 0xb00327c8, 0xbf597fc7, 0xc6e00bf3, 0xd5a79147,
+        0x06ca6351, 0x14292967, 0x27b70a85, 0x2e1b2138, 0x4d2c6dfc,
+        0x53380d13, 0x650a7354, 0x766a0abb, 0x81c2c92e, 0x92722c85,
+        0xa2bfe8a1, 0xa81a664b, 0xc24b8b70, 0xc76c51a3, 0xd192e819,
+        0xd6990624, 0xf40e3585, 0x106aa070, 0x19a4c116, 0x1e376c08,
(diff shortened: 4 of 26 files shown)


rileyr:
Review round addressed in 2cf0e1a. Every merge-blocking finding is either fixed or explicitly re-scoped; the finding-by-finding map:

**1 — ekind collision (blocker):** fixed structurally, your way. Recorder events (consume/enq/journal/consumer-identity) now live on **process-private per-thread rings** that never exist in the public segment — an attaching iris consumer sees only protocol records. In the file, private-ring entries carry the ring field's high bit, so the namespaces cannot be confused even in the artifact. Nothing recorder-owned touches ekinds 8/9/11/12 (or any public ekind) anymore. I took the private-substrate route rather than amending iris because it also answers your architectural point: the recorder no longer rides the observation ring at all for its own events.

**2 — execution identity (blocker):** recordings carry `exec_digest` = fnv64(compiler version, every source file name + content), stamped through the same prelude path as `model_hash`. Admission is digest-first (your `on_order * 2` example is now rejected as *different build inputs*, with both digests named); `shape_hash` demotes to the secondary structural check; unstamped recordings are refused without `--allow-unverified-model`. Honest residue, in spec: the digest does not cover the LLVM/libc toolchain outside the hale binary — a reproducible codegen digest is staged with the artifact layer.

**3 — live side effects (blocker):** replay is now refused by default when the program's inferred effect frontier reaches `syscall`/`ffi`, with the residue named and `--allow-live-effects` as the explicit, alarming override. This is deliberately coarse — the class granularity flags a lone `sleep`, and over-refusal is the safe direction. Your per-effect replay-behavior table (deterministic / record-return / virtual-handle / suppress-and-compare / unsupported) is the right long-term shape and is recorded in spec as the staged refinement.

**4 — losslessness (blocker):** capture allocation failure now **fails the run** (no silent hole, no "detected later" hand-wave); the byte budget is reserved atomically before allocation; the drain flushes every sweep; trailer/fflush/fclose failure fails the run; the file opens `0600 O_EXCL O_NOFOLLOW`. Not done, honestly staged: framed checksums, crash-prefix recovery (the flight-recorder-crashes case you called out), and the captured/written/durable high-water distinction — those belong to the dedicated artifact-writer layer.

**5 — trailer validation (blocker):** all three readers require the parse to end exactly at the trailer and the trailer count to match parsed entries. Pinned by your concrete negative test: bytes cut mid-file with the original trailer intact is refused.

**6 — false success (blocker):** the runtime writes a machine-readable verdict (journal misses, order-hold releases, unconsumed journal entries, unconsumed deliveries) and **any nonzero fails `--diff`** — the success message now derives from the verdict, not from one comparator's blind spots. The comparator itself is bidirectional over payloads (topic + flags + bytes) and journal streams; served reads re-journal under the verification recording so the streams are actually comparable.

**7 — journal identity (blocker):** one unified input stream per consumer; each entry carries kind + a 32-bit argument hash; replay refuses to serve on kind/argument mismatch *without consuming the entry*, so the first wrong read is the named divergence. Both concrete bugs fixed exactly as you described (`rand.next_int(max<=0)` consuming the next call's value; getrandom length mismatch consuming uncounted). Full argument/result encoding and callsite ids are staged with the canonical-codec work.

**8 — identity width (blocker):** `pub_id` is now consumer:12 | seq:32 (no 8-bit truncation for any current id range), and threads without stable identity get run-unique anonymous ids at ring claim — the shared-999 bucket is gone. Same-publish-to-two-subscribers-on-one-consumer remains publish-keyed; the per-queue FIFO makes their relative order deterministic, which is stated in spec rather than implied. Full delivery_id/cause_id separation is staged.

**9 — payload encoding (blocker):** raw in-process struct captures are now **flagged as ABI snapshots** and compared by size only — no ASLR false divergence, no pretense of portability; wire captures remain canonical bytes, unflagged. Payload topic identity became a stable subject hash after your point was demonstrated live: the racing-publishers test caught registration-order topic ids diverging between record and replay the moment the diff went bidirectional. Compiler-synthesized recording codecs for codec-less topics are the staged fix and the reason flagged payloads exist at all.

**Secondary:** `post_entry`'s TLS stores are guarded off the default path (residual: one TLS read per queued cell build — I'll take an A/B bench number before merge if you want it as a gate); the determinism test reads the manifest id at `e+16` (it was indeed reading `shape_hash`'s low bits and always falling back to `<unknown-N>`); `--at consumer:N` is the stable multi-consumer coordinate, with the process-wide form documented as single-consumer-only.

**Re-scoping:** spec, book, and CHANGELOG no longer claim "same outputs." The shipped claim is: *same schedule and same journaled inputs, with an explicit, checked coverage boundary* — external ingress and non-journaled I/O re-execute live behind the safety gate. Your three-layer split (semantic event substrate / dedicated artifact writer / safe executor) is recorded in the spec's staging notes as the architecture for phases 5+; the private recorder rings in this round are the first step of that separation.

Workspace suite: 2560 passed, 0 failed, including the new negative controls (mid-file corruption with intact trailer; recorder-purity of the public stream).

rileyr:
A/B bench for the dormant-path cost question (the secondary finding on `post_entry`'s TLS traffic), measured the established way: released binaries, disasm diff alongside.

**Setup.** A = `hale build` from `main` (d52b029), B = from this branch's tip (2cf0e1a), both from fresh worktree builds of the CLI with a cleared runtime-object cache. Bench program: a pinned publisher pushing 2M messages through the cross-thread queued path into a main-tree sink — the exact path that gained the cell-build TLS read, the `post_entry` branch, and the drain gates. Producer-side self-timed (`monotonic_ns` around the publish loop, which includes backpressure — so the number is end-to-end queued-dispatch throughput). All `LOTUS_*` env unset. 12 interleaved trials.

**Result — ns/msg:**

| | min | median | mean |
|---|---|---|---|
| A (main) | 154.7 | 179.4 | 177.1 |
| B (branch) | 133.0 | 149.6 | 151.7 |

B measures *faster* (min −14%, median −17%). I am explicitly **not** claiming a speedup: nothing on this branch removes work from that path, and the within-side spread (A ranges 155–190 across trials) exceeds any plausible effect of the added instructions — the sign is code-layout/alignment luck. The honest conclusion is **no measurable regression**: whatever the TLS read and gate branches cost, it is below this bench's noise floor on a path that costs ~150ns/msg to begin with.

**Disasm diff** (fresh A vs B binaries, instruction counts):

| function | A | B | delta |
|---|---|---|---|
| `lotus_mailbox_post` | 266 | 270 | +4 (TLS read → cell store) |
| `lotus_coop_pool_post` | 282 | 291 | +9 (same, both copies) |
| `lotus_bus_queue_enqueue` | 2 | 2 | 0 (tail-call wrapper) |
| `lotus_bus_local_dispatch` | 130 | 184 | +54 static |
| `lotus_bus_queue_drain` | 203 | 380 | +177 static |
| `lotus_mailbox_drain_one` | 161 | 158 | −3 (noise) |

The large static growth in the fanout/drain functions is the recorder/replay machinery inlined into branch-dead blocks — executed dormant cost is one load + predictable branch per site (the same discipline as the existing probes), and the A/B above says the resulting I-cache pressure doesn't materialize on this workload.

One incidental find while setting this up, worth its own note: the first B binaries came out 5× larger because the shared cargo `target/` dir had contaminated state from this session — a fresh-worktree rebuild resolved it (fresh-vs-fresh text: 271KB → 292KB, i.e. +21KB from the recorder machinery). The per-build runtime-object cache itself keyed correctly by content hash; the earlier "runtime cache contamination" risk from the Windows-target review remains a real class, and profile-keying that cache is cheap insurance — noting it rather than folding it into this PR.

rileyr:
Round 2 addressed in c979f28. All eight blockers plus the hardening list; two of your findings reproduced live while building the fixes, which is worth reporting first:

- The **transport-bound canary caught a second bug beyond the gate**: a prelude binding registration probes — and lazily creates the recording header — *before* the identity stamps run, so bound programs produced unstamped recordings. Identity fields are now re-stamped at finalize. Without your requested canary this would have shipped as "bound programs mysteriously fail admission."
- The **racing-publishers test caught the per-run topic-id race** the moment the diff went bidirectional, and now also journals `rand` on both pinned threads concurrently — pinning your finding 6 exactly.

**1 — safety gate:** admission consumes the typed `EffectManifestRow`s (no manifest-text splitting) and fails closed on `unclassified` and on any transport-bound `bindings` block — your exact framing: the user-level effect of a bound publish is `publish`; the send is generated deployment machinery. Canary: a unix-socket-bound publisher records fine and is refused at replay without `--allow-live-effects`, before the binary runs. Suppress-and-compare for external output (rather than refuse) is the staged next step, with the per-effect replay-class table.

**2 — exec identity:** framed SHA-256, full 32 bytes in a widened header, over: the toolchain source hash (`HALE_CODEGEN_SRC_HASH` — the stale-CLI build hash over compiler + runtime + stdlib implementation bytes, so two commits under one version differ), CLI version, build options, and every source file's **full path + length + contents**, each length-framed — your `("a","bc")` vs `("ab","c")` ambiguity is gone. Stated residue, in spec: the LLVM/libc toolchain outside the hale binary; a post-link binary digest is the staged stronger form. Unstamped recordings refuse without `--allow-unverified-model` (already in round 1, retained).

**3 — journal identity:** entries carry their exact encoded arguments, length-framed; replay memcmps and refuses on any mismatch *without consuming the entry*. The `value | 1` folding is gone with the hashes. Callsite ids remain staged.

**4 — loader validation:** `rp_load` now validates the whole artifact independently: one `open(O_NOFOLLOW)` + `fstat` + private `mmap` (the fopen-then-open pair is gone; `MAP_PRIVATE` means post-validation file mutation can't change served bytes), checked arithmetic on every frame, exact version + header size, exact end-at-trailer, trailer-count match, per-kind value shapes (i64 sizes, journal-string NUL termination), checked reallocs. The inherited-fd handoff from CLI to child is a reasonable further step; the independent validation makes the runtime safe for direct `LOTUS_REPLAY` invocation either way.

**5 — schedule visibility:** the recording now carries run-stable identity maps (topic id → subject, public ring → consumer) as meta blobs, and `--diff` compares **per-consumer public bus streams** — synchronous direct dispatch is no longer invisible to the comparator. The delivery identity is **(target locus, msg_id)**: the consume record carries the target in its id field, the parser keeps it, and the enforcement gate matches both — your two-subscribers-one-consumer `[X, X]` ambiguity is resolved. Full `cause_id` lineage remains staged.

**6 — journal comparison:** grouped per consumer; the incidental drain interleaving no longer produces false divergence. Test extended as you asked (two pinned consumers reading rand concurrently, replayed twice).

**7 — identity width:** `msg_id = consumer:16 | seq:48` full width — you were right that the private format removes any reason to squeeze into 44 bits. The pinned range is guarded against the anonymous floor (loud failure, not overlap) and the seq guard fails the run rather than wrapping. At 1M msg/s the 48-bit horizon is ~8,900 years.

**8 — secrets:** env values are **withheld from recordings by default** — names, existence, and lengths recorded; values require `LOTUS_OBS_RECORD_ENV=full`; the header carries the policy; a withheld read replays as a named divergence and fails `--diff`. Raw in-process payload captures now store **metadata only** (declared size, zero bytes) — your "record only metadata for raw local payloads" suggestion, which also deleted the pointer/padding leak wholesale. Canary asserts the plaintext value never appears in the artifact and that opting in restores exact replay. The fuller policy surface (redact/encrypt/refuse per source, compile-time secret-reach detection via the `@sealed` machinery) is staged with the artifact layer.

**Hardening:** oversized captures fail immediately (exit 74) instead of sleeping forever; the verification recording inherits the original's env policy (a full-env recording no longer spuriously diffs against a redacted verification run — found by the redaction canary); the success line states what was compared ("canonical payloads identical, raw ABI payload sizes matched"); `--at` help documents `<consumer-id>:<ordinal>`; the status file is pre-created 0600 by the CLI; the stale 8/36 comments and PR-body phase text are updated.

Format is v0.3; all three readers track it and enforce whole-artifact validation (the third-reader consolidation into one schema library is still the right call and remains staged with the artifact layer).

Workspace: 2562 passed, 0 failed, +2 canaries.

rileyr:
Round 3 addressed in 1af96e6. All four blockers, the artifact-trust preference, and the follow-up list:

**1 — toolchain identity:** you were right that wrapping the three-file 64-bit prehash in SHA-256 restored nothing. `build.rs` now computes a length-framed SHA-256 over the **full** hale-syntax/-types/-codegen/-cli source trees — parser, analyses, IR emit, every runtime TU *including `lotus_obs.c`*, stdlib seeds, and the replay CLI — plus `rustc --version` and the git commit. Your negative control (change `lotus_obs.c` alone, rebuild, replay an old artifact) now changes the digest. Application sources use logical entry-relative ids (portable across checkout roots), and the options fingerprint derives from the real `BuildOptions`. Still honestly short of a post-link binary digest, which stays staged.

**2 — module recursion:** both walkers recurse through `TopDecl::Module` — the bindings scan and both effect-row enumerations (declared and inferred, with module-qualified names). One finding from building it: the callgraph summarizer doesn't descend into inline modules yet, and a missing summary key inferred **PURE** — so recursing the enumeration alone still failed open. Module-scoped rows whose key misses now render `unclassified`, which your gate then refuses — fail closed, scoped to modules so non-module manifests are byte-identical. I also reordered the safety gate **before** identity admission: a program-inherent refusal shouldn't be masked by a recording-mismatch message. Both requested negative controls are in (module-contained live-effect fn; module-contained transport binding).

**3 — publish pairing:** exactly your sketch. `lotus_obs_bus_publish` returns its assigned sequence as a token and every `lotus_obs_bus_deliver` receives it — threaded as a local through all seven C fanouts (twelve deliver sites) and as an SSA value / entry-block slot through both codegen direct flavors, so a nested same-subject republish inside a handler cannot clobber it (the reason I avoided a TLS handoff). The old reload-high-water path survives only where no publish probe fired (dormant publish, inbound redispatch) — i.e., only where no exact pairing exists to get wrong. Canary added: two pinned publishers racing on one subject plus a handler republishing to that subject mid-fanout, replayed twice, exact both times.

**4 — determinism test:** fixed as specified — the public bus stream and the private consume stream compare independently; nothing infers a total order from drain interleaving. Agreed it was pinning a contract the artifact never made.

**Artifact trust:** the CLI now passes the validated recording as an inherited fd (`dup2` to a fixed descriptor post-fork), so the child reads the same file *object* admission validated — no path re-resolution window — and the runtime loader **reads the bytes into anonymous memory** rather than MAP_PRIVATE-mapping them, conceding your point that private-mapping visibility of later writes is unspecified. Direct `LOTUS_REPLAY` invocation still revalidates fully. Authenticated artifacts remain staged for the shareable/WAL milestone.

**Follow-ups:** withheld env entries record the value's length (a missing value and a 4KiB withheld credential are now different facts in the artifact); identified deliveries past the recorded stream's end count as `unexpected_deliveries` in the verdict, with structural cells and never-seen consumers distinguished from exhaustion; the hold-buffer realloc is checked and fails namedly; the PR body's stale v0.2 / `8|36` / model_hash passages are rewritten to the shipped design.

On the **optionality/WAL fit** section: agreed on all counts, and it maps onto the staged layering — your selection/batch/lag/durability/required dimension table is the recording-policy surface I'd expect the dedicated artifact-writer layer to expose, with `required-at-admission` becoming a claims-checkable deployment law. I've kept the PR's claim at "pre-stable deterministic replay substrate with an explicit coverage boundary" accordingly; none of the durability language appears anywhere in spec or docs.

Workspace: 2564 passed, 0 failed (+2 canaries). CI running on the push.

rileyr:
Round 4 addressed in dd4fc47. Both blockers, with one finding to report from the second canary:

**1 — eager recording init:** exactly as prescribed — `obs_on()` runs at the end of the constructor once `LOTUS_OBS_RECORD` (or `LOTUS_REPLAY`) is parsed, so the file is created (replacing whatever sat at the path), the drain spawns, and the finalizer registers before any user code — or the run fails. Finalize-time restamping covers the prelude-supplied identities, as you noted. The canary is verbatim your spec: sentinel at the path, probe-free program, artifact replaced, parses as a clean current recording with zero semantic events and a nonzero execution identity.

**2 — qualified module summary keys:** fixed as prescribed (`FnKey::method(format!("{}{}", prefix, l.name.name), …)` for both fn members and lifecycle hooks). The finding: writing your collision canary revealed the literal same-name shape **cannot typecheck** — module-scoped names share the flat top-level namespace, so `locus Worker` + `module inner { locus Worker }` is rejected as a duplicate top-level name, and `hale replay` checks before admitting. The aliasing was therefore unreachable through any checked path — but the qualified key lands regardless, as defense-in-depth for any future namespace relaxation and for unchecked callers of the manifest API, and the canary pins the reachable variant: a module locus lifecycle body with a live effect renders `unclassified` under its qualified key and is refused. If you'd rather the unreachable-collision fact be pinned too, a hale-types unit test asserting the duplicate-name diagnostic on that shape is a two-minute add — say the word.

Workspace: 2566 passed, 0 failed (+2 canaries). CI running on the push.

rileyr:
Round 5 addressed in the pushed head. Your preferred shape, exactly: the constructor now only parses the env; the generated prelude calls `lotus_obs_eager_init()` immediately after the model/exec identity setters — still before any user code or probe, so probe-free programs record, and the segment is born with its immutable header complete rather than repaired later. No-op unless recording or replay is active.

`obs_model_hash.rs` gains the requested case: a live recording process's segment must carry the same nonzero model identity as an ordinary `LOTUS_OBS=1` run (it kept a child alive, read `/dev/shm/hale-obs-<pid>`, asserted nonzero and equal). Your diagnosis of why the round-4 canary missed this was exact — finalize-time restamping silently repaired the artifact while the live header stayed wrong; the new test reads the surface the artifact test couldn't see.

Workspace: 2567 passed, 0 failed. CI running on the push.

PR #466, opened 2026-08-14
replay phase 5: durable crash prefixes, hermetic wire, ingress injection, and feed mode

rileyr:
Closes the two loudest gaps in the v0.17.0 replay story — "a crashed run loses its recording" and "a replayed server talks to the real world" — and adds the backtesting mode on top. GH #296 phase 5. Spec (`spec/runtime.md`), the book chapter, and the CHANGELOG move in the same PR. Hardened by two full adversarial review rounds (10 + 8 findings, all resolved — see the review-response comments for the point-by-point).

## 1. Durable recording / crash-prefix recovery (`--allow-truncated`, `LOTUS_OBS_RECORD_DURABLE`)

A durable flight recorder, stated precisely — **not a WAL**: the application never gates on the recording reaching stable storage. Header identity (`model_hash`, full `exec_digest`, policy flags) is committed atomically after the generated setters and stamped eagerly by the drain from the immutable committed copy, so a SIGKILLed run leaves an attributable artifact exact up to one torn frame at the tail — down to a header-only 96-byte file, the earliest crash window. Replaying a trailer-less artifact is opt-in (`--allow-truncated`): both loaders stop at the first incomplete frame and replay the prefix; under `--diff` the runtime verdict and comparator agree that post-crash events are surplus (`post_prefix_live_fallback`), while any in-prefix mismatch still fails. `LOTUS_OBS_RECORD_DURABLE=1` is the power-loss grade: fdatasync per sweep and on the finalize trailer, parent-directory sync at creation, grade recorded in the header.

## 2. Hermetic wire + ingress injection

Hermeticity is a **binding-kind capability**: native `unix://`/`udp://` transports (and the transport-locus form) are suppressed at realization — husk entry, no socket, outbound fanout sends nothing (still recorded and compared under `--diff`) — while a backend with no replay class fails closed: `shm_ring` is refused by the CLI (named) and independently at the runtime shm-open seam.

Injection runs as an explicit **boot phase**: codegen emits `lotus_replay_start_ingress()` at the main locus's boot/run boundary; the runtime snapshots the subscription registry on the main thread (workers never read the live registry) and spawns **one worker per recorded ingress source**, each carrying its source's recorded consumer identity so `--diff` aligns per-consumer streams across multiple listeners. Tape identity is the **full subject string + canonical payload shape** carried in each ingress record (FNV-32 demoted to reporting; the colliding pair is a live canary); shape mismatch refuses injection as `incompatible`. Only messages the original application **accepted** enter the tape. **Strict injection is paced** against recorded consume progress, so bounded queues shed nothing they didn't shed originally. Every tape entry ends in one named class — injected / rejected / unmatched / `late_subscription_uncovered` (subscriber created after the boot snapshot; distinguished by a teardown-time registry rescan) / incompatible / unprocessed-at-shutdown / start-failure — all machine-readable, all strict-replay divergences.

Artifact trust: **one file object** carries the recording from CLI admission into the child (O_NOFOLLOW open, `parse_file` on that object, dup2 of the same descriptor — no reopen, no substitution window), plus a runtime identity-vs-binary refusal for direct `LOTUS_REPLAY` invocations.

Consequence: covered `bindings` no longer trip the `--allow-live-effects` gate — the remaining residue (`syscall`/`ffi`/`unclassified`) is genuinely user-level.

## 3. Feed mode (`--feed`) — same recorded ingress, changed code, live nondeterministic environment

The recording is consumed as an input tape only: recorded ingress injected, wire hermetic, no journal serving, no order enforcement, no model admission (mismatch prints as information — feeding changed code is the point), no `--diff`/`--at`. The **effects gate still applies**: `--feed` bypasses identity, never safety. The exit report classifies every tape entry and derives its unclassified remainder at report time — an `exit(0)` before injection cannot fail open — and any unfed remainder fails by default (`--allow-unmatched-feed` to accept).

## Tests

Beyond the happy paths: SIGKILL-prefix admissibility with a complete four-word digest asserted in the crash artifact; header-only fixtures (96/97/103/111/112 bytes) refusing by default and admitting under the flag; a deterministic admission-vs-substitution canary (test hook holds between admission and spawn while the path is atomically replaced — the child replays the admitted artifact); truncated `--diff` accepting the post-crash surplus while an in-prefix withheld read still fails; feed effects-gate and early-exit fail-closed canaries; rejected wire excluded from the tape; the FNV-32 collision pair; capacity-one bounded pacing; binding-less feed targets; incompatible shapes; `shm_ring` refusal; two listeners under CLI `--diff`.

## Boundary (stated)

The injected tape covers `bindings` ingress into boot-registered subscribers; late subscriptions are the named `late_subscription_uncovered` coverage class. Raw user sockets are syscall-class live effects (gated); adapter loci re-execute their own logic. Live-ingest loss under load (observed while storming the canaries; independent of replay) is GH #468. Fleet-scale replay is the next milestone.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,63 +8,97 @@ behavior.
 
 ## Unreleased
 
-### Record & replay, phase 5: durability, the hermetic wire, and feed mode (GH #296)
+### Record & replay, phase 5: durable recording, the hermetic wire, and feed mode (GH #296)
 
 Three deliverables close the two loudest gaps in the v0.17.0
 replay story — "a crashed run loses its recording" and "a replayed
-server talks to the real world":
+server talks to the real world" — hardened by a full adversarial
+review round (10 findings, all resolved; the notes below fold
+them in).
 
-**WAL durability.** The drain already appended whole frames in
-stream order; now the header identity (`model_hash`,
-`exec_digest`, policy flags) is stamped eagerly — not only at
-finalize — so a SIGKILLed run leaves an artifact that is
-attributable and exact up to one torn frame at the tail. Replaying
-a trailer-less recording is an explicit opt-in
-(`hale replay --allow-truncated` / `LOTUS_REPLAY_ALLOW_TRUNCATED=1`):
-both loaders stop at the first incomplete frame, report the parsed
-extent and dropped tail bytes, and replay that prefix; `--diff`
-against a truncated baseline compares the recorded prefix (surplus
-replay events past the crash point are not divergences; missing or
-different ones still are). A finalized artifact keeps the exact
-full-parse + count checks. `LOTUS_OBS_RECORD_DURABLE=1` adds an
-`fdatasync` per flushed drain sweep for power-loss durability.
+**Durable recording / crash-prefix recovery.** Named precisely: a
+durable flight recorder, NOT a write-ahead log — the application
+never gates on the recording reaching stable storage. The header
+identity (`model_hash`, `exec_digest`, policy flags) is stamped
+eagerly — not only at finalize — so a SIGKILLed run leaves an
+artifact that is attributable and exact up to one torn frame at
+the tail. Replaying a trailer-less recording is an explicit
+opt-in (`hale replay --allow-truncated` /
+`LOTUS_REPLAY_ALLOW_TRUNCATED=1`): both loaders stop at the first
+incomplete frame, report the parsed extent, and replay that
+prefix. Under `--diff` the runtime verdict and the comparator
+AGREE on prefix semantics: recorded-history-exhausted events
+count into a separate `post_prefix_live_fallback` status key —
+the unknown post-crash suffix, executed live — while any mismatch
+inside the prefix stays a divergence (a withheld env read inside
+the prefix still fails the diff; the surplus past the tape does
+not). A finalized artifact keeps the exact full-parse + count
+checks. `LOTUS_OBS_RECORD_DURABLE=1` is the power-loss grade:
+`fdatasync` per flushed sweep AND on the finalize trailer,
+parent-directory sync at creation, grade recorded in the header.
 
-**Hermetic wire + ingress injection.** Under replay, `bindings`
-transports are suppressed at realization — the entry stays a husk
-(`transport` NULL, the shape the reclaim path already leaves), no
-socket is created, and outbound fanout to bound subjects sends
-nothing (still recorded and compared under `--diff`). In the
-listeners' stead an injector thread re-dispatches the recording's
-ingress tape in artifact order with each delivery's RECORDED
-`pub_id` pinned, so per-consumer order enforcement matches
-injected deliveries to their recorded consumes — replay of a
-server binary with no peers and no network produces byte-identical
-output with zero divergences. To make the tape injectable, reader
-threads now capture each received message's verbatim WIRE bytes
-(the struct-bytes record was metadata-only and could not be
-re-fed), with the identity pinned across the capture-then-dispatch
-pair; the artifact grows a subject-hash → name meta map (subtype
-3) so payloads route and report by name. Consequence: `bindings`
-blocks no longer trip the `--allow-live-effects` gate — the
-remaining residue (`syscall`/`ffi`/`unclassified`) is genuinely
-user-level.
+**Hermetic wire + ingress injection.** Hermeticity is a
+binding-kind capability, not a blanket assumption: native
+`unix://`/`udp://` transports (and the transport-locus form) are
+suppressed at realization — husk entry, no socket, outbound
+fanout sends nothing (still recorded and compared under
+`--diff`) — while a backend with NO replay class fails closed:
+`shm_ring` is refused by the CLI (named) and independently at the
+runtime shm-open seam. Injection runs as an explicit boot phase:
+codegen emits `lotus_replay_start_ingress()` at the main locus's
+boot/run boundary, where the runtime snapshots the subscription
+registry on the main thread (injector workers never read the live
+registry) and spawns one worker per RECORDED ingress source, each
+carrying its source's recorded consumer identity so `--diff`
+aligns per-consumer streams across multiple listeners. Tape
+identity is the full subject string plus the subject's canonical
+payload shape (FNV-32 kept for reporting only — it is
+collision-prone, and a colliding pair now routes correctly by
+name); a shape mismatch refuses injection as *incompatible*
+rather than feeding plausible wrong values. Only ACCEPTED
+messages enter the tape — wire capture runs after the
+deserializer says yes, so rejected traffic is never re-fed.
+Strict injection is PACED against recorded progress (inject one,
+wait for its recorded consumes) so bounded queues shed nothing
+they didn't shed originally. Every tape entry is classified
+(injected / rejected / unmatched / incompatible / unprocessed-at-
+shutdown / start-failure); under strict replay every non-injected
+class is a machine-readable divergence. Found along the way: the
+injector, as a cross-thread producer, must select the locked and
+GATED main-queue drain (`lotus_bus_mark_pinned`) — without it a
+replaying main drained ungated and cross-source order silently
+diverged with a clean verdict. Consequence of hermetic native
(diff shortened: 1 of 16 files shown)


rileyr:
Review round addressed in 9506db1 (all ten findings + the docs wording). Point-by-point:

1. **WAL → durable recording / crash-prefix recovery.** Took the "for this PR" shape: terminology renamed everywhere (spec, book, CHANGELOG, comments, tests), durability grade recorded in header flags (bit 1), finalize trailer now `fdatasync`'d under the durable grade, parent directory synced at creation. The spec states explicitly that the application never gates on the recording and that a load-bearing WAL grade would be a different, later mode.

2. **Truncated `--diff` coherence.** The runtime verdict now agrees with the comparator: history-exhausted events (journal past tape, deliveries past consume stream, consumers the recording never saw) count into a new `post_prefix_live_fallback` status key, excluded from the divergence sum by both the runtime totals and the CLI. In-prefix mismatches stay divergences. End-to-end canaries: `--allow-truncated --diff` succeeds past the tape boundary on a SIGKILLed run; a withheld env read *inside* the prefix still fails it.

3. **Feed keeps the effects gate.** `--feed` bypasses identity admission only; syscall/ffi/unclassified still refuse without `--allow-live-effects`. Canary: sleep-bearing program refused under `--feed` alone with the residue named, proceeds with both flags.

4. **Binding-kind hermeticity; shm_ring fails closed.** CLI scan refuses a program with any `shm_ring` binding (backend named, no override), and the runtime independently refuses at the shm-open seam — the standalone-driver builds keep the guard inert via weak externs. Canary asserts no shared-memory object is created under replay.

5. **Boot-phase injection.** Codegen emits one `lotus_replay_start_ingress()` at the main locus's boot/run boundary (params born, bindings realized/suppressed, all boot subscriptions registered, `run()` not entered). The registry snapshot happens there, on the main thread — injector workers never read the live registry, so there is nothing to race. Tape presence decides injection: the binding-less feed target canary receives its tape. (No TSan CI wiring in this PR — the design removes the shared read rather than synchronizing it; the concurrent-registration case that remains is dynamic post-boot registration, which the snapshot deliberately does not see.)

6. **Paced strict injection.** One message, then a bounded wait until its recorded consumes have re-consumed (pacing timeout degrades into the existing order/unconsumed divergences, which are the accurate report). Feed stays unpaced by design. Capacity-one canary: slowly-recorded ingress replays with zero shedding, byte-identical.

7. **Full tape identity.** Each ingress record now carries its complete subject string and the subject's canonical payload shape (`[u32 slen][subject][u32 shlen][shape][wire]`, flag bit 2); FNV-32 is demoted to reporting. Your colliding pair (`i4lt8ja7`/`w6s0e3sz`) is a live canary — asserted to collide, then asserted to route by name. Shape mismatch refuses injection as `incompatible`, never fed as a plausible value; intentional migration remains future work (an explicit adapter override, as you suggested).

8. **Accepted-only tape.** Wire capture moved after the deserializer's verdict in both readers; identity is allocated per accepted message on both sides. Canary: a runt frame followed by three valid String-payload messages leaves a 3-entry tape. (Instructive detour: an Int-payload "garbage" frame *deserializes fine* — the accepted/rejected boundary really is the codec's.)

9. **Classified accounting + feed verdict.** Six classes (injected/rejected/unmatched/incompatible/unprocessed-at-shutdown/start-failure), all in `LOTUS_REPLAY_STATUS`, all counted as strict divergences; the injector is joined at teardown and its remainder classified. Feed fails by default on any unfed remainder — `--allow-unmatched-feed` is the explicit acceptance; the "all matched" fabrication is gone.

10. **Per-source injector identity.** One worker per recorded ingress source, each carrying the recorded anonymous consumer id (fresh anon claims floored above the recorded range). Two-listener CLI `--diff` canary passes.

**Docs wording**: feed is now "same recorded ingress, changed code, live nondeterministic environment" throughout.

**Bonus find from the new canaries**: the injector, as a cross-thread producer, must call `lotus_bus_mark_pinned()` — the suppressed reader threads used to be what set it, so a replaying main was draining through the single-threaded *ungated* path: cross-source pairs consumed in enqueue order under a clean 0-divergence verdict. The colliding-subjects canary caught it; fixed and explained inline.

Not done, stated honestly: forced injector-thread-creation-failure and deterministic unprocessed-at-shutdown tests (the accounting paths exist and are exercised indirectly by the feed-fail canaries; forcing `pthread_create` failure portably is not worth the contortion); TSan CI wiring (see 5).

Full workspace: 2581 passed, 0 failed. The stacked #467 will be rebased on top next.

rileyr:
Two follow-ups on the round-response above (which referenced the wrong short hash — the review-round commit is 5b43ab1; test hardening followed in the two commits after it):

**Observation worth its own investigation** — while storming the new multi-listener canaries locally, the *live recorded runs* (not replay) occasionally lose a wire message under heavy parallel load, in two shapes: a whole socket's traffic when the peer connects between listener realization and the subscriber's bus registration (the known silent no-deserializer window), and a single tail message on a storm-descheduled reader when the peer closes before the reader drains. Both are properties of live binding ingest, untouched by this PR (only capture placement changed on that path). The canaries now handshake on a ready-file the subscriber writes at the top of `run()` — after all boot registrations — which is also an honest illustration of why the boot/run boundary is the right injection point. Filing the live-loss behavior separately rather than burying it here.

rileyr:
Round two addressed (the `fix(replay): review round 2` commit; title and body refreshed to match the code). Point by point:

**Blocker 1 — one file object.** The CLI opens the recording once (`O_NOFOLLOW`), parses admission through that object (`replay::parse_file(&File, …)`), and dup2's the *same* descriptor into the child — the later reopen is deleted, so there is no path re-resolution window. Defense in depth: at `lotus_obs_eager_init` (post-setters, pre-user-code) the runtime refuses a recording whose header model identity disagrees with the binary's own — covering direct `LOTUS_REPLAY` invocations too; feed is exempt by design and `--allow-unverified-model` plumbs through as an env override. Deterministic canary: a test hook holds between admission and spawn while the test atomically replaces the path with a distinguishable tape (different journaled argv count) — the child must print the admitted tape's value.

**Blocker 2 — committed identity.** Publication is a protocol now: the generated `lotus_obs_eager_init` — which codegen already emits after every identity setter — copies the complete identity into an immutable committed array and release-publishes a flag; the drain acquire-loads the flag and stamps that copy exactly once. No cross-thread reads of the setters' globals, no XOR key, no half-published digest possible. The kill canary now builds with a designed four-word digest and asserts every word survives into the crash artifact.

**Blocker 3 — header-only prefixes.** Both loaders' minimum is now the 96-byte header. Fixture canary at 96/97/103/111/112 bytes: default refuses as truncated, `--allow-truncated` admits each prefix.

**Blocker 4 — feed cannot fail open.** `rp_feed_report` derives the unclassified remainder itself (`ingress_len − Σclassified`) and folds it into `unprocessed` regardless of whether the join path ran — `std::process::exit(0)` in `birth()` now reports "0 of N … unprocessed" and fails; `--allow-unmatched-feed` remains the explicit acceptance. Canary included.

**Finding 5 — late subscriptions are first-class.** Kept the boot-snapshot design, made the boundary named: at injector join (teardown, main thread) the unmatched subjects are rescanned against the live registry — a subject whose subscriber exists *now* was created after the snapshot and reclassifies as `late_subscription_uncovered` (own status key, own feed-report line, counted in the strict verdict and the feed fail-set), distinct from "no subscriber exists". The spec states the snapshot boundary explicitly.

**6** — `strerror(rc)` for `pthread_create`. **7** — the env-redaction warning is replay-only; feed gets its own wording ("feed does not replay env reads; the current process environment is used"). **8** — PR title de-WAL'd; body corrected (effects gate retained, strict pacing); spec `pub_id` width fixed to `consumer:16 | seq:48`; the `--diff` paragraph is explicitly scoped "strict replay only — feed rejects it".

**Live-ingress loss** is now [#468](https://github.com/hale-lang/hale/issues/468) — the two-listener canary's bounded retry references it for deletion, and the spec/docs already scope capture to "messages the original application accepted" (nothing implies capturing what live ingest discarded).

CI is green on this head. #467 is rebased on top and carries its own round-2 response.

PR #467, opened 2026-08-14
replay phase 6: `where async_io` pools replay their recorded schedule

rileyr:
Stacked on #466 (retarget/rebase onto main after it merges — full Linux/macOS workflow set runs then; only docs CI attaches to a feature-branch base). The last refusing surface joins the replay story: `where async_io` pools record their drain's scheduling decisions and replay them. Hardened by two review rounds (5 + 6 findings, all resolved — see the review-response comments).

## The model

An async pool's nondeterminism is its drain's **scheduling**: which cell starts when (START), which fd-parked coroutine resumes when readiness arrives (RESUME), and when a timed park expires relative to both (EXPIRE). Recording stamps each decision on the pool worker's private ring (event kinds 5–7, additive to the format; the artifact header carries an async-capability bit committed with the identity). Coroutines are named by **birth ordinal** — the index of the recorded START slot that created them.

## Replay drives the drain

- **START** steps reuse the phase-4 cell gate; the slot's ordinal is assigned only to the *matching* cell. A slot whose cell never arrives times out, counts an `async-schedule` divergence, and **retires** both its ordinal and its consume slot — so one missing delivery stays one divergence instead of shifting every later coroutine into an earlier identity (and steps naming a retired slot skip immediately).
- **RESUME** steps wait for the named coroutine's readiness; early-ready coroutines are held on `ready_head` until their recorded turn.
- **EXPIRE** steps resume immediately with the timed-out sentinel — recorded sleeps fast-forward; the recording proves the deadline fired at that point in the sequence.
- **Coverage is named, never silent**: the dry-tape flag is set the moment the peek reports empty (before any flush), so post-tape work always classifies — `async_post_tape` and `unconsumed_async_steps` are status keys and divergence rows on a finalized tape, the coverage boundary on a truncated one, and a one-shot note on a pre-phase-6 artifact. `--diff` compares the per-consumer step streams bidirectionally, skipping them (with a printed note) only when the original artifact predates the capability.

## Tests

The proof surface is adversarial, not incidental: a recorded **2.5s park** whose replay must finish far below the recording's own measured wall time (only a tape-driven EXPIRE can); **readiness delivered in the opposite order** of the recorded resumes across a mixed START/RESUME/EXPIRE stream, byte-identical with zero divergences; the **cascade canary** (a live-fs-gated publisher suppressed at replay — the skipped slot counts, nothing impersonates it, the survivor keeps its identity); two replays of one recording agreeing; and the public CLI surface end to end — `hale replay --diff` exact match, a mutated schedule step failing it, and a rebuilt pre-phase-6 artifact passing it with coverage notes from both layers. The module is Linux-gated (async_io is epoll-only).

## Boundary (stated)

The schedule replays; the data of unjournaled I/O does not — a coroutine resumed at its recorded turn re-executes its recv against the live world (still gated), and bindings ingress arrives via the phase-5 injector.

**Do not merge yet** — stacked for review after #466.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,20 @@ the clock:
   never deadlock; a dry or pre-phase-6 tape hands the pool back to
   the live drain.
 
-Hardened by a review round: the artifact carries an
+Hardened by two review rounds. Round two closed the degradation
+and compatibility paths: a skipped START retires its birth ordinal
+AND its consume slot (one missing delivery no longer cascades into
+wrong resume pairings — steps naming a retired slot skip
+immediately); the dry-tape flag is set BEFORE the ready-head/held
+flushes so post-tape work is always classified; the CLI comparator
+carries the artifact's async-capability bit and skips schedule
+comparison for pre-phase-6 artifacts (matching the runtime's
+coverage note instead of contradicting it); the one-shot warning
+is atomic; the test module is Linux-gated; and the public
+`hale replay --diff` path is exercised end to end (exact match,
+mutated-step failure, and old-artifact compatibility).
+
+Round one: the artifact carries an
 async-capable header bit, the dry-tape states are named instead of
 silent (pre-phase-6 artifact → one-shot coverage note; truncated
 tape → stated coverage boundary; a finalized tape running dry
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,13 +33,25 @@ the clock:
   never deadlock; a dry or pre-phase-6 tape hands the pool back to
   the live drain.
 
+Hardened by a review round: the artifact carries an
+async-capable header bit, the dry-tape states are named instead of
+silent (pre-phase-6 artifact → one-shot coverage note; truncated
+tape → stated coverage boundary; a finalized tape running dry
+under continued async work → `async_post_tape` divergences, with
+schedule steps left unconsumed at exit as
+`unconsumed_async_steps`), and `--diff` compares the per-consumer
+async step streams bidirectionally like every other stream.
+
 Boundary, stated: the schedule replays; the data of unjournaled
 I/O still re-executes live (syscall-class, gated), and bindings
 ingress arrives via the phase-5 injector. Tests: a staggered
 two-locus async pool records and replays byte-identically with
-zero divergences, finishes faster than its recorded sleeps (proof
-the tape, not timers, drove it), and two replays of one recording
-agree with each other.
+zero divergences; a 2.5-second recorded park replays in a fraction
+of its own measured recorded wall time (the fast-forward PROOF —
+the old bound was satisfiable live); readiness arriving in the
+OPPOSITE order of the recorded resumes is held on ready_head and
+replayed in tape order across a mixed START/RESUME/EXPIRE stream;
+and two replays of one recording agree with each other.
 
 ### Record & replay, phase 5: durable recording, the hermetic wire, and feed mode (GH #296)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,39 @@ behavior.
 
 ## Unreleased
 
+### Record & replay, phase 6: `where async_io` pools replay (GH #296)
+
+The last refusing surface joins the replay story. The
+nondeterminism of an async pool is its drain's SCHEDULING — which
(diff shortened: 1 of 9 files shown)


rileyr:
Review round addressed in the `fix(replay): phase-6 review round` commit — point by point:

1. **Fast-forward now proven, not vibed.** The old three-second bound is gone from the staggered test. The dedicated proof records a 2.5s coro park, asserts the *recorded* run's own wall time (≥2.4s — the park demonstrably happened), then requires the replay under 1.5s. Only a tape-driven EXPIRE gets there; a re-timed live execution cannot.

2. **RESUME exercised adversarially.** New test: two coros parked on real fds (TCP `recv_into`) plus one on a timer, on one async pool — a mixed START/RESUME/EXPIRE stream. During replay the test delivers readiness in the *opposite* order of the recorded resumes; the early-ready coro must be pulled off epoll onto `ready_head` and held until its recorded turn. Byte-identical stdout, zero divergences, looped clean.

3. **`unconsumed_async_steps`** — summed per consumer at exit, in the status file, the human summary, and the divergence total (excluded under a truncated tape, where the remaining schedule is the stated coverage boundary rather than a divergence).

4. **`--diff` compares the schedule.** The CLI comparator now parses the private ASYNC_START/RESUME/EXPIRE events into per-consumer step streams and compares (kind, value) bidirectionally, exactly like consume and journal streams.

5. **Dry-tape states are named.** The artifact header gains an async-capable bit (flag 4, stamped by this runtime alongside the env/durable bits). Distinct outcomes: pre-phase-6 artifact → one-shot "recording predates async-schedule support" coverage note, nothing counted; truncated tape → coverage boundary (rp_report already announces it); a finalized capable tape that runs dry while re-execution continues async work → each action counts as `async_post_tape`. Nothing collapses into a silent live fallback.

On the venue notes and the article: agreed the published record/replay article is the v0.16 account and needs the substantive revision you sketched (crash-prefix recovery, hermetic wire, ingress injection, feed, async schedule, the retained not-a-WAL section) — I'll take that as post-merge site work once both PRs land, along with the sharpened framing ("replay as a checked property of the programming model").

Still not merging per instruction — stacked on #466, awaiting your verdicts on both.

rileyr:
Round two addressed in `fix(replay): phase-6 review round 2` (and the PR body is refreshed to describe the current branch):

1. **Slot-retiring degradation.** START-slot ordinals are explicit: assigned only to the matching cell, retired (bitmap) on a timeout skip — and the skip also advances the slot's consume entry, which the new canary proved necessary: without it the consume expectation stays pinned on the never-arriving delivery and every later START mismatches (your cascade, resurfacing in consume-space). RESUME/EXPIRE steps naming a retired slot skip immediately rather than burning the hold. The gate_idle "some other held cell fulfills the slot" path is gone from the replay START branch — a held cell is taken only when it *is* the slot's expected delivery. Canary: A's publisher gated on a live fs read (deleted before replay) — A's slot counts and retires, B runs correctly under its own identity, nothing impersonates A.

2. **Early dry-tape classification.** `rp_tape_dry` is set the moment the peek first reports empty, before the `ready_head`/held-cell flushes, and both flush paths classify through `lotus_async_note_live_action` — the finalized-tape flush window can no longer execute unrecorded async work with a clean verdict.

3. **Capability-aware comparison.** `Recording.async_schedule_capable` (header flag bit 2) is preserved by the CLI parser; the comparator skips schedule comparison for a non-capable original and the CLI prints the coverage note — agreeing with the runtime's one-shot "predates async-schedule support" instead of contradicting it. Canary rebuilds an old-style artifact (async events stripped, bit cleared, trailer count corrected) and asserts `--diff` passes with both layers' notes.

4. **Warning race**: `_Atomic` exchange.

5. **Linux gating**: `#![cfg(target_os = "linux")]` on the module.

6. **Public CLI contract**: three `hale replay --diff` integration tests — exact schedule match on an async program, a mutated RESUME/EXPIRE ordinal failing it, and the pre-phase-6 compatibility path passing it.

Branch is rebased on #466's current head (which carries its own round-2 fixes: one admitted file object, committed identity publication, 96-byte prefixes, fail-closed feed verdict, `late_subscription_uncovered`). Agreed on the merge sequence: #466 first, then retarget/rebase this onto main for the full workflow set before any merge. Still not merging per instruction.

Issue #436, opened 2026-08-08
Secrets: confine to a locus, classify the operation, claim the reach


rileyr:
## Where this comes from

Hale ships one security-shaped feature, `@secret`, and it **fails open silently**. `check_secret_block` (`frontier.rs:501-600`) walks `IfStmt.then_block` and never touches `else_block`; it does not handle `Stmt::Match`, `Stmt::Let`, or `Stmt::Assign`. Both of these check clean today:

```hale
fn leak(@secret token: String, flag: Bool) {
    if flag { print("nothing"); }
    else    { Out <- Msg { v: token }; }   // 0 diagnostics
}

fn leak2(@secret token: String) {
    let alias = token;
    Out <- Msg { v: alias };                // 0 diagnostics
}
```

Move the publish into the `then` branch and it errors. Unsupported forms do not produce `uncertified` — they disappear from the proof. This is the only place in the stack that violates the fail-closed posture everything else is built on.

The category error underneath is bigger than the traversal gaps: **"must not reach a sink" is a whole-world property written as a local parameter annotation and checked inside one function body.** A local walker cannot be made sound for a global property; making it exhaustive only makes it a more thorough local walker.

Four rounds of design review explored progressively larger answers — value-flow analysis with a semantic-event IR, then declaration-port summaries, then type-indexed operation law. This issue proposes something much smaller than any of them, and the reduction is not a compromise.

## The principle

Hale is an application kernel: the domain expresses the business, the library and runtime provide machinery, the compiler assembles and asserts. "The domain should say as little as possible" needs one distinction before it can decide anything, because `claims { }` is domain text and we want more of it rather than less.

> **Tax** is what the domain must write *about mechanism* — wrappers, per-site releases, ownership modifiers, parameter markings. Every unit of it is a failure to push work down into the library or the compiler.
>
> **Law** is what the domain must write *about itself* — "plugins never touch secrets", "one signing operation per request". It is irreducible: no library and no compiler can supply it, because it is business knowledge.

Minimize tax. Law is not tax; law is the point.

The corollary does most of the work: **the same feature is tax or machinery depending on who declares it.** `opaque type SigningKey` written by an application is tax; shipped by `std::crypto` it is machinery the domain merely uses. `only ops T` written by an application is tax; shipped by the type's owning library it is machinery. That sorts better than annotation-versus-keyword, because it asks who has to type it.

### 1. The effect-classified privileged handler — no compiler change

This already typechecks:

```hale
effect secret_use;

[...]

PR #471, opened 2026-08-14
types: stdlib values get their real nominal types — closes the fail-open Ty::Unknown class

### LINKED ISSUE #470: fail-open: stdlib-typed values are Ty::Unknown, so interface-satisfaction checks are skipped — a wrong-signature middleware compiles and memory-corrupts at runtime
Found investigating a downstream server whose responses curl rejected with `Unsupported HTTP/1 subversion in response`.

## Symptom

A middleware locus with the WRONG `after` signature:

```hale
locus Logging {
    fn before(ctx: std::http::Context) -> std::http::Context { ... }
    fn after(ctx: std::http::Context) -> std::http::Context { ... }   // contract: after(ctx, resp) -> Response
}
...
router.use(Logging { ... });
```

**typechecks cleanly**, and at runtime the Router's fat-pointer call of `after(ctx, resp)` hits an ABI-mismatched method. The HTTP response is memory garbage:

```
HTTP/1.1 100538325861808 Status\r\n
Content-Type: \x84\xaf\xefNp[\r\n
Content-Length: 1\r\n
Connection: close\r\n
GET\r\n
\r\n
/
```

— a nonsense status integer, a dangling Content-Type, the request's METHOD as a stray header line, and the request's PATH as the body. With the correct two-arg `after`, the same program serves a perfect response.

## Root cause

The structural-satisfaction verifier (`check_structural_impl`) is thorough — name, arity, param types, return type — and a pure user-code reproduction is refused exactly as it should be:

```
iface.hl:21:10: type error: locus `BadMw` method `after` arity does not match interface `Mw`: expected 2 arg(s), locus has 1
```

The hole is that the check is never REACHED for stdlib receivers. This is the known stdlib-path typing gap (PR #217's territory): a stdlib locus literal types as `Ty::Unknown` —

```hale
let router = std::http::Router {};
let x: Int = router;        // typechecks!
```

— so `router.use(BadMw {})` performs no argument checking at all: no signature lookup, no locus→interface coercion, no `check_structural_impl`. Codegen then coerces the locus to the interface fat pointer blindly, and the first `after` call reads/writes through a frame with the wrong shape.

PR #217 added a codegen backstop for stdlib method fallibility/arity, but interface coercion at stdlib call sites has no backstop — it fails open into undefined behavior.

## Severity

This is a soundness hole with a memory-corruption blast radius, on the *happy path* of the promoted Router/middleware battery: the very first thing a user writes against `router.use` with a slightly-wrong signature compiles silently and produces garbage I/O (or worse). Every stdlib surface that accepts an interface (`RouteHandler`, `Middleware`, metrics endpoints, log sinks) is exposed.

## Fix directions

1. **Real fix**: give stdlib locus literals their mangled nominal types (`__StdHttpRouter` etc.) instead of `Ty::Unknown`, so ordinary method-call checking and the existing coercion path just work. This retires the whole `Unknown` class, including the `let x: Int = router` absurdity above.
2. **Interim backstop** (if 1 is a larger lift): at codegen's locus→interface fat-pointer coercion, verify the concrete locus against the interface's method table (names/arity/param count are all available) and fail the build with a proper diagnostic — the same pattern as #217's fallibility backstop.

Either way, add a canary: the wrong-arity `after` above must be a compile error, not a corrupted HTTP response.

## Repro

Full minimal pair (wrong vs right signature) verified against main; the wrong one emits the garbage response above, the right one serves `HTTP/1.1 200 OK`.

rileyr:
Fixes #470 — the real fix, not the backstop.

## What was fail-open

Stdlib qualified paths resolved to `Ty::Unknown`, which is bidirectionally compatible — so `router.use(badMw)` performed no signature lookup and no interface-satisfaction check, and codegen coerced the locus to the interface fat pointer blindly. The downstream symptom that surfaced it: a middleware with `after(ctx) -> Context` (contract: `after(ctx, resp) -> Response`) compiled clean and the ABI-mismatched call corrupted the HTTP response in memory. `let x: Int = router;` typechecked. `std::log::TotallyFakeSink {}` — a nonexistent name — typechecked.

## The fix

The one the GH #436 sealed-only injection explicitly deferred ("its own change with its own measurement"):

- **The entire Hale-source stdlib surface registers into the checker's top scope** — signatures only; stdlib bodies are never added to the checked bundle, and registration diagnostics are swallowed, so no diagnostic can land on stdlib source.
- **Stdlib qualified literals resolve through `PATH_RENAMES` and validate fully** — fields checked against real params, methods against real signatures, interface coercions through the structural verifier (at literal-init sites and call sites). A `std::` literal matching nothing is a hard error; a renamed-but-Rust-implemented handle keeps the historical tolerance rather than trading fail-open for false errors. Builtin free-fn calls were already name-validated by `stdlib_surface` and are unchanged.
- **Diagnostics render public spellings** (`std::http::Middleware`, not `__StdHttpMiddleware`) via the existing demangle pass:

```
type error: locus `Passthru` method `after` arity does not match interface `std::http::Middleware`: expected 2 arg(s), locus has 1
type error: let `x`: expected `Int`, got `std::http::Router`
type error: unknown stdlib type `std::log::TotallyFakeSink` in struct/locus literal
```

## The measurement (as the deferral demanded)

Whole fixture corpus: clean. Full workspace: 2569/2569 after updating exactly two tests that pinned the old permissiveness — and one of those was **an invalid fixture the tolerance had been hiding**: `topology_v2` called `Router.get(...)`, a method that does not exist (now `add`). The `std_secret` sealed-only pin is inverted to assert the new contract, and the #470 canaries are permanent: wrong-arity middleware refused at the call site, bogus stdlib-handle field refused, typo'd literal refused.

Ergonomics follow-through: `std::http::Request` gains defaults for `version`/`headers`/`body`, keeping the documented hand-built-Request pattern (`Request { method, path, body }`) one line now that literal fields are checked for real.

Spec (semantics/verification/stdlib), the checker preamble, and the resolve rationale updated in the same change. The downstream program that surfaced the bug typechecks and serves correctly against this branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,44 @@ behavior.
 
 ## Unreleased
 
+### Stdlib values get their real types - the fail-open `Ty::Unknown` class is closed (GH #470)
+
+Found via a downstream server whose responses curl rejected: a
+middleware locus with the wrong `after` arity typechecked and the
+Router's interface fat-pointer call corrupted the HTTP response
+in-memory (garbage status integer, dangling header, the request's
+method and path bleeding into the output). Root cause: stdlib
+qualified paths resolved to `Ty::Unknown`, which is bidirectionally
+compatible - so `router.use(badMw)` performed no signature lookup,
+no interface-satisfaction check, and codegen coerced blindly. Even
+`let x: Int = router;` and `std::log::TotallyFakeSink {}` (a
+nonexistent name) typechecked.
+
+The fix is the one the sealed-loci injection (GH #436) deferred:
+the ENTIRE Hale-source stdlib surface - loci, types, interfaces,
+free fns - is now registered into the checker's top scope
+(signatures only; stdlib bodies are never added to the checked
+bundle, so no diagnostics can land on stdlib source). `std::...`
+type expressions and struct/locus literals resolve through
+PATH_RENAMES to their nominal symbols and validate for real:
+literal fields are checked, methods resolve against true
+signatures, interface coercions run the structural verifier, and a
+`std::` literal that matches nothing is a hard error. Diagnostics
+render the public spelling (`std::http::Middleware`, not the
+mangled name). Rust-implemented builtins keep their historical
+tolerance (their path-call names were already validated by the
+stdlib surface registry).
+
+Measured blast radius: the whole fixture corpus and workspace pass
+untouched except two tests that pinned the old permissiveness -
+one of which was an invalid fixture the tolerance had been hiding
+(a call to a Router method that does not exist). Ergonomics
+follow-through: `std::http::Request` gained defaults for
+`version`/`headers`/`body` so the documented hand-built-Request
+pattern stays one line. Canaries: the wrong-arity middleware, the
+`Int = router` absurdity, and the typo'd-literal case are all
+compile errors now.
+
 ### Record & replay, phase 5: durable recording, the hermetic wire, and feed mode (GH #296)
 
 Three deliverables close the two loudest gaps in the v0.17.0
@@ -114,7 +152,6 @@ without touching shared memory; two listeners surviving CLI
 `--diff` with per-source identity.
 
 ---
-
 ## v0.17.0 — the run comes back the same (2026-08-13)
 
 ### Record & replay, phases 2–4: `hale replay` (GH #296)
--- a/crates/hale-cli/tests/topology_v2.rs
+++ b/crates/hale-cli/tests/topology_v2.rs
@@ -146,7 +146,7 @@ locus Gate {
 }
 fn main() {
     let r = std::http::Router { };
-    r.get("/", Hello { });
+    r.add("GET", "/", Hello { });
     let req = std::http::Request { method: "GET", path: "/", body: "" };
     println(Gate { }.probe(r, req));
 }
--- a/crates/hale-stdlib/hl/file.hl
+++ b/crates/hale-stdlib/hl/file.hl
@@ -64,6 +64,40 @@ locus __StdIoFileFile {
         // "never spawned" default is safe.
         std::io::file::__close(self.fd);
     }
+
+    // GH #470: the handle operations are REAL locus members (the
+    // io_tcp.hl Stream style), so both the (now-nominal) checker
+    // and codegen's fallible-member machinery see one declaration.
+    // The free-fn spellings below delegate here, keeping the
+    // std::io::file::write_line(f, ...) path-call form working.
+
+    /// Read one '\n'-terminated line (newline stripped); "" at EOF.
+    fn read_line() -> String {
+        return std::io::file::__read_line(self.fd);
+    }
+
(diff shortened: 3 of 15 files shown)


rileyr:
Coverage audit round (pushed as the `test(types): coverage round` commit). The audit question was: what does this branch *not* test — and the answer found three real defects, not just gaps.

**The asymmetry:** the branch had strong negative canaries and zero committed positive coverage. `router.use` with the *correct* contract appeared in no fixture, no native test, nowhere — a regression rejecting valid stdlib use would have shipped silently. The keystone addition is `tests/hale/router_middleware_test.hl`: the correct before/after contract through `use`/`add`/`dispatch` with a hand-built Request, onion order asserted on the response body, 404 path included — typechecked *and run* (no sockets; `dispatch` directly).

**What writing the pins surfaced:**

1. **Demangling was CLI-only.** The public-spelling pins failed against `check_program` directly — LSP and library callers still saw `__StdHttpMiddleware`. Now applied centrally in `check_bundle_opts` — which promptly exposed a latent bug in `demangle_imports` itself: bare-alias renames substring-corrupted messages (`MyParseError` → `Mystd::str::ParseError`). Rewriting is now restricted to `__`-prefixed names.

2. **The documented File pattern was dead — pre-existing, verified on released 0.17.** The surface table's stale fd-era `open -> Int` row beat everything, so `let f = open(..) or raise; f.write_line(..)` couldn't typecheck. Renamed Hale-source free fns now check against their *registered* signatures (path-call sites and the or-disposition path), with handle-method sugar mirrored in the checker (`f.write_line(x)` → the namespace free fn, receiver as arg 0, fallibility intact — so bare fallible handle methods are now refused at *check* time, the typecheck twin of #217's codegen backstop).

3. **And then codegen rejected what check accepted**: `file.hl`'s handle ops were free fns with no member for the fallible-member machinery to find. They're real locus members now (the `io_tcp` Stream style), free-fn spellings delegating — `tests/hale/file_handle_test.hl` runs the open/write/reopen/read round-trip end to end for the first time.

**Also added** (`crates/hale-types/tests/stdlib_nominal.rs`, 9 tests): both failure directions pinned — acceptance of the correct contract at stdlib and user-fn sites; refusals with public spellings; unknown/wrong-arity stdlib methods; nominal annotations; the non-std qualified-literal tolerance preserved; the stale-`open` regression; and a renames↔declarations consistency canary so the literal path's tolerant branch (dead code today — every `__Std` rename target is `.hl`-declared) can only revive as a conscious decision.

Workspace 2578/2578, corpus clean, native suite 13/13, and the downstream server still checks and serves.

Honest residue: the surface table likely has more stale rows of the `open` shape (any Rust-era signature superseded by a `.hl` wrapper) — the registered-fn-wins rule neutralizes them wherever a rename exists, but a table audit against the registry would close it properly; and the codegen handler-classifier restriction on fallible stdlib calls as `or`-handlers (site three of the old `signature_for` trio) still over-refuses renamed fns with the nested-form guidance — safe direction, not touched.

Issue #476, opened 2026-08-15
Epic: canonical semantic model — one typed, provenance-bearing ApplicationModel/FleetModel behind claims, artifacts, fleet, and lowering


rileyr:
## Why

Hale already evaluates claims over finite derived relations and fails closed on unresolved edges. That is enough for the finite-model verification thesis: the structural proof halts because its subject is finite, not because the program halts. This epic is an **engineering-integrity project**, not a conference blocker:

- calls, bus topology, phases, effects, ownership, placement, artifact rows, and fleet edges live in several partial authorities (`AllocSummary`, `BusGraph`, `model::Model`, effects/frontier, codegen `DeploymentPlan`, topology serialization, fleet reconstruction);
- a downstream subsystem can omit a relation, filter, policy, or hole before the shared traversal (`model_graph`) ever sees it — the shared BFS centralizes mechanics but callers still choose the facts;
- artifacts and fleet composition *reconstruct* semantic views rather than consuming one typed value;
- diagnostics and witnesses repeatedly re-derive author spelling and provenance.

The receipts are recent and specific: the fleet claim shape defects (#426–#431) were "a downstream judgment omitted a relation the upstream knew"; the stdlib `Ty::Unknown` fail-open (#470) was a representation seam reading absence as permission; and #468's unlocked-queue race existed because `program_has_offthread` was a partial AST walk (adapter bindings counted, unix/udp forgotten) — one fact, two hand-synchronized derivations, and the comment in the fix literally reads "the compile-time union must stay the superset." Under this epic, that fact is one model row both codegen and the runtime consume, with a canary.

Build one **demand-gated, typed, provenance-bearing `ApplicationModel`**; lower every law surface to `ClaimIr`; migrate judgments to structured evidence by family; project the model into artifacts; compose admitted application models plus explicit routes into `FleetModel`. This is not a replacement for the AST or LLVM IR.

## Architectural law

```text
Source --check/derive--> Model --judge(ClaimIr)--> Evidence
                             |
                             +--project--> Artifact

Plan --admit/elaborate-------------------------------> Model
Model --derive---------------------------------------> LoweringPlan
```

No downstream AST or generic-JSON walk may rederive a fact already represented in the model. Keep distinct: source, plan, model, `ClaimIr`, evidence, artifact, lowering plan, execution evidence.

## Minimum model inventory

Typed entities and relations, not a generic property graph:

[...]

PR #477, opened 2026-08-15
feat(cli): hale topology graph — deterministic visuals from the artifact (#476 Track A)

rileyr:
The Track A deliverable from #476: render a committed `--dump-topology` artifact into SVG / Mermaid / DOT, deterministically.

```text
hale topology graph .hale.topology                          # SVG to stdout
hale topology graph .hale.topology --view claim --claim apart -o claim.svg
hale topology graph .hale.topology --format mermaid
```

**An artifact client by construction.** It reads exactly the JSON a third party reads, never Hale source, and has no dependency on the future `hale-model` crate — the input adapter later swaps from artifact rows to `ApplicationModel` without changing render configs or output. `RenderGraph` is deliberately presentational and disposable; it carries no semantic authority.

**Views**: `system` (loci with phase/effect-chipped fns, topics with subject + payload identity, publish/subscribe/call edges), `code`, `bus`, `claim` (highlights the groups the claim's normalized form names, verdict on a card), `residue` (unresolved holes as first-class nodes — and every *other* view notes residue on a card rather than omitting it: unknowns are never invisible). `--config` JSON can retitle/focus/hide/highlight — presentation only, never new semantics.

**Determinism is the product**: fixed character-cell text metrics (no font measurement → byte-stable across machines and installed fonts), stable IDs from artifact names, sorted collections, integer-only layout, no timestamps. The two tests that matter most:

- **byte-determinism** in every format, and
- **source-motion invariance**: a leading comment shifts provenance spans and the file digest (the artifacts demonstrably differ — premise-checked) but the render is *byte-identical*. That's what makes generated slide decks safe to commit and regression-test.

Plus checked-in mermaid + SVG goldens for a pinned fixture (drift becomes a reviewable diff, never silent), claim-view highlight/verdict assertions with unknown-claim refusal listing what exists, residue visibility in residue *and* non-residue views, and fail-closed handling of non-JSON input, unsupported schema families (this is a 1.x adapter), unknown views/formats, and missing `--claim`. 7/7 locally; `hale fmt --check` clean; docs claims chapter gains a "Rendering the artifact" section.

Sample renders (from the pinned fixture): system view shows the two loci + free-fns box, keyed topics with payload identity, publish (green) / subscribe (violet dashed) / call (gutter-routed) edges; claim view paints `stores`/`workers` members gold with the `forbid reaches` verdict card; residue view attaches the `indirect_call` hole to `call_it` in red.

Experimental surface pre-1.0, marked as such in usage and docs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,26 @@ behavior.
 
 ## Unreleased
 
+### `hale topology graph` - deterministic visuals from the artifact (GH #476 Track A)
+
+A committed `--dump-topology` artifact can now be rendered:
+`hale topology graph <artifact> [--view system|code|bus|claim|residue]
+[--format svg|mermaid|dot] [--claim NAME] [--config file.json] [-o out]`.
+An ARTIFACT CLIENT by construction - it reads exactly the JSON a third
+party reads, never Hale source, and has no dependency on the future
+canonical-model crate (its input adapter swaps later without changing
+output). Deterministic by design: fixed character-cell text metrics
+(no font measurement - byte-stable across machines), stable IDs from
+artifact names, no timestamps; rendering twice is byte-identical and
+moving source lines does not move a pixel (spans change, the model
+shape does not - pinned by test). The claim view highlights the named
+claim's groups and states its verdict on a card; the residue view
+renders unresolved holes as first-class nodes, and every other view
+notes them on a card rather than omitting them. Experimental surface
+pre-1.0. Pinned by `topology_graph_cli.rs` (7 tests incl. checked-in
+mermaid/SVG goldens for a pinned fixture); docs claims chapter gains
+the rendering section.
+
 ### Listen-binding ingest no longer loses messages at the edges (GH #468)
 
 Three defects, one delivery contract. The issue's two observed
--- a/crates/hale-cli/src/main.rs
+++ b/crates/hale-cli/src/main.rs
@@ -32,6 +32,7 @@ mod mcp;
 mod pkg;
 mod replay;
 mod sign;
+mod topology_graph;
 
 fn main() -> ExitCode {
     let args: Vec<String> = env::args().collect();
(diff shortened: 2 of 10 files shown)


rileyr:
All findings addressed in `4a81cb9`:

- **P1 admission**: fleet-loader order — `verify_artifact_digest` first (your hand-edited-claims attack is now a test: edited under the original digest → refused as tampered), then semantics equality, then an explicit schema floor (1.4+, since `topics` landed in 1.2 and `verdict` in 1.4), then typed presence checks for every section the adapter reads (absence errors naming the drift, never an empty graph). Verdict deliberately not required — violations render. The tests re-stamp the FNV tripwire to exercise the checks *behind* the integrity gate with a valid digest.
- **P1 imported members**: membership by longest declared-locus prefix; `p::Store::get` lands in the `p::Store` box, `p::helper` renders free; focus ownership resolves from built boxes. Two-seed fixture added.
- **P1 literal/wildcard subjects**: bus nodes = union of relation-row subjects ∪ declared topics; literal and pattern subjects get visible (marked) nodes and keep their edges in all three formats. Fixture added.
- **P2 claim residue**: early return removed — the claim view now shows the `N unresolved` card beside the verdict (the regenerated claim golden diff is the fix), and free-fn group members highlight their line.
- **P2 Mermaid IDs**: injective — kind prefix + bytewise hex escaping (`A::B` → `n_A_3a_3aB`, `A__B` → `n_A_5f_5fB`); collision canary added.
- The focus test now actually tests dropping (focus Store → App and Readings drop, neighbor Cmds stays).

12 tests (was 7); goldens regenerated.

rileyr:
Round 2 addressed in `9d9f42e`:

- **P1 row validation**: admission now checks every row this adapter reads, not just section shapes — your exact counterexample (`{"from": 7, "to": ...}` under a valid digest) is refused naming the location (`relations.calls[i].from must be a string`), never dropped into a false edge absence. Covered: all four relation row shapes, topics, claims, unknowns (residue is as undroppable as edges — a malformed unknowns row is also a refusal, tested), `sorts` elements, groups member arrays, phases, and effects. Both tests restamp the FNV tripwire so the digest is valid and the row check is what fires.

13 tests.

rileyr:
Round 3 addressed in `c6ad1c1`: admission validates referential integrity after row shape — call/stdlib-call endpoints and `publishes.fn` must exist in `sorts.fns`; subscription loci in `sorts.loci` and constructed `{locus}::{handler}` identities in `sorts.fns`; unknowns fns, phase keys, and effect keys in `sorts.fns`; group members in loci-or-fns. Your `Ghost::run` counterexample and a phantom subscription locus are both tests (restamped digest-valid), refused naming the ghost — never a silently deleted edge. 14 tests.

rileyr:
Round 4 (shared P1) addressed in `bd2fd26`: `uninhabited_interface_call:*` rows are partitioned from genuine residue — dead sites render neutrally (grey "dead dispatch: Iface.method", "dead (uninhabited)" edge), never join the unresolved count, and carry their own neutral card stating the call relation stays exact. A dead-only application shows no unresolved card at all, and its residue view states "no unresolved residue" — pinned by a mixed-artifact test (count stays 1, both species rendered distinctly) and a dead-only test. 15 tests.

Issue #468, opened 2026-08-14
live binding ingest can lose messages under load (boot-window drop + tail loss at peer close)


rileyr:
Observed while storming the replay phase-5 canaries (PR #466) — this is a property of LIVE binding ingest, independent of replay; nothing in that PR touches the live delivery path beyond capture placement.

Two shapes, both reproducible under a heavily loaded machine (full parallel test-suite storms):

**1. Boot-window drop (whole-socket).** The listener socket is bound at transport realization, which happens BEFORE later subscriber registrations in the same birth. A peer that connects and publishes in that window has its messages dispatched into an empty registry — the reader's no-deserializer path drops them silently (by documented design for relay-shaped programs, but silent and permanent for a subscriber that registers milliseconds later). Under load the window stretches: an entire socket's early traffic can vanish.

**2. Tail loss at peer close.** With a storm-descheduled reader that has not drained its socket when the peer closes, the tail message of a burst is occasionally lost (observed repeatedly with SEQPACKET unix bindings: first message delivered, second missing, other socket unaffected). Needs a precise root cause — receive-queue behavior at close vs a reader-loop exit path.

The phase-5 canaries in PR #466 sidestep this by having the subscriber touch a ready-file at the top of `run()` (all boot registrations complete by then) before the test publisher starts, plus a bounded, loudly-logged record-session retry in the CLI two-listener test. Those workarounds mark exactly where this issue bites; when it is fixed the retry should be deleted.

Worth deciding deliberately: should a listen binding buffer (or refuse) pre-registration traffic instead of silently dropping it? The publish contract language in spec/semantics.md is the anchor.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #478, opened 2026-08-15
feat(model): hale-model schema crate — canonical semantic model Change 1 (#476)

rileyr:
Change 1 of the #476 epic: the typed schema and its laws, nothing more. No derivation (that's Change 2), no consumers, **zero behavior change** — this PR exists to pin exactly the facts the epic says are expensive to retrofit, with the design note shipping as the crate's rustdoc.

## What's pinned

- **Typed sorts with the declaration/instance split** (application claims will count declarations; fleet claims count deployed instances — the distinction starts in the schema), and typed relation tables whose rows are real structs: `Call` carries dispatch kind + loop flag, `Subscribe` carries its key predicate and bounds. **Resolved rows only** — an unresolved callee or computed subject is a typed `Hole`, never a stringly edge.
- **Keyed delivery first-class**: `TopicKey`, per-publish `KeyDomain`, per-subscription `KeyPredicate` (incl. `EqReplica`), replica indices — with the two-sided polarity documented as law: `may_deliver` adds possible edges on unknowns (conservative for `forbid`), `must_deliver` requires exact coverage and is *structural* ("dispatched ⇒ delivered, process persisting" — #468's boot/quiesce temporal semantics stay out of the static relation).
- **Bounds and loss policy recorded before claims need them**: `Capacity`, `ShedPolicy`, `PublishDisposition`, `BindingLossBehavior`. A shedding policy never removes a possible-reachability edge; it invalidates an unqualified must-arrive guarantee.
- **Holes + capabilities as dual bookkeeping with a validated no-contradiction law**: `validate()` refuses a model that claims `exact_calls` while any hole hides `CALLS`, and refuses a hole that hides nothing. The positive and negative completeness accounts cannot drift.
- **`ModelHashKind::TopologyShapeV1`** names the legacy hash algorithm now, so the Change-3 projection cannot ship an unnamed identity and `.halerec` admission compatibility is a named contract from day one.
- **Canonical order as law**: unsorted or duplicated tables are not models — determinism for hashing/projection/diff is enforced once at the boundary rather than through ordered maps on hot paths.

## Architecture canaries (tests/architecture.rs, 8 tests)

The crate depends on **nothing** — the dependency canary fails on *any* dependency line, which enforces both source independence (never `hale-syntax`) and the no-serialization-promise (no serde before the Change-3 projection) with one law stronger than a denylist. Plus: dangling provenance/IDs rejected, empty holes rejected, capability-vs-hole contradiction rejected then accepted once the claim is withdrawn, unsorted/duplicate tables rejected.

Relation to the other open work: independent of #477 (Track A renderer — deliberately has no dependency on this crate); the demand-gated builder that populates these tables from a checked bundle is Change 2.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,29 @@ behavior.
 
 ## Unreleased
 
+### `hale-model` - the canonical semantic model schema (GH #476 Change 1)
+
+The first change of the canonical-model epic: a new source-independent
+crate holding the typed schema and its laws - nothing derives it yet
+(Change 2) and nothing consumes it yet, so there is zero behavior
+change. What Change 1 pins is exactly the set of facts that would be
+expensive to retrofit: typed entity sorts (declaration/instance split
+included) and relation tables with per-row provenance; keyed delivery
+as first-class schema (topic keys, publish key domains, subscription
+predicates incl. `EqReplica`, the may/must-deliver polarity documented
+as law); bounds and loss policy (capacity, shed policies, publish
+dispositions, binding loss behavior) recorded BEFORE any must-arrive
+claim exists to consume them; typed holes that must hide at least one
+relation family; positive capabilities with a validated
+no-contradiction law (a model cannot claim `exact_calls` while a hole
+hides CALLS); `ModelHashKind::TopologyShapeV1` naming the legacy hash
+algorithm so the Change-3 projection cannot ship an unnamed identity;
+and canonical-order validation (an unsorted table is not a model).
+The crate depends on NOTHING - source independence and
+no-serialization-promise are enforced by an architecture canary that
+fails on any dependency line at all. Design note ships as the crate's
+rustdoc; 8 canary tests in `tests/architecture.rs`.
+
 ### `hale topology graph` - deterministic visuals from the artifact (GH #476 Track A)
 
 A committed `--dump-topology` artifact can now be rendered:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -114,6 +114,10 @@ dependencies = [
 ]
 
 [[package]]
+name = "hale-model"
+version = "0.17.0"
+
+[[package]]
 name = "hale-stdlib"
 version = "0.17.0"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,6 +5,7 @@ members = [
     "crates/hale-corpus",
     "crates/hale-syntax",
     "crates/hale-types",
+    "crates/hale-model",
     "crates/hale-codegen",
     "crates/hale-lsp",
     "crates/hale-cli",
--- a/crates/hale-model/Cargo.toml
+++ b/crates/hale-model/Cargo.toml
@@ -1,0 +1,24 @@
+# GH #476 — the canonical semantic model (Change 1: schema skeleton).
+#
+# ARCHITECTURE LAW: this crate is source-independent. It must never
+# depend on `hale-syntax` (or transitively on the AST): the model
+# carries source-neutral provenance records, and `hale-types` maps
+# compiler spans INTO those records while deriving the model. The
+# dependency-direction canary in tests/architecture.rs enforces this
+# at test time; review enforces it at the boundary.
+#
+# Deliberately dependency-free: schema types, invariants, and laws —
+# nothing else. No serde: the external serialization surface is
+# Change 3+ (the legacy topology projection), and promising a wire
+# format from the schema crate would freeze exactly what the epic
+# says must stay evolvable during consolidation.
+
+[package]
+name = "hale-model"
+version.workspace = true
+edition.workspace = true
+authors.workspace = true
+license.workspace = true
+description = "Hale language: canonical typed semantic model (ApplicationModel schema)"
+
+[dependencies]
--- a/crates/hale-model/src/application.rs
+++ b/crates/hale-model/src/application.rs
@@ -9,15 +9,15 @@
 
 use crate::capability::Capabilities;
 use crate::entity::{
-    Binding, Function, LocusDecl, LocusInstance, PayloadContract, Phase,
-    Seed, Subject, ThreadDomain, Topic,
+    Binding, Function, Group, LocusDecl, LocusInstance, PayloadContract,
+    Phase, Seed, Subject, ThreadDomain, Topic,
 };
 use crate::hole::Hole;
 use crate::ids::{EntityRef, ProvenanceId};
 use crate::provenance::{Provenance, ProvenanceTable};
 use crate::relation::{
-    AffinedTo, Call, DeclaredIn, MemberOf, Owns, PhaseOf, PlacedIn,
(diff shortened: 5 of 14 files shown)


rileyr:
All five findings addressed in `fe0a790`:

- **P1 canonical order**: every relation table now has a documented semantic key (the `validate()` rustdoc carries the full key table) checked strictly-increasing — including `publishes`, `supervises`, `labels`, `weights`, `holes`, and the set-like nested vectors (`KeyDomain::Exact`, `CoreSet`). Your duplicated-`publishes` counterexample is now a test verbatim, and the perturbations deliberately hit tables near the end of the schema.
- **P1 unfalsifiable capabilities**: `exact_key_filters`/`exact_cardinality`/`exact_delivery_guarantees` map to new `KEY_FILTERS`/`CARDINALITY`/`DELIVERY` families; `vouched_families()` returns named triples and a canary asserts all nine flags are mapped. Went one stronger on your suggestion: an inline `KeyDomain::Unknown`/`KeyPredicate::Unknown` now *requires* a typed hole hiding `KEY_FILTERS` at its anchor (`UnrepresentedUnknown`), so your no-hole counterexample fails at the right layer and the capability contradiction follows from the hole.
- **P1 topic bounds**: `Topic.bound: Option<TopicBound>` (capacity + `on_full: Fail`); `Bounded(0)` and zero topic bounds rejected (`InvalidBound`).
- **P1 dual authorities**: `realizes` validated total + unique + agreeing with `LocusInstance::decl`; same audit applied to the binding-subject repetition (`BindingSubjectDisagrees`).
- **P2 provenance contents**: dangling `SourceId` and inverted spans are `InvalidProvenanceRecord`.

17 canaries (was 8).

rileyr:
Round 2 addressed in `418f61b`:

- **P1 site grain**: `Call`/`Publish`/`Subscribe` rows are now sites — `site` (0-based source-order ordinal within the containing function/bus block, same stability class as the shape hash) joins each canonical key. Your motivating program (`Orders <- a or wait; Orders <- b or discard;`) is two lawful rows with their own dispositions and provenance spans, pinned by test; the same-site duplicate is still rejected. `Call` additionally carries `unbounded` as a separate fact from `in_loop`, matching the topology model's deliberate separation, and the endpoint-grained merge (loop/unbounded OR) is documented as a derived projection — so Change 3 can reproduce `TopologyShapeV1` and witness spans without discarding information.
- **P1 supervision**: `error_type` added to the row and canonical key — one row per `on_failure` handler; two handlers for one child with different error types are both representable, and the same-triple duplicate is rejected.
- **P1 routing contract** (verified against spec/semantics.md, not just the review): `KeyValue` covers Bool / Int / Time / Duration (ns) / no-payload enum (by variant name — the runtime tag is derivation) / Decimal (the spec-defined u128 comparison pair, since key equality over Decimal *is* two i64 compares by contract) / String. `TopicKey.on_unmatched: swallow|fail|fallback`, and `KeyPredicate::Fallback` is the `where key == _` catch, distinct from `Any`. The fallback contract validates both ways: `_` on a non-fallback topic is `IllegalFallback`; a fallback topic without its catch is `FallbackUncovered` (the resolve-time law mirrored in schema).

22 canaries (was 17).

rileyr:
Round 3 addressed in `86ce727`:

- **Groups are model rows**: `Group` entity (name + declared `may_be_empty` — the vacuity-relevant intent selectors need) and `GroupMember` relation over `EntityRef`s (loci and free fns), canonically keyed `(group, member)`, provenanced, reference-checked. Change 3 can now project the artifact's shape-hashed groups section from the model alone — no side channel, no second walk.
- **Unkeyed key contract**: `Publish.key_domain` is `Option<KeyDomain>` — an unkeyed publish is `None`, not a builder-invented meaning. Validated in all four directions (`KeyContract`): domain-on-unkeyed and missing-on-keyed both refused, and an unkeyed topic admits only the plain `Any` subscription — `EqLiteral`/`EqReplica`/`Fallback`/`Unknown` predicates there are refused.

24 canaries.

rileyr:
Round 4 addressed in `de8e523`:

- **Dead interface calls are neither holes nor omissions**: new `dead_interface_calls` relation `(from, site, interface, method)` — representable without hiding `CALLS`, so `exact_calls` stays claimable (pinned: a dead site coexists with `exact_calls: true`), while the row keeps the fact inside the model shape so a later conformer changes the hash — the artifact's own closed-world rule, now typed.
- **Undeclared endpoints**: `Publish`/`Subscribe` are subject-grained — `subject: SubjectId` is the endpoint identity, `declared_topic: Option<TopicId>` links a declaration when one exists (subject agreement validated: `DeclaredTopicDisagrees`). Literal and wildcard endpoints are lawful rows with **no fake Topic entering the declared sort** (pinned: `topics.len()` unchanged). Keyed-ness flows from the declared topic; undeclared endpoints are unkeyed and admit only `Any`.
- **Declaration universe**: `TypeDecl` + `InterfaceDecl` entities; `EntityRef` gains `Group`/`Type`/`Interface`; `declared_in` now covers the full rename table — pinned by a test enrolling all six declaration kinds in one seed. The suggested corpus differential (a seed exporting locus/fn/topic/type/interface/group) is noted for Change 2/3, where the builder and projection exist to differentiate against.

28 canaries.

rileyr:
Round 5 addressed in `54f015e`:

- **Complete nameable universe**: verified against `top_decl_name` in `mangle.rs` — the four remaining kinds (Perspective, Const, RingLayout, Target) enter as an opaque `Declaration { kind, name, display, provenance }` with `EntityRef::Declaration`, keyed `(name, kind)`. The law is documented on `DeclKind`: every nameable TopDecl is representable (specialized sort where structure is needed, opaque row where seed membership is the only fact; Module/Claims/Constitution are deliberately nameless there and absent here). The universe canary now enrolls **all ten** nameable kinds in one seed, with the mirror-by-test note (this crate cannot depend on the AST, so the compiler-side list is mirrored by canary, not type).
- **Endpoint payloads**: `payload: PayloadContractId` on both bus rows — the `of type T` fact on literal/wildcard endpoints no longer evaporates, endpoint compatibility is representable off the rows, declared endpoints must agree with their topic's payload (`EndpointPayloadDisagrees`), and dangling payload ids are refused.

30 canaries.

rileyr:
Round 6 addressed in `780e191`: both grains now coexist — `GroupSelector { group, ordinal, selector, provenance }` keeps the AUTHORED list ordered (ordinal is a semantic fact since the artifact hashes the list ordered), with `SelectorForm::Named { member, display }` carrying the canonical ref plus authored spelling and `SelectorForm::SeedGlob { seed, display }` kept **unexpanded** as written; `GroupMember` remains the resolved grain judgments quantify over. The canary pins the exact case from the review: `{ lib::* }` vs `{ lib::A, lib::B }` — identical resolved membership, distinct authored shapes — plus the zero-member glob keeping its authored row, duplicate ordinals refused, and dangling glob seeds refused. No hash flag day. 31 canaries.

Issue #233, opened 2026-07-22
Connection-loss supervision via transports-as-loci (F.37 follow-up)


rileyr:
Follow-up to #227 / PR #232, which made binding *realization* failure a birth failure. This tracks the other half: an **established transport dying mid-run** is currently log-and-continue on the connect side (sends keep "succeeding") and silent EOF-stop on the listen side — the same broker-accepts-what-it-can't-deliver failure mode, later in the process lifetime.

The settled direction (spec/decisions.md F.37, "Deferred" section; full discussion on #227): substrate transports become **real loci** — children of `main`, instantiated by the binding entry, converging with the adapter path (adapter-binding loci are already pinned-equivalent by construction). Loss handling then reuses existing machinery instead of growing a policy DSL on binding entries:

- `birth()` = realization (#232's synchronous split is already birth-shaped)
- `run()` = the recv loop (`std::bus::__local_dispatch` exists for inbound)
- loss = `violate link_lost;` (F.27 pattern, no new syntax)
- policy = recovery primitives on main's `on_failure`: `restart` **is** reconnect, `quarantine … for d` **is** backoff, `bubble` **is** fail; default (no handler) = bubble = non-zero exit
- down-window publishes = the general "publish to a topic whose transport child is quarantined/restarting" question, answered once by quarantine semantics
- hot path stays direct C (`lotus_transport_send` against the locus-owned handle)

**Explicitly rejected:** `on_loss:`-style kwargs on binding entries — a shadow lifecycle DSL. Entries carry only guarantee parameters (`role:`, `codec(...)`, `on_overflow:`).

Sequencing:
- [ ] 1. Transports-as-loci restructure — behavior unchanged; #232's regression tests pass verbatim
- [ ] 2. Listen-side re-arm: peer EOF loops back into `accept()` on the already-bound listener (no policy needed; unblocks rolling restarts of connect-side binaries)
- [ ] 3. Loss → `violate`, default bubble (loss becomes fatal-by-default) — must land in the same change as:
- [ ] 4. `on_failure`-on-main routing (`lotus_root_panic`'s documented seat) + `restart`-as-reconnect

Guardrail (F.37): never ship loss-is-fatal without the reconnect policy beside it.

Issue #392, opened 2026-08-04
Claims after #382: the normalized model, library-tier claims, shape_hash reconciliation, interface dispatch


rileyr:
The successor to #382, which shipped its full build order in v0.14.0 (PRs #386–#391). Four threads stayed live past the close; they share one spine, so they live in one issue.

## 1. The normalized, provenance-bearing model (the spine)

Every reviewer of the #382 stack independently converged on this as the next architectural milestone: one typed model — sorts, relations, labels, weights, **per-edge source spans** — derived once, consumed by every judgment form. What it unlocks, in dependency order:

- **Witnesses that say where to edit.** Today's countermodels are paths of names; with edge provenance they carry the callsite span, publish-site span, subscription span, and destination decl — "which source decision introduced this edge and where to change it next." That is also the prerequisite for claim failures acting as a guide (capability introduction point, boundary crossed, smallest authority-preserving edit — the `hale next` direction sketched in the claim-driven-development article).
- **Artifact completeness.** The v1 topology artifact honestly scopes what it can replay (user call/bus graph, groups, bus-end existence/cardinality, carrier labels). The model export closes the rest: weights, the phase relation, a seed sort, per-edge spans — after which every claim row becomes independently re-derivable and the "compiler-certified report row" caveat disappears.
- **§8 lowering.** `@effects(...)`, `@budget`, `@phase_effects` lower onto the claim IR as pointwise sugar — one engine, per the #265 doctrine — deliberately deferred until the claim IR had survived real use. It has now (four review passes, a release); the lowering should ride the model, not precede it.
- **`during` from a real phase relation.** The shipped implementation is a source slice (phase = the named lifecycle hook/method per source locus, documented as such). A phase relation in the model replaces name-matching with semantics.

## 2. Library-tier claims

Claims that travel with an import and re-check in every closing build — the two-tier scoping story from the issue body (a seed swears about *itself and its own boundary*; world-quantification stays main's). The ownership rule needs no user-facing seed names (the compiler knows which `Program` declared each decl); the open design question is only the enforcement surface for "a dependency may not brick downstream builds with world-claims."

## 3. iris `shape_hash` reconciliation

[...]

PR #480, opened 2026-08-16
feat(types): the ApplicationModel builder — demand-gated derivation (#476 Change 2)

rileyr:
Change 2 of the #476 epic: one derivation entry point, `hale_types::model_builder::derive_application_model(&Bundle)`, assembling the canonical model from the same trusted analyses the topology artifact consumes — `AllocSummary` (calls at **site grain**, effect sites, unresolved residue), `BusGraph` (endpoint sites, payloads, spans), `model::Model` (phases, seeds, decl spans), effects/frontier (derived classes, declared carriers) — plus direct AST reads for the facts no summary carries: topic key/bound policy, subscription filters and bounds, **authored group selectors with globs unexpanded**, the full declaration universe, `@sealed`. The through-stdlib contraction is the artifact's own walk. Uninhabited dispatches land in `dead_interface_calls`; genuine residue becomes typed holes; **capabilities are computed from the holes**, so the two completeness accounts cannot drift by construction. Ownership/placement/bindings stay empty with capabilities `false` — Change 8 completes them; the artifact exports none today, so Change-2 parity doesn't need them.

One schema amendment rode along: `Supervises.child` became `SupervisedRef::{Locus, External}` — transport supervision (#233) supervises substrate types that are not locus declarations, and the schema shouldn't pretend.

## The demand contract, proven cross-process

The builder runs **only** for the new `hale model dump <target>` surface (internal, explicitly non-stable rendering; the same ill-typed refusal as the artifact; a `validate()` failure on a derived model reports as an internal error, loudly). Plain `hale check` — *with claims present*, *with `--dump-topology`* — provably never derives it: `HALE_MODEL_TRACE=1` stays silent, pinned by tests that run the real binary. In-process instrumentation via `model_builder::builds()`.

## Tested three ways

- **Family fixtures**: keyed delivery (EqReplica + fallback catch + `AnyOfType` publish domains — typed, no spurious holes), topic bounds, bounded subscriptions with shed policies, literal/wildcard endpoints, per-handler supervision with retry bounds, authored selectors in order, dead-vs-indirect separation, clean-program exactness.
- **A model/artifact differential**: both extractions agree on fns, loci, call endpoints, through-stdlib edges, publishes, subscribes, residue anchors, authored group selectors, supervision rows, and per-fn effects — the seed of Change 3's projection differential.
- **A whole-corpus property**, two-sided: derivation never panics on any parseable program, and a derived model may violate a schema law *only where the checker also refuses the program* (negative fixtures' models mirror the checker's refusals; checks-clean-but-unlawful is a builder bug). This property earned its keep on its first run: it caught a fail-open glob-alias default and two non-total lookup paths in this very builder — fixed with a totality pre-pass (unknown endpoints get subject rows + `?` payload contracts; unresolved glob aliases synthesize their authored Seed).

Local runs: builder 6/6, model CLI 4/4, renderer suite unaffected 15/15, hale-types + hale-model full suites 919/919, `hale fmt --check` clean.

Next per the build order: Change 3 (legacy topology projection + `TopologyShapeV1` differential over the corpus).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,43 @@ behavior.
 
 ## Unreleased
 
+### The ApplicationModel builder - demand-gated derivation (GH #476 Change 2)
+
+`hale_types::model_builder::derive_application_model(&Bundle)` - one
+entry point assembling the canonical model from the same trusted
+analyses the topology artifact consumes (AllocSummary, BusGraph,
+model::Model, effects/frontier) plus direct AST reads for facts no
+summary carries (topic key/bound policy, subscription filters and
+bounds, authored group selectors, the declaration universe,
+@sealed). Sites stay site-grained; through-stdlib contraction uses
+the artifact's own walk; uninhabited dispatches land in
+`dead_interface_calls` while genuine residue becomes typed holes;
+capabilities are COMPUTED from the holes so the two completeness
+accounts cannot drift. Ownership/placement/bindings stay empty
+tables with capabilities false (Change 8 completes them - the
+artifact exports none today).
+
+DEMAND-GATED, proven cross-process: the builder runs only for
+`hale model dump <target>` (new, internal non-stable format,
+same ill-typed refusal as the artifact); plain `hale check` - even
+with claims present, even with `--dump-topology` - provably never
+derives it (HALE_MODEL_TRACE=1 stays silent; pinned by test).
+
+Tested three ways: family fixtures (keyed delivery incl. fallback
++ replica, bounds, literal/wildcard endpoints, supervision,
+groups, dead-vs-indirect separation); a model/artifact
+DIFFERENTIAL asserting both extractions agree on fns, loci, calls
+(endpoint projection), through-stdlib edges, publishes,
+subscribes, unknown anchors, authored group selectors,
+supervision, and per-fn effects; and a whole-corpus property -
+derivation never panics on any parseable program, and a derived
+model may violate a schema law ONLY where the checker also
+refuses the program (negative fixtures' models mirror the
+checker's refusals; checks-clean-but-unlawful is a builder bug).
+The corpus property earned its keep immediately: its first run
+caught a fail-open glob-alias fallback and two non-total lookup
+paths in the builder.
+
 ### `hale-model` - the canonical semantic model schema (GH #476 Change 1)
 
 The first change of the canonical-model epic: a new source-independent
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -141,6 +141,7 @@ name = "hale-types"
 version = "0.17.0"
 dependencies = [
  "hale-corpus",
+ "hale-model",
  "hale-stdlib",
  "hale-syntax",
  "serde_json",
--- a/crates/hale-cli/src/main.rs
+++ b/crates/hale-cli/src/main.rs
@@ -162,6 +162,33 @@ fn main() -> ExitCode {
         return topology_graph::run_topology(&rest);
     }
 
+    // GH #476 Change 2: `hale model dump <target>` — derive and print
+    // the canonical ApplicationModel (internal, non-stable format).
+    // The DEMAND surface: this command is what builds the model;
+    // plain `hale check` provably never does (HALE_MODEL_TRACE=1
+    // shows the derivation line here and stays silent there).
+    // Implemented as a shim into the check pipeline so bundle
+    // loading, imports, and the ill-typed refusal are identical.
+    if cmd == "model" {
+        let rest: Vec<String> = args.iter().skip(2).cloned().collect();
+        if rest.first().map(String::as_str) != Some("dump") {
+            eprintln!(
+                "usage: hale model dump <file.hl | dir>\n\
+                 \n\
+                 Derives the canonical ApplicationModel (GH #476) and \
+                 prints an internal,\nnon-stable dump. Experimental \
+                 surface pre-1.0."
+            );
+            return ExitCode::from(2);
+        }
+        // The check pipeline's dump section reads PROCESS argv (it
+        // is a top-level-command scope), so the flag cannot ride the
+        // rest-args; the shim marks the demand via env instead.
+        std::env::set_var("HALE_DUMP_MODEL", "1");
+        let shim: Vec<String> = rest[1..].to_vec();
+        return run_check_cli(&shim, false);
+    }
+
     // `bench` is discovery-driven like `test`: *_bench.hl files,
     // bench_* fns, self-calibrating harness.
     if cmd == "bench" {
@@ -222,6 +249,7 @@ fn usage() {
     eprintln!("        (`hale check --help` for all)");
     eprintln!("    hale verify <file.hl | dir>   check + FAIL on any advisory (discipline gate)");
     eprintln!("    hale topology graph <artifact> render a --dump-topology artifact (svg|mermaid|dot; experimental)");
+    eprintln!("    hale model dump <file.hl | dir> derive + print the canonical ApplicationModel (internal; experimental)");
     eprintln!("    hale run   <file.hl | dir>    compile + run as a native binary");
     eprintln!("    hale build <file.hl | dir>    parse + typecheck + emit native binary");
     eprintln!("    hale replay <rec> <file.hl>   re-run a LOTUS_OBS_RECORD recording");
@@ -2891,6 +2919,7 @@ fn run_fleet(rest: &[String]) -> ExitCode {
 /// evaluations.
 const PER_SEED_FLAGS: &[&str] = &[
     "--dump-topology",
+    "--dump-model",
     "--check-topology",
     "--check-topology-shape",
     "--dump-effects-manifest",
@@ -2910,6 +2939,10 @@ const CHECK_FLAGS: &[(&str, bool)] = &[
     ("--dump-effects-manifest", false),
     ("--dump-resource-budget", false),
     ("--dump-topology", false),
+    // GH #476 Change 2: derive + print the canonical
+    // ApplicationModel (internal format). The demand surface the
+    // `hale model dump` shim routes through.
+    ("--dump-model", false),
     ("--json", false),
     ("--no-warn-unbounded-alloc", false),
     // Retired opt-in spelling from when the unbounded-alloc survey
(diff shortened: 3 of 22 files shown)


rileyr:
All six round-6 blockers addressed in `8d50097` (note: the review examined pre-PR head `e31c888`; this PR is the same work rebased onto main as #480):

- **Raw vs display identity**: topics carry three identities kept apart — raw post-merge name (wire_subjects/graph key), display spelling (`Topic.name`), and the raw WIRE subject. An imported topic's explicit subject survives demangling; a subject-less import keeps its mangled wire identity; named selectors store the demangled author spelling. Cross-seed canary constructs the bundle the way the import pass does (explicit-subject, subject-less, and named-member `lib::Worker` cases).
- **Structural payload identity**: shape-only contracts through one shared renderer (`canonical_type_shape` extracted from the topic renderer so literal `of type T` endpoints cannot drift). Pinned: equal shapes across subjects share a contract; a field change on a literal endpoint's type changes it; a structure-preserving rename does not. The fused subject+shape hash is now explicitly a Change-3 projection.
- **Dispositions preserved**: span-indexed Send walk over every container (match arms included) joined by containment — `or discard` + `or raise` on one topic in one fn are two rows with their own dispositions, pinned.
- **Authored sites + two-component lattice**: conformer alternatives share one site via `dispatch_group` (2-vs-3-conformer canary proves the later call's ordinal doesn't move); unresolved/dead edges consume ordinals as authored sites; contraction propagates `PathFlags{in_loop, unbounded}` with revisit-on-strengthen. The `(true,false)` cell is pinned at direct grain (`while i < 3` is proven bounded); no stdlib surface re-emerges through a proven-bounded interior today, so the contraction canary pins path truth and documents that open cell for when one exists.
- **Declaration-derived function universe**: AST-enumerated (free fns, hooks by kind, modes as `FunctionKind::Mode`) ∪ summary keys — an empty free fn exists, group-resolves, and carries its decl span; the artifact differential is now subset-directed with Change 3 filtering back down.
- **`AnyOfType` names the key's type**: resolved through the payload decl — `keyed_by sensor` ⇒ `AnyOfType("Int")`, asserted by value.

14 builder tests (was 6); 927/927 across hale-types + hale-model; fmt clean.

rileyr:
Round 7 addressed in `6dff3e1`:

- **Canonical vs display**: `name` = RAW post-merge symbol everywhere; Topic and Group grew `display` fields (they had none to carry the spelling). The alias-independence canary derives one lib under `p::` and `db::` and asserts canonical equality with display-only difference. Interface names in `DispatchKind`/`DeadInterfaceCall` are raw; the internal dump renders display.
- **Declared publisher ends**: new `DeclaresPublish` relation with the same agreement laws as sends, derived from bus-block declarations — pinned by a no-send locus whose declared end exists with the topic's wire subject.
- **Unanalyzed bodies**: module-scoped bodies and `on_failure` handlers enter the universe as entities (failure handlers as Hooks named `Locus::on_failure(Child,Err)`) and hole out via the new `HoleKind::UnanalyzedBody` hiding CALLS|PUBLISHES|EFFECTS — the capability trio goes false instead of lying. Pinned with an indirect call in a module body + a publish in a failure body.
- **The legacy bridge**: `LegacyProjection.topology_v1_fns` records the exact legacy fn sort on the model root, validated (sorted, in-range) — the differential regains **exact** equality through the projection on top of the subset law, and Change 3 needs no side channel. (Alternative hash-transition path not taken; the named `TopologyShapeV1` contract stands.)

17 builder tests; 1216/1216 across the three crates; fmt clean.

rileyr:
This round reviewed `8d50097` — that is the **round-6 head**. All four findings here are the round-7 findings, and all four were fixed in `6dff3e1` (pushed before this review posted). Per-finding state at the current head (`01b061a`):

**1. Canonical identity.** The `name()` demangler no longer feeds `name` fields anywhere. Entity construction keeps the raw post-merge symbol as `name` and uses the demangled alias spelling only for `display` — for functions, loci, types, interfaces, declarations, and (since they had nowhere to carry a spelling before) `Topic.display` / `Group.display`. `DispatchKind` and `DeadInterfaceCall` interface names are raw too. Canary: `canonical_identity_is_alias_independent` derives the same library under `import events as p` vs `import events as db` and asserts the two models' canonical names, dense-ID order, and relation keys are byte-identical while only `display` differs. The Change-3 note is recorded: legacy artifact projection renders `display`.

**2. Declared publisher ends.** `DeclaresPublish { locus, subject, declared_topic, payload, provenance }` exists in `hale-model` with the same subject/payload agreement laws as sends, and the builder derives it from `bus { publish T; }` declarations — not from `effect_sites`. Canary: `declared_publisher_ends_survive_without_sends` is exactly the `Gateway` program above — declares `publish Orders`, never sends; `relations.declares_publish` has the row, `relations.publishes` is empty, and both facts coexist.

**3. Unanalyzed bodies.** Module-scoped functions and `on_failure` handlers now enter the universe as real entities (failure handlers as `FunctionKind::Hook` named `Locus::on_failure(Child, Err)`) and each emits a `HoleKind::UnanalyzedBody` hole hiding `CALLS | PUBLISHES | EFFECTS` — so `exact_calls`, `exact_bus_endpoints`, and `exact_effects` all go **false** in their presence rather than silently staying true. The fixture puts an indirect call in the module body and a publish **plus** an indirect call (through a fn-typed locus param) in the failure body — the symmetric coverage this round asked for, strengthened in `01b061a`.

**4. TopologyShapeV1 projection.** `LegacyProjection.topology_v1_fns` on the model root records the exact legacy fn sort (`summary.fns.keys()` order), validated like every other table (sorted, deduplicated, non-empty entries). The differential test regained its **exact** equality assertion: artifact fn sort `==` projection, on top of the universe-⊆ law. Change 3 filters through this field — no summary or AST side channel — and the `TopologyShapeV1` contract stands unchanged; no hash transition needed.

Full run at `01b061a`: 17 builder tests, 1216/1216 across hale-types + hale-model + hale-cli, fmt clean.

rileyr:
All three round-9 blockers fixed in `b2ab170`. Each behavioral fix is negative-tested — the pinning test fails with the fix disabled.

**1. Calls into unanalyzed bodies.** Correct diagnosis: the summary resolves callees only against its analyzed-body set, so a direct call to a module-scoped fn arrives `Callee::Unresolved` and fell off the end of the branch chain. The unresolved branch now falls through to the declaration-level universe: when the bare name resolves there (and no receiver is present), the builder emits the known `Direct` call edge to the unanalyzed callee, whose `UnanalyzedBody` hole still bounds reasoning past it — the edge and the hole coexist, which is exactly what "a concrete path beats a hole" requires. Pinned by `direct_call_into_unanalyzed_callee_is_a_known_edge`: analyzed `caller` → module `helper`, edge present, hole present, `exact_calls` false.

**2. Raw opaque and key-type identities.** `shape_of_type`'s fallback is now `opaque:<raw post-merge symbol>` and `key_type_of`'s enum branch returns the raw name into `AnyOfType` — display spelling no longer enters any identity. The `opaque:?` collapse is gone too: a payload that is not a bare named type gets a source-independent structural descriptor (`(Int,Bool)`, `[Int; 4]`, `bounded[T; N]`, `fn(..) -> ..`, generic and multi-segment names with raw segments), so distinct forms stay distinct contracts. Pinned by `opaque_identities_are_alias_independent_and_distinct`: the same enum imported as `p::Status` vs `db::Status` yields byte-identical contract tables and `AnyOfType("__lib_e_events_Status")` in both, and tuple/sized-array payloads get distinct descriptors.

**3. The legacy walk is shared, not reimplemented.** The legacy contraction (one Boolean, no-revisit `seen`, LIFO, MAX_STEPS) is extracted verbatim into `callgraph::legacy_via_stdlib_contraction`, and **both** `dump_topology` and the new `LegacyProjection.topology_v1_calls_via_stdlib` call it — the projected rows and the serialized rows cannot drift by construction, including the order-dependent loop bits you identified. The model's lattice relation is untouched; where the two disagree on a loop bit that is now recorded data on the model root, not silent hash drift. New validation laws: rows sorted/deduped by (from, to), and both endpoints must be members of `topology_v1_fns` (a legacy row only ever connects legacy fns). Pinned by `legacy_via_stdlib_projection_matches_the_artifact_exactly` using the #392 Router recipe to manufacture a real contracted edge: projection == artifact rows **including loop bits**, and the lattice agrees on endpoints. The full suite confirms `dump_topology` output is byte-identical after the extraction.

One scope note from Riley mid-round: strict backwards compat is not a product requirement — the `TopologyShapeV1` bridge is being kept as (a) replay-admission continuity for existing recordings and (b) the exact-differential oracle that has caught a builder bug nearly every round, and the Change-6 versioned hash transition is pre-authorized to break it cleanly.

20 builder tests, 1219/1219 across hale-types + hale-model + hale-cli, fmt clean.

rileyr:
All three round-10 blockers fixed in `220c510`, each negative-tested (pinning test fails with its fix disabled).

**1. Resolution by call shape.** Confirmed — the round-9 recovery had no receiver guard, so a typed method miss could falsely wire to a same-named free fn. The branch chain now resolves exactly as you specified: free call → raw free identity, with author-spelled qualified misses (`alias::helper`) remangled to their raw imported symbol through the inverse rename map; typed method → `RecvTy::method` in the declaration universe (where module-locus methods land); opaque receiver → the existing `UntypedReceiver` hole. A typed method that resolves nowhere (stdlib/external) emits nothing rather than a false edge — the effect frontier and stdlib contraction own those paths. Pinned by `unresolved_method_calls_resolve_by_shape` (`w.tick()` does NOT wire to free `tick`; `s.get(1)` on a module locus resolves to `Store::get`) and `qualified_calls_into_unanalyzed_imports_resolve_raw`.

**2. The BusSubject variant survives extraction.** Confirmed at all four sites — subject/payload registration, subscribe rows, send sites, and `DeclaresPublish` all decided declaredness by string lookup. The AST-side sites now branch on the variant (`Literal` never consults the topic table; `Topic`/`QualifiedTopic` resolve through it), and for send sites the summarizer's effect site now carries the syntactic form — `EffectSiteKind::Publish(Option<PublishSubject { text, literal }>)` — since `"Orders" <- v` and `Orders <- v` were indistinguishable as strings. All seven existing consumers of the effect site were updated semantics-preserving (artifact output byte-identical; full suite green). Pinned by `literal_endpoints_never_conflate_with_topics` — your exact fixture plus the send and declared-end variants: the literal endpoint keeps address `"Orders"`, `declared_topic: None`, and the `LiteralPayload` contract, while the topic endpoints keep `wire.orders` + `NamedPayload`, with both subjects present as distinct rows.

**3. One contract rule for type expressions.** `contract_of_te(&TypeExpr)` is now the single rule — bare named struct → canonical structural shape, every other form → the structural descriptor — used by declared topic payloads and explicit `of type` clauses alike (the topic path was refactored through it too, so a second rule cannot reappear). `subscribe "a" of type Int / (Int, Bool) / [Int; 4]` now yield `opaque:Int`, `opaque:(Int,Bool)`, `opaque:[Int; 4]` — three distinct contracts. Pinned by `explicit_endpoint_payloads_keep_structural_identity`.

24 builder tests, 1223/1223 across hale-types + hale-model + hale-cli, fmt clean.

PR #481, opened 2026-08-16
feat(types): model-backed artifact projection + TopologyShapeV1 corpus differential (GH #476 Change 3)

rileyr:
Change 3 of the canonical-model epic (#476): the first model-backed artifact encoder, with the epic's exit criterion enforced in-tree.

## What ships

- **`topology_projection::project_model_half(&ApplicationModel) -> String`** — renders the topology artifact's hashed model half (sorts, sealed, relations, groups, labels, phases, seeds, effects, supervision, unknowns) from the canonical model alone. **`project_shape_hash`** reproduces `TopologyShapeV1` — FNV-1a/64 over that string, exactly as `dump_topology` stamps it.
- **The corpus differential** (`hale-types/tests/topology_projection.rs`) — for every checkable corpus program, `project_model_half(derive_application_model(bundle))` must be **byte-identical** to the substring `dump_topology` hashes, and the projected hash must equal the stamped one. Plus a dense fixture interleaving every hashed section (sealed loci, supervision with a retry bound, keyed + literal endpoints, groups, stdlib contraction) and a cross-seed bundle proving the display-spelling mapping.

## Projection rules

The model keys everything by raw post-merge symbol; the legacy artifact renders author spelling. The projection is that mapping, made explicit:

- `sorts.fns` / fn-keyed sections restrict to `legacy.topology_v1_fns`; **phases deliberately do not** — the legacy phase relation is AST-derived and includes empty-bodied hooks (`run() { }`) the behavior summary never keys. This was the one divergence class the corpus sweep caught, in two recording-oriented fixtures.
- Site-grained model call/publish rows merge to the legacy endpoint grain: loop bits OR, any interface-dispatch site tags the edge, sets dedupe.
- `calls_via_stdlib` renders `LegacyProjection.topology_v1_calls_via_stdlib` — the preserved one-Boolean no-revisit walk rows — never the model's lattice rows, whose loop bits may legitimately be stronger.
- Declared endpoints render their topic's display NAME (legacy relations never carried wire subjects); undeclared endpoints render their authored pattern. Edges the model recovers into unanalyzed callees have no legacy row and are filtered.
- Typed holes + `dead_interface_calls` re-fold into the legacy `unknowns` vocabulary (`indirect_call`, `untyped_receiver_call:<callee>`, `uninhabited_interface_call:<iface>.<method>`, `computed_publish`).

## The gate is permanent until Change 6

Both derivations stay live. Any model-builder change that would alter the artifact identity — and therefore silently re-key `.halerec` replay admission — now fails the differential loudly instead. At the Change-6 versioned identity transition, the legacy derivation and `LegacyProjection` get deleted together and the artifact becomes a pure projection. (Per Riley: backwards compat is not sacred — the Change-6 transition is pre-authorized to break cleanly; until then this differential is the drift alarm.)

The differential is negative-tested: perturbing a rendering rule (hook/method inversion) fails the corpus sweep.

## Verification

- 1226/1226 across hale-types + hale-model + hale-cli; `hale fmt --check` clean.
- Corpus sweep covers >100 checkable programs including modes, supervision, seeds, sealed loci.
- No user-visible behavior change: `dump_topology` output and `shape_hash` values are untouched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,23 @@ behavior.
 
 ## Unreleased
 
+### The model-backed artifact projection (GH #476 Change 3)
+
+`hale_types::topology_projection::project_model_half(&ApplicationModel)`
+renders the topology artifact's hashed model half from the canonical
+model alone, and `project_shape_hash` reproduces `TopologyShapeV1`
+exactly - proven byte-for-byte against `dump_topology`'s own
+derivation over every checkable corpus program by a permanent
+differential gate (`tests/topology_projection.rs`). The projection
+maps the model's raw canonical identities back to the artifact's
+display spelling, merges site-grained rows to the legacy endpoint
+grain, renders `calls_via_stdlib` from the preserved legacy
+contraction rows, and re-folds typed holes + dead interface
+dispatches into the legacy `unknowns` vocabulary. Both derivations
+stay live until the Change-6 versioned identity transition: a model
+change that would silently re-key `.halerec` replay admission now
+fails the differential instead. No user-visible behavior change.
+
 ### The ApplicationModel builder - demand-gated derivation (GH #476 Change 2)
 
 `hale_types::model_builder::derive_application_model(&Bundle)` - one
--- a/crates/hale-model/src/application.rs
+++ b/crates/hale-model/src/application.rs
@@ -737,9 +737,14 @@ impl ApplicationModel {
         }
         check_sorted_keys(
             "supervises",
-            r.supervises
-                .iter()
-                .map(|x| (x.parent, x.child.clone(), &x.error_type)),
+            r.supervises.iter().map(|x| {
+                (
+                    x.parent,
+                    x.child.clone(),
+                    &x.error_type,
+                    x.authored_ordinal,
+                )
+            }),
         )?;
         for (i, x) in r.supervises.iter().enumerate() {
             let child_ok = match &x.child {
--- a/crates/hale-model/src/application.rs
+++ b/crates/hale-model/src/application.rs
@@ -792,7 +792,36 @@ impl ApplicationModel {
         }
 
         // --- labels / weights.
-        check_sorted_keys("labels", self.labels.iter().map(|l| (l.at, &l.label)))?;
+        // Labels are grouped by entity in canonical entity order,
+        // but WITHIN one entity the label order is semantic — the
+        // declared class order the artifact hashes (`zebra` declared
+        // before `alpha` renders before it) — so rows are NOT fully
+        // key-sorted (review round 11; mirrors `Function.effects`,
+        // whose declaration order was already preserved).
+        {
+            let mut prev_at: Option<crate::ids::EntityRef> = None;
+            let mut seen: std::collections::BTreeSet<(
+                crate::ids::EntityRef,
+                &String,
+            )> = std::collections::BTreeSet::new();
+            for (i, l) in self.labels.iter().enumerate() {
+                if let Some(p) = prev_at {
+                    if l.at < p {
+                        return Err(ModelError::NotCanonical {
+                            table: "labels",
+                            index: i,
+                        });
+                    }
+                }
+                prev_at = Some(l.at);
+                if !seen.insert((l.at, &l.label)) {
+                    return Err(ModelError::NotCanonical {
+                        table: "labels",
+                        index: i,
+                    });
+                }
+            }
+        }
         for (i, l) in self.labels.iter().enumerate() {
             if !ref_ok(&l.at) {
                 return Err(ModelError::DanglingId {
--- a/crates/hale-model/src/hole.rs
+++ b/crates/hale-model/src/hole.rs
@@ -58,8 +58,12 @@ impl RelationSet {
 pub enum HoleKind {
     /// A call through a fn pointer / indirect site.
     IndirectCall,
-    /// A method call on a receiver the summarizer cannot type.
-    UntypedReceiver,
+    /// A method call on a receiver the summarizer cannot type. The
+    /// dispatched method name is MACHINE data (the legacy artifact's
+    /// `untyped_receiver_call:<callee>` row requires it), so it
+    /// lives here in the kind — never parsed back out of the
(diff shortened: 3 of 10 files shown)


rileyr:
All five round-11 blockers fixed in `e0d859d`. They were one family — the legacy artifact hashes *authored* order in places the canonical model deliberately normalizes — and each fix recovers the authored fact explicitly instead of hoping canonical order coincides. Every behavioral fix is negative-tested (pin fails with the fix disabled).

**1. `via_interface` selects the greatest authored site.** Confirmed — the projection's overwrite over canonical rows made the lexicographically last interface win. Model call rows already carry authored site ordinals (round 6's `dispatch_group` work), so the projection now keeps `(site, display)` and last-authored wins, exactly reproducing the legacy source-order walk. Pinned by `via_interface_tie_follows_authored_order`: dispatch through `ZIface` then `AIface` on one `(from, to)` pair — V1 and the projection both select `AIface`.

**2. Label order is declaration order, in the model itself.** The builder's `BTreeSet` flattening was the bug — it lexicalized `["zebra","alpha"]`. `LabelRow` rows now preserve per-entity class order (render_effects_named order), and the schema law changed to match: grouped by entity in canonical entity order, order *within* an entity semantic, duplicates rejected — mirroring `Function.effects`, which already preserved declaration order. New `hale-model` canary `label_order_is_semantic_within_an_entity` pins all three sides of the law. Pinned end-to-end by `label_order_follows_declaration_not_lexical` with your exact `effect zebra; effect alpha;` program.

**3. Supervision ties keep authored order via `authored_ordinal`.** I first tried your provenance suggestion and it failed honestly: sourceless test bundles resolve spans to synthetic records, so provenance cannot carry authored position everywhere the differential runs. `Supervises` rows instead gain `authored_ordinal` (payload, not canonical key) — handler order is an authored fact the artifact hashes — and the projection orders by it before reproducing the legacy stable sort by `(locus, child)`. Pinned by `supervision_ties_keep_authored_order` (`ZTrouble` authored before `ATrouble` stays first). A bonus from your counterexample: two `on_failure` handlers per child check clean, and the corpus provider harvests the pin fixture, so the shape is corpus-covered from now on.

**4. One raw→display map, applied everywhere `name()` ran.** Built over every declaration table (loci, topics, types, interfaces, groups, opaque declarations, functions) and applied to undeclared subject patterns, `SupervisedRef::External`, and `error_type` — including the collision case: a literal subject spelled like an imported raw symbol demangles under V1 and now under the projection. Pinned by `literal_subject_colliding_with_raw_symbol_demangles` (`subscribe "__lib_x_kv_Item"` renders `kv::Item` in both).

**5. `HoleKind::UntypedReceiver { callee }`.** The callee is machine data and now lives in the kind; the backtick parse is gone, so a producer with different witness prose can no longer change a projected identity. The type system enforces this one — there is no string path left to negative-test.

31 projection+builder tests, 34 model canaries, 1231/1231 across the three crates, fmt clean.

rileyr:
Both round-12 blockers fixed in the new head, both negative-tested.

**1. Exact-renames display scope.** The map now inserts `FunctionKind::Free` functions only — a free fn's raw name is exactly an `import_renames` key, while method/hook/mode identities never are (the mangler leaves them alone, so legacy `name()` leaves them verbatim). Every other table already held only top-level declarations. Pinned by `method_shaped_literal_subject_does_not_demangle`: `subscribe "__lib_x_kv_Store::bump"` stays verbatim in both derivations, sitting right next to the round-11 type-symbol collision pin that must demangle — the pair documents the exact boundary of the rule.

**2. Labels and effects filter through `legacy.topology_v1_fns` again.** Correct on both counts — including that it doesn't fire with today's builder. Since no bundle-derived fixture can construct the case yet, the pin (`labels_and_effects_are_restricted_to_the_v1_universe`) is built directly against the public projector with a hand-assembled model: a non-V1 function carrying both a label and an effect set appears in no fn-keyed section, and enrolling it into the V1 universe surfaces it in both — the filter is the universe, not the function. Phases deliberately remain unfiltered (the legacy phase relation is AST-derived and includes empty-bodied hooks; corpus-pinned in the original submission).

9 projection tests, 1233/1233 across the three crates, fmt clean.

rileyr:
Both round-13 blockers fixed in the new head, both negative-tested.

**1. Sealed is raw-order, display-value.** Exactly as diagnosed — the projection's `BTreeSet` of displays re-sorted what the legacy encoder orders by raw name. `entities.loci` is already canonically raw-sorted, so the fix is to collect sealed displays into a `Vec` without re-sorting. Pinned by `sealed_order_is_raw_not_display` with your requested two-import shape: raw `__lib_a_pack_Z` < `__lib_b_pack_A` with displays `z::Z` / `a::A`, asserting the artifact renders `["z::Z", "a::A"]` — non-lexical display order as the correct output — and byte equality.

**2. `retry_bound` is `i64`, as written.** Confirmed check-clean divergence: `for 4294967296` projected as `0` through the `as u32` cast. `SupervisionPolicy.retry_bound` is now `Option<i64>` through schema, builder, and projection — the literal as authored, exactly what the legacy artifact hashes. No source rejection, per your note on scope. Pinned by `retry_bound_keeps_the_full_literal` asserting the full-width literal in both derivations.

11 projection tests, 1235/1235 across the three crates, fmt clean.

rileyr:
Round-14 blocker fixed in the new head, negative-tested.

**`authored_ordinal` joins the canonical key.** Confirmed on both counts: the builder's `(parent, child, error_type)` map silently kept only the last duplicate-signature handler, and the validator's matching canonical key made the model structurally unable to represent both rows. The key is now `(parent, child, error_type, authored_ordinal)` in the schema law, the validator, and the builder's collection map — one row per declaration, exactly the legacy grain. The projection needed no change: it already orders by authored ordinal before the legacy stable sort, so both rows render in authored order (`restart` then `quarantine` in your fixture).

Pinned by `duplicate_signature_supervision_rows_both_survive` — your exact program, asserting both rows appear in authored order plus byte equality — and the corpus provider harvests the fixture, so the shape stays corpus-covered. No checker change, per the scope note.

12 projection tests, 1236/1236 across the three crates, fmt clean.

PR #482, opened 2026-08-16
feat(model+types+cli): ClaimIr — every law surface lowered to one typed vocabulary (GH #476 Change 4)

rileyr:
Change 4 of the canonical-model epic (#476): **`ClaimIr` lowering only** — one typed variant per law form across every surface, with the old evaluators fully active and authoritative.

## What ships

- **`hale-model::claim_ir`** — the schema. One variant per law form: the eight claims-block forms (`ForbidReaches`, `OnlyEdges`, `Bound`, `RequireEndpoint`, `RequireSealed`, `RequireAttributed`, `Cover`, `Count`), the annotation surfaces (`EffectForbid`/`EffectPublishSet`/`EffectCauses`/`EffectOnly`/`NoPanic` for `@effects`/`@no_panic`, `DependsSet` for `@effects(depends:)`, `PhaseEffects`, `AllocBudget` + `QuantBudget` for both `@budget` families), and the four deployment-plan forms. Reference doctrine matches the model: raw canonical spelling + author display + optional resolved id — `None` is lawful residue (Change 5's `invalid` verdict consumes it), and fleet targets stay name-level until Change 7's `FleetModel`. Rows keep **authored order** via contiguous ordinals (Change 5's diagnostics-parity differential walks outcomes and rows in lockstep); `ClaimIrTable::validate(&ApplicationModel)` enforces ordinal contiguity, in-range ids, named rows, and provenance.
- **`claims::enumerate_clauses`** — the clause-collection phase of `claims_report_inner` (world tier + constitution adoption + library-tier attribution) extracted into one shared function, so the lowering walks **exactly** the clauses the evaluator walks. Same one-authority move as the Change-3 shared contraction walk; the evaluator's behavior is unchanged (full claims/constitutions/library suites green).
- **`claim_lowering::lower_claims(bundle, model)`** — program surfaces: claims through the shared enumeration with origins (`Main` / `Constitution { name }` / `Library { alias }` — origin matches `ClaimOutcome::source`), annotations through a declaration-order AST walk. Total over parseable programs; never drops a law.
- **`fleet::lower_plan_claims`** — plan rows, one `ClaimIr` row per SENTENCE (a multi-bound `CountSpec` stays one conjunctive law; a multi-verb or verbless spec is a structured issue), wired behind the `HALE_MODEL_TRACE` demand surface at the fleet check site.

## What deliberately does NOT lower

`@effects(is: {…})` — carries is a classification **fact**, not an obligation; the model already records it as `labels` rows and no judgment evaluates it. Pinned in the exact-form test.

## Verification

- **Evaluator parity over the corpus**: for every corpus program, the claims-family rows equal `claims_report_with_identities` outcomes on count, name, authored order, and constitution source — and the lowering is total (no panics) and lawful (`validate`) everywhere, with a vacuity guard (>10 corpus programs must actually carry claims).
- Exact-form fixtures for every claims-block and annotation variant with resolved-id assertions; fleet unit test; `hale-model` canaries (ordinal contiguity, dangling id, unresolved-is-lawful).
- 1241/1241 across hale-types + hale-model + hale-cli; fmt clean; no user-visible behavior change.

Next per the epic: Changes 5a–5e migrate the judgment families onto `ClaimIr` × `ApplicationModel` — the epic's largest tranche, now with a stable typed target on both sides.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,31 @@ behavior.
 
 ## Unreleased
 
+### ClaimIr - every law surface, lowered (GH #476 Change 4)
+
+`hale-model` gains `ClaimIr` - one typed variant per law form across
+every surface the language has grown: the eight claims-block forms
+(reachability, boundary grants, path bounds, endpoint existence,
+sealing, attribution, coverage, cardinality), constitution clauses
+with recorded origin, library-tier claims with alias attribution,
+the annotation surfaces (`@effects(none/publish/causes/only)`,
+`@no_panic`, `@effects(depends:)`, `@phase_effects`, `@budget` in
+both alloc and quantitative forms), and deployment-plan claim rows
+(name-level until Change 7's FleetModel).
+`hale_types::claim_lowering::lower_claims(bundle, model)` lowers
+the program surfaces through the evaluator's OWN clause enumeration
+(`enumerate_clauses`, extracted from `claims_report_inner` so two
+walks cannot drift); `fleet::lower_plan_claims` lowers plan rows.
+Lowering is total over parseable programs (unresolved refs keep
+raw+display spellings with no id - Change 5's `invalid` residue),
+rows keep authored order, and a corpus-wide differential holds the
+claims-family rows equal to the evaluator's outcomes on count,
+name, order, and constitution source. Lowering only: the old
+evaluators stay active and authoritative; nothing consumes these
+rows yet. `@effects(is:)` deliberately does not lower - carries is
+a classification fact the model already records as labels.
+
+
 ### The model-backed artifact projection (GH #476 Change 3)
 
 `hale_types::topology_projection::project_model_half(&ApplicationModel)`
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -78,6 +78,7 @@ version = "0.17.0"
 dependencies = [
  "hale-codegen",
  "hale-lsp",
+ "hale-model",
  "hale-syntax",
  "hale-types",
  "libc",
--- a/crates/hale-cli/Cargo.toml
+++ b/crates/hale-cli/Cargo.toml
@@ -11,6 +11,7 @@ name = "hale"
 path = "src/main.rs"
 
 [dependencies]
+hale-model = { path = "../hale-model" }
 hale-syntax = { path = "../hale-syntax" }
 hale-types = { path = "../hale-types" }
 hale-codegen = { path = "../hale-codegen" }
--- a/crates/hale-cli/src/fleet.rs
+++ b/crates/hale-cli/src/fleet.rs
@@ -1393,106 +1393,108 @@ fn fnv(s: &str) -> String {
 }
 
 /// GH #476 Change 4 — lower the plan's claim rows to `ClaimIr`.
-/// Fleet targets are PLAN-level names (instances, plan groups, wire
-/// subjects); they resolve against a typed `FleetModel` at Change 7,
-/// so every row lowers name-level here and the old evaluator stays
-/// authoritative. A `CountSpec` with several bounds present lowers
-/// one row per bound (eq, then max, then min) — each is its own law.
+/// ONE row per `ClaimSpec`, mirroring the evaluator's
+/// one-name/one-sentence rule (a claim name identifies one sentence
+/// and one verdict): a spec with zero or several verbs set lowers to
+/// a structured [`hale_model::LoweringIssue`], never to a split or
+/// silently dropped law, and a `CountSpec`'s eq/max/min bounds stay
+/// ONE conjunctive law. Fleet targets are PLAN-level names
+/// (instances, plan groups, wire subjects); they resolve against a
+/// typed `FleetModel` at Change 7, so every row lowers name-level
+/// and the old evaluator stays authoritative.
 pub fn lower_plan_claims(
     claims: &[ClaimSpec],
 ) -> hale_model::ClaimIrTable {
     use hale_model::{
-        ClaimIr, ClaimIrTable, ClaimOrigin, ClaimRow, CountCmpIr,
-        ProvenanceId,
+        ClaimIr, ClaimIrTable, ClaimOrigin, ClaimRow, ProvenanceId,
     };
     let mut table = ClaimIrTable::default();
-    let push = |table: &mut ClaimIrTable,
-                    name: &str,
-                    law: ClaimIr| {
+    let mut prov = |table: &mut ClaimIrTable, name: &str| {
         let pid = ProvenanceId(table.provenance.records.len() as u32);
         table.provenance.records.push(
             hale_model::Provenance::Synthetic {
                 origin: format!("deployment plan claim `{}`", name),
             },
         );
+        pid
+    };
+    for c in claims {
+        let verbs = [
+            c.forbid_reaches.is_some(),
+            c.require_subscribes.is_some(),
+            c.require_publishes.is_some(),
+            c.count_publisher_instances.is_some(),
+            c.count_subscriber_instances.is_some(),
+            c.only_edges.is_some(),
+        ]
+        .iter()
+        .filter(|v| **v)
+        .count();
+        if verbs != 1 {
+            // The evaluator's one-name/one-sentence rule: this spec
+            // is not one law. Structured invalidity, not a split.
+            let pid = prov(&mut table, &c.name);
+            table.issues.push(hale_model::LoweringIssue {
+                message: format!(
+                    "plan claim `{}` sets {} verbs — a claim is ONE \
+                     sentence with one verdict",
+                    c.name, verbs
+                ),
+                provenance: pid,
+            });
+            continue;
+        }
+        let law = if let Some(fr) = &c.forbid_reaches {
+            ClaimIr::FleetForbidReaches {
+                from: fr.from.clone(),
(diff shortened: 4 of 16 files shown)


rileyr:
All six round-15 blockers fixed in the new head; behavioral fixes negative-tested.

**1. Invalidity is preserved.** `ClaimIrTable` gains `issues: Vec<LoweringIssue>`; every diagnostic the shared enumeration reports (unknown/cyclic constitution, illegal library-tier `adopt`, duplicates, collisions) lowers to a structured issue with provenance. Pinned by `invalid_law_selection_becomes_issues` — your `adopt Missing` program now lowers to zero rows plus one issue naming `Missing`.

**2. Fleet one-name/one-sentence.** `lower_plan_claims` emits exactly one row per `ClaimSpec`: a spec with zero or several verbs lowers to a structured issue mirroring the evaluator's refusal, and `FleetCountInstances` keeps `eq`/`max`/`min` as one conjunctive law. The unit test now pins the multi-verb and verbless specs as issues and the two-bound count as one row — the previous test's four-row expectation was, as you said, lowering a claim the evaluator rejects; gone.

**3. Typed effect classes.** The model gains `entities.effect_classes` — `EffectClassDecl { name, declared, composition }` with the normalized atomic expansion for composed classes, populated from the same `declared_of`/`defs_of` facts the evaluators read. `EffectClassRef` is now `{ class: Option<EffectClassId>, builtin, name, provenance }`. Pinned by `effect_class_vocabulary_is_typed`: your exact declared-vs-typo pair now differ (`declared: true` vs `false`), `effect io = { syscall, block }` carries `["block", "syscall"]`, and built-ins never resolve into the user table (validated).

**4. Publish sets are topic-space.** `EffectPublishSet` lowers `Vec<TopicIrRef>` through the same topic-resolution path as claim topic refs. Pinned by `publish_set_is_topic_space` — your `Orders`/`wire.orders` program resolves the entry to the declared topic.

**5. Per-reference provenance.** `GroupRef`, `TopicIrRef`, `PhaseIrRef`, `SeedIrRef`, `SubjectIrRef`, `EffectClassRef` (and grants via their topic refs) each carry their own `ProvenanceId`, interned at the reference span (`Ident.span`, `TopicRef.span`, `class_span`, …); annotation class lists, which carry no spans in the AST, anchor at the annotation. Pinned by `references_carry_their_own_provenance`.

**6. Entity-sourced references + agreement law.** When an id resolves, the reference's `name`/`display` come from the model entity — which also fixes the real imported-method display bug you found (`__lib_x_kv_Store::bump` now displays `kv::Store::bump`). `validate` enforces the agreement on every resolved ref (`ClaimIrError::NameDisagreement`), canary-pinned in hale-model, and end-to-end pinned by `resolved_refs_are_entity_sourced`.

8 lowering tests + fleet unit + 2 new schema canaries; corpus parity differential still green (now also exercising validate's agreement law on every corpus program); 1247/1247 across the three crates, fmt clean.

rileyr:
Both round-16 blockers fixed in the new head, both negative-tested.

**1. `EffectClassDefinition` is explicit.** Exactly your suggested shape: `Atomic` / `Composed { atoms }` / `InvalidCycle`. The builder's expansion now reports in-path re-entry and marks the cycle instead of silently returning — `effect a = { b }; effect b = { a };` produces `InvalidCycle` on both rows, structurally distinct from `effect a;`, preserving the invalidity the checker diagnoses at the declaration (frontier resolves a cycle to no effect, so contracts naming it are vacuous). The validate law moved onto the enum: only `Composed` carries atoms (sorted, deduplicated, non-self). Pinned by `cyclic_effect_class_definitions_stay_invalid`.

**2. `TopicSelector` for annotation topics.** The typed selector you asked for: `Exact(TopicIrRef)` for qualified spellings, `Unqualified { name, candidates }` for bare names — with candidates computed by the evaluator's own `topic_tail` rule (the function is shared, not reimplemented), sorted/deduplicated/range-validated. One deliberate strengthening beyond the minimal fix: an unqualified spelling is *always* the candidate-set selector, never shortcut to Exact on a local hit — the evaluator doesn't privilege exact matches for unqualified names, so a local topic and a same-tailed import both match, and the selector says so. Pinned by your requested fixture shape in `unqualified_imported_publish_ref_gets_candidates`: the annotation is authored inside the imported seed, `Recalled` tail-matches the merged `__lib_r_main_Recalled`, and the multi-candidate semantics are representable where a single `Option<TopicId>` was not.

10 lowering tests, 1249/1249 across the three crates, fmt clean.

rileyr:
Round-17 blocker fixed in the new head, negative-tested.

**One shared `BusSelector` for both bus-set surfaces.** `TopicSelector` is gone; `@effects(publish:)` and `@effects(depends:)` now lower through the same selector: `Exact(TopicIrRef)` for qualified alias paths, and `Match { name, topics, subjects }` for every other spelling — candidate sets over declared topics *and* wire subjects, computed with the evaluator's own exact-or-trailing-name rule (`topic_tail` is shared, not reimplemented, same as `topic_ref_matches` parity in round 16). This closes both mismatches:

- your literal certificate `@effects(publish: { "audit.log" })` now resolves as a subject candidate (`subjects: [audit.log]`, `topics: []`) — a valid law, not residue. Pinned by `literal_publish_subject_resolves_as_subject_candidate`.
- an imported seed's unqualified `@effects(depends: { "Recalled" })` tail-matches the merged `__lib_r_main_Recalled` — pinned by `unqualified_imported_depends_entry_resolves` with the annotation authored inside the imported seed, per your fixture request.

Validation covers the selector in both positions (sorted, deduplicated, in-range candidate sets, per-reference provenance). The now-dead `SubjectIrRef` exact-lookup vocabulary is deleted rather than left as a second path. Negative control: disabling subject candidates fails the literal pin.

12 lowering tests, 1251/1251 across the three crates, fmt clean.

rileyr:
Round-18 blocker fixed in the new head.

**One `Match` shape, candidates from `topic_ref_matches` directly.** Exactly the clean fix you described: the `contains("::")` guess is gone, `BusSelector` collapsed from an enum to the single candidate-set struct `{ name, topics, subjects, provenance }`, and both `publish:` and `depends:` entries build their topic and subject candidate sets by calling `effects::topic_ref_matches` itself — exact string equality first, trailing-name second, for every spelling. No syntax reconstruction from string contents, and no `Exact` arm to under-represent the evaluator's qualified-path permissiveness. Making paths exact stays out of scope per your note (it would be an evaluator change plus an AST change to preserve literal-vs-path syntax).

Pins added, both shapes you requested:
- `double_colon_literal_subject_resolves_on_both_surfaces` — `"audit::log"` as a `publish:` entry and a `depends:` entry, resolving as a subject candidate on both (your fixture, extended to the second surface).
- `qualified_path_tail_matches_same_tailed_imports` — `"a::Recalled"` with two same-tailed imports produces both topics as candidates, pinning the current qualified-path behavior as-is.

Non-blocking cleanup done too: the PR description's "one row per bound" sentence now states the conjunctive single-law semantics the implementation has had since round 15.

14 lowering tests, 1253/1253 across the three crates, fmt clean.

rileyr:
Both round-19 blockers fixed in the new head.

**1. Candidate sets validate as derived data.** The matching rule now lives in `hale-model` as the canonical definition — `bus_ref_matches` / `bus_topic_tail` — and `hale-types::effects::topic_ref_matches` *delegates* to it (a delegation, not a second implementation, so the evaluator and the schema cannot drift while both are live). `sel_ok` recomputes the expected topic and subject candidate sets over the model's tables and requires exact equality: your `"Orders"`-with-`payments_topic_id` example is `NameDisagreement`, and so is an omitted same-tailed candidate. Both canaries added (`bus_selector_candidates_must_match_the_name` pins the widening and the narrowing case, with the lawful full same-tailed set as the positive control).

**2. Provenance records resolve.** `ClaimIrTable::validate` now applies the same record-contents boundary `ApplicationModel::validate` draws: a `SourceId` out of range or an inverted span is `ClaimIrError::InvalidProvenanceRecord`. Both of your examples are canary-pinned (`claim_table_provenance_records_must_resolve`).

A side benefit of the recompute-and-compare law: the corpus parity differential now proves lowering-vs-validation agreement on every candidate set for every corpus program, since both sides call the one canonical rule.

38 model canaries, 14 lowering tests, 1255/1255 across the three crates, fmt clean.

rileyr:
Round-20 blocker fixed in the new head.

**The builtin discriminator is validated against a canonical vocabulary.** Took your first option, since it keeps the reference shape uniform for renderers: `BUILTIN_EFFECT_CLASSES` / `is_builtin_effect_class` now live in `hale-model`, `validate` holds every `EffectClassRef.builtin` to exact agreement with the name (`NameDisagreement` on either mismatch direction — both your examples are canary-pinned), and `QuantDimIr::UserClass` additionally rejects a built-in reference (third canary, per your list). The lowering computes the flag through the same function, so producer and validator share one authority.

The one seam this creates — `hale-model` cannot see the compiler's `EffectClass` — is pinned by a two-directional conformance test in `hale-types` (`builtin_vocabulary_mirrors_the_compiler`): every model builtin must be a compiler builtin with the same spelling, and every compiler builtin must appear in the model's list. Adding a built-in class without updating both lists fails that test, which is the same integration-seam discipline `class_universe` documents for `only:`.

39 model canaries, 15 lowering tests, 1257/1257 across the three crates, fmt clean.

PR #483, opened 2026-08-17
feat(judgment): the reachability family on the canonical model — Change 5a (GH #476)

rileyr:
Change 5a of the canonical-model epic (#476): the first judgment family — reachability (`forbid reaches`) + holes — migrated onto `ClaimIr` × `ApplicationModel`, meeting the epic's migration bar: **diagnostics parity, old/new differential over the corpus, negative controls**. The old evaluator stays live and authoritative until Change 9.

## What ships

- **`judgment::judge_forbid_reaches(&ClaimIrTable, &ApplicationModel, source_bases)`** — verdicts + diagnostics for every `forbid reaches` row, byte-identical to `claims.rs`'s evaluator: witness paths (interior stdlib spellings included), crossing-edge/delivered-at/destination-decl related spans with the `is_bundle_fn` gating, the full validation pass (unknown groups with did-you-mean, `effects()`-in-source, undeclared classes, avoiding-overlap, projection vacuity, duplicate-name pre-pass), fail-closed holes with the evaluator's exact spellings, `during`/`avoiding`/`via` semantics, and the `Saturated`/`Uncertified`/`Holds`/`Violated`/`Invalid` verdict mapping.
- **The walk** reuses `model_graph::search` with a two-kind vertex: `User(FunctionId)` over model rows (site-grained calls; publish × subscribe composition; `member_of ∩ topology_v1_fns` group projection; `phase_of` for `during`; typed holes), and `Interior(entry, node)` over a new **`LegacyProjection.stdlib_absorption`** sidecar — the interior *graph* the evaluator's merged-summary walk traverses through stdlib bodies. It is kept as a graph, not a flattened summary, because BFS layering decides hole-vs-hit timing; displays demangle through the shared `stdlib_bodies` table so `std::io::tcp::Stream::send` renders as the evaluator spells it. Quarantined legacy-parity data, deleted at Change 9.
- **Model additions**: `Function.direct_effects` (the `effects(C)` destination test is DIRECT on purpose; computed by calling the evaluator's own `claims::direct_effects`) with a validate law; `stdlib_absorption` validate laws (non-strict `(from, site)` order — one dispatch site fans to several stdlib conformers; in-range targets; non-empty node lists).

## The gate

`tests/judgment_reachability.rs` — for every corpus program carrying the family: verdict parity by claim name AND byte-equal diagnostic sequences (message + span, table pre-pass first, then per-row in authored order). The differential drove the whole implementation: it surfaced and then retired six divergence classes (span anchors, related diags, the validation pass, vacuity, stdlib displays, BFS hole timing) before going green.

Negative controls: clearing `relations.calls` flips a Violated claim to Holds; clearing `model.holes` flips an Uncertified claim to Holds — the verdicts provably derive from the rows the model claims they do.

1260/1260 across the three crates; fmt clean; no user-visible behavior change.

Next per the epic: 5b (boundary grants / `only edges`) on the same engine skeleton.


rileyr:
All four foundations corrected at 25abdf3 (children rebased on top).

**Group projection = `ResolvedGroup::fn_set`.** The judgment's group universe now mirrors the evaluator's exactly: member loci contribute through the summary/v1 universe, and member FREE fns enroll unconditionally — including empty and declaration-only fns that never reach the summary. Pinned by `declaration_only_free_fns_stay_in_the_universe` (a declaration-only free fn in the source group flips the verdict when dropped). Roots iterate in `FnKey` order (locus `None` before `Some`), matching the evaluator's map order.

**Stdlib interiors can satisfy the destination.** The model gained a `stdlib_absorption` interior-graph sidecar (`AbsorbedNode { display, carries, direct_effects, events }` + entry facts), and the walk now runs over a two-kind vertex (`V::User(FunctionId) | V::Interior(entry, node)`), so `forbid reaches(G, effects(C))` fires when the only C-classified site lives inside an absorbed stdlib body. Truncated absorption lowers to a step-ceiling hole rather than silent closure. Pinned by `stdlib_interior_effect_sink_is_found` (the Router recipe — alloc lives interior-only; the pin also caught a missing crossing-call diagnostic, fixed by carrying `entry_provenance` on entry steps).

**Holes consume by hides-mask.** Hole relevance is now `h.hides.intersects(<relation family mask>)` — the kind only selects the wording, with `FnHole::Other` as the fallback rendering. A typed hole variant added later can no longer slip past a family it hides. Pinned by `hole_before_crossing_refuses_first`.

**`InvalidCycle` classes refuse evaluation.** A cyclically-defined class resolves the claim to `Invalid` before any target-set construction, never to an empty (vacuously-holding) set.

Differential green over the corpus with negative controls per family.

rileyr:
Round 2 fixed at 6a48f07 (children rebased on top).

**Mask-driven hole consultation, all three sites.** The judgment-side hole value now retains the model's `RelationSet`, and each consultation site tests it against the family that site walks: the `via_calls` scan triggers on `hides ∩ CALLS`, the `via_bus` scan on `hides ∩ PUBLISHES` (any kind — an `UnanalyzedBody` now refuses a bus walk instead of letting it exhaust known publish rows into Holds), and a NEW scan for `effects(C)` destinations triggers on `hides ∩ EFFECTS` (a reachable fn whose effects are hidden judges Uncertified even when no known row carries the class). Kind selects wording only. The converse is pinned too: a PUBLISHES-only hole no longer poisons a calls-only walk. Pins: `bus_walk_fails_closed_on_publishes_hiding_hole`, `publishes_only_hole_does_not_poison_a_calls_walk`, `effects_destination_fails_closed_on_effects_hiding_hole`.

**Truncation is saturation.** `AbsorbedEvent::Truncated` now sets a distinct signal beside the hole halt; an `Uncertified` search result caused by truncation maps to Violated, exactly like `Search::Saturated` — message unchanged, machine verdict now matches the evaluator. Pin: `absorption_truncation_is_violated_not_uncertified`.

**Every publish site consumes an ordinal** (the shared #484/#486 blocker, fixed here in the builder): known-subject rows and computed-subject holes now share one source-order ordinal space — a computed publish no longer leaves the counter untouched for its successor. Pin: `every_publish_site_consumes_an_ordinal`.

Workspace 2726/2726 at this head; corpus differentials byte-green.

rileyr:
Round 3 fixed at 87311d3 (children rebased).

**Authored-order hole selection.** The 5a hole index retains `authored_site`, and each space selects its earliest relevant hole (min authored site among family-intersecting holes) instead of inheriting the model's canonical `(at, kind, reason)` sort. Your exact scenario is the pin: `call_holes_refuse_in_authored_order` — an `UntypedReceiver` at site 0 beside an `IndirectCall` at site 1 now reports the untyped-receiver diagnostic, though `IndirectCall` sorts first canonically.

**Subject-grain SUBSCRIBES holes.** The hole index carries non-function anchors: a hole at `Subject(S)` hiding SUBSCRIBES refuses bus fan-out through S at both user vertices (by subject id) and interior publish events (by text, indexed under both the wire pattern and the declared topic name). Pin: `subject_subscribes_hole_fails_bus_walk_closed`, with your first counterexample verbatim — publish, no known subscriber row, subject hole ⇒ Uncertified, not Holds.

**Truncation at the frontier, prefix preserved.** The builder now materializes every discovered-but-unwalked absorption node with a lone `Truncated` event — which also closed a latent bug your review surfaced indirectly: interior targets referencing unwalked nodes previously pointed past `nodes.len()`. The judgment no longer halts on `Truncated`: the flag defers, the entire known prefix is searched, and only a `NotFound` search surfaces the step-ceiling Violated. Pin: `truncation_does_not_hide_a_known_witness` (a concrete alloc witness in the prefix wins — no "exceeded" diagnostic), alongside round-2's `absorption_truncation_is_violated_not_uncertified` (still green).

Workspace 2726/2726 at this head; corpus differentials byte-green.

rileyr:
Round 4 fixed at 86249ae (children rebased).

**Subscriber holes cover by the delivery predicate.** Subject-grain SUBSCRIBES holes are kept as patterns, and coverage uses the exact same exact-or-wildcard predicate known delivery applies to subscription rows (`pattern == written || (pattern.contains("**") && wildcard_match(...))`) — your counterexample is the pin: a hole at `audit.**` hiding SUBSCRIBES refuses a bus walk publishing to `audit.event` (`wildcard_subscriber_hole_covers_matching_publishes`). Both spellings a publish is known by are tested (topic name and wire subject), and the interior-event check no longer requires the subject to resolve in the known subject sort before holes are consulted. The shared helper `subject_hole_covers` is the one predicate for all three bus-composing engines (#484 and #486 adopt it in their own commits).

rileyr:
Round 5 fixed at c462e8e (children rebased).

**One two-grain bus-hole index.** `BusHoles` is now the single shared relation-hole lookup for every judgment composing publishes with subscribers: subject-grain holes as covering patterns (round 4's delivery predicate) AND topic-grain holes keyed by topic name, with each publish projected through every spelling it is known by (topic name and wire subject). A hole at `Topic(T)` hiding SUBSCRIBES now refuses the reachability bus walk exactly as `judge_endpoints` treats it — the same canonical hole has the same reach in every judgment. #484, #485 (`endpoint_unknown` now delegates to the same index), and #486 adopt it in their own commits. Pin: `topic_grain_subscriber_hole_fails_bus_walk_closed`.

**`entry_provenance` is range-checked.** `ApplicationModel::validate` rejects a dangling `StdlibAbsorption.entry_provenance` — the reachability judgment renders it as the authored crossing span, and a dangling id would silently render 0..0. (The inverted-ForeignSpan validation lands on #487, where the variant lives.)

rileyr:
Round 6 fixed at 01a7c70 (children rebased).

**Typed identities.** `BusHoles::blocks` now takes the model's typed inputs — `declared_topic: Option<TopicId>` matched only against topic-grain holes, and the publish's WIRE pattern covered only by subject-grain holes. Strings never cross the divide. `AbsorbedEvent::Publish` gained `declared_topic` (the builder resolves it with the same syntactic-form rule user publish rows follow: literal ⇒ wire address even under a name collision), so stdlib interiors carry the same typed identity. Pin: `topic_hole_does_not_block_literal_publish` — the literal `"Orders"` / topic `Orders` collision — beside the existing topic-grain pin for the declared direction.

**Monotonic deferral.** A set-level subscriber hole has no authored position relative to the known subscriber rows, so it no longer preempts them: both bus-hole sites defer to a flag instead of halting, known subscriber edges still compose, a found counterexample stays Violated, and the flag downgrades only a would-be Holds to Uncertified after the search. Pin: `known_bus_violation_survives_subscriber_hole` (your F → H example verbatim).

rileyr:
Round 7 fixed at 00f3acb (children rebased).

**The hole shape matrix is closed.** `allowed_hole_families` in hale-model defines, per (anchor grain, kind), the families a hole may hide; `ApplicationModel::validate` rejects everything else, plus any hides bit outside the new `RelationSet::ALL_KNOWN`, plus an `authored_site` on a non-Function anchor (set-level holes have no authored position — round 6's deferral doctrine is now a validated shape law). Your two counterexamples are pinned: `RelationSet(1 << 31)` and the locus-grain endpoint hole are both refused (`hole_shape_matrix_is_closed`, four rejected shapes plus a lawful-shape control — which also exercises the pre-existing capability-contradiction law: an endpoint hole requires `exact_bus_endpoints` lowered). Extending the matrix is a reviewed schema change made together with the judgment/BusHoles support that consumes the new shape. One correction the corpus caught: the call-hole kinds hide CALLS ∪ EFFECTS (the builder's real shape — an unfollowable call also hides the effects beyond it), not CALLS alone.

**Absorbed-publish identity is validated.** A present `declared_topic` must be in range AND its topic's name must equal the event's delivery spelling — your both-directions counterexamples (dangling `TopicId(999)`, `"Unrelated"` vs `orders_id`) are refused. Pin: `absorbed_publish_identity_is_validated`. Set-level test holes now use the matrix-lawful `DynamicEndpoint` kind throughout the train.

rileyr:
Round 8 fixed at cace916 (children rebased).

**Consumed-only hole shapes.** Fn-grain `UnanalyzedBody` drops SUBSCRIBES (set-level knowledge; no engine consults the fn-grain bit — the shape would be a valid invisible hole, now pinned as rejected). Set-level PUBLISHES holes poison every judgment whose projection requires publish completeness: `BusHoles` records the first publisher-incomplete spelling, and a `via { bus }` walk with no known counterexample judges Uncertified — an unknown publisher may create an edge the composition cannot see (#484/#486 adopt in their commits; #485's endpoint judgments already consumed the bit). Pin: `publisher_hole_fails_bus_walk_closed`.

**The authored-site event partition.** Site-shaped holes (IndirectCall / UntypedReceiver / OpenInterface / ComputedSubject) REQUIRE their ordinal; whole-body holes must not carry one; one (function, site) in call space holds resolved alternatives XOR dead interface rows XOR one typed call hole, and in publish space a known row and a computed-subject hole are exclusive. Your crossing-call-plus-IndirectCall-at-site-0 example is refused at validation. Pin: `authored_site_partition_is_validated` (all four shapes).

**Absorption residue joins the exactness account.** CallHole ⇒ `!exact_calls ∧ !exact_effects`, PublishHole ⇒ `!exact_bus_endpoints`, Truncated ⇒ calls/publishes/effects all inexact — derived by the builder and enforced by the validator, so the dual accounts cannot disagree wherever the hole lives. Pin: `absorption_residue_lowers_capabilities`. (One canary fix rode along: the exactness canary's IndirectCall hole now carries its authored site, per the new partition law.)

rileyr:
Round 9 fixed at 3b5d9f0 (children rebased).

**Required and allowed masks.** `allowed_hole_families` returns `(required, allowed)` and validation demands `required ⊆ hides ⊆ allowed`. Your EFFECTS-only IndirectCall is pinned as rejected: judgments select holes by family, so a call hole that drops its CALLS bit would be invisible to call traversal while occupying its site. Per your note, EFFECTS is required too — Hale has no independent mechanism for proving the effects of an unresolved call, so for every fn-grain kind the two masks are equal; set-level endpoint bits stay independently meaningful with only non-emptiness required.

**Exactly one event per site.** The call-site partition counts hole occupants instead of collapsing to a boolean, and the publish partition rejects duplicate computed-subject holes — your IndirectCall + UntypedReceiver pair at one site is pinned as rejected in both spaces (`one_hole_per_authored_site`). No judgment ever picks a diagnostic by canonical kind order again.

**Contracted rows agree with the interior.** Per-fn set equality between `ViaStdlib` targets and absorption re-emergences, both directions, validated — your rows-without-interior model is refused, as is a re-emergence the contracted relation denies (`via_stdlib_rows_agree_with_absorption`, on a fixture whose `Gate::probe → Hello::handle` contraction is real). The whole corpus satisfies the equality, so validation rather than a judgment-side fail-safe is sufficient: absorption presence is now guaranteed wherever a contracted row exists.

PR #484, opened 2026-08-17
feat(judgment): the boundary-grant family on the canonical model — Change 5b (GH #476)

rileyr:
Change 5b of the canonical-model epic (#476), stacked on #483 (5a): the `only edges` family migrated onto `ClaimIr` × `ApplicationModel` with the same bar — diagnostics parity, the shared corpus differential, negative controls. The old evaluator stays authoritative until Change 9.

## What ships

- **`judgment::judge_only_edges`** — direct-edge law, no walk. Call edges from the source group into the destination group are never grantable (violation pair: claim line + crossing-call span). Bus edges match grants by subject text, with delivery joined on the subscription's **canonical spelling** — topic name for declared subscriptions, authored pattern for literal/wildcard ones, exact-or-wildcard-covering — exactly the graph keying the evaluator's `subscribers_of` iterates. Fail-closed holes (this family's spellings carry no "reachable from"), projection vacuity, unknown-group validation, per-edge dedup, granted-list rendering, and the `FnKey` iteration order (free fns first, then methods by locus) are all ported byte-exact.
- **A latent 5a fix the differential surfaced**: the reachability fan-out also now joins on canonical spelling with wildcard covering — a `subscribe "log.**"` delivery appears in `forbid reaches` witnesses, as the evaluator has always seen it.

## The gate

The 5a corpus differential now judges **both families** and merges rows by ordinal before comparing against the evaluator's outcomes and diagnostic stream — one permanent gate per migration wave, extended per family as the epic prescribes. Two divergences were surfaced and retired on the way: wildcard subscription coverage and the wire-subject-vs-canonical-spelling join.

Negative control: clearing `relations.subscribes` flips a Violated boundary claim to Holds.

1261/1261 across the three crates; fmt clean; no user-visible behavior change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,21 @@ behavior.
 
 ## Unreleased
 
+### The boundary-grant judgment - family 5b (GH #476 Change 5b)
+
+`judgment::judge_only_edges` - the `only edges` family on the
+canonical model, direct-edge law with no walk: call edges into the
+destination group are never grantable; bus edges match grants by
+the subscription's canonical spelling (topic name for declared
+subscriptions, authored pattern for literal/wildcard ones - the
+same graph keying `subscribers_of` iterates, wildcards included).
+Fail-closed holes, projection vacuity, unknown-group validation,
+and the full three-diagnostic violation rendering (claim line,
+un-granted publish site, delivered-at subscription) are ported
+with the evaluator's exact spellings, held by the same corpus
+differential as 5a. Negative control: dropping subscribe rows
+flips the verdict.
+
 ### The reachability judgment - family 5a on the canonical model (GH #476 Change 5a)
 
 `hale_types::judgment::judge_forbid_reaches(&ClaimIrTable,
--- a/crates/hale-types/src/judgment.rs
+++ b/crates/hale-types/src/judgment.rs
@@ -1657,11 +1657,13 @@ pub fn judge_only_edges(
         }
         out
     };
-    // Holes trigger by relation family; the kind picks wording.
-    // `authored_site` (first occurrence) lets the boundary check
-    // interleave holes with known edges as authored — the evaluator
-    // walks fs.calls in source order and refuses at the hole's
-    // authored position, BEFORE any later crossing is reported.
+    // Holes retain their hides-mask; each event space consults the
+    // family IT walks (call space → CALLS, publish space →
+    // PUBLISHES) and the kind picks wording only. `authored_site`
+    // (first occurrence) lets the boundary check interleave holes
+    // with known edges as authored — the evaluator walks bodies in
+    // source order and refuses at the hole's authored position,
+    // BEFORE any later crossing is reported.
     #[derive(Clone)]
     enum FnHole {
         Indirect,
@@ -1669,8 +1671,10 @@ pub fn judge_only_edges(
         Computed,
         Other { reason: String },
     }
-    let mut holes_of: BTreeMap<FunctionId, Vec<(Option<u32>, FnHole)>> =
-        BTreeMap::new();
+    let mut holes_of: BTreeMap<
+        FunctionId,
+        Vec<(hale_model::RelationSet, Option<u32>, FnHole)>,
+    > = BTreeMap::new();
     for h in &model.holes {
         let EntityRef::Function(f) = h.at else { continue };
         let families = hale_model::RelationSet::CALLS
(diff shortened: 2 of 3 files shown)


rileyr:
Fixed at 8a61f54 (rebased onto the corrected #483 foundations, which it inherits).

**Ordered event stream.** `judge_only_edges` no longer partitions calls and holes into separate passes: each fn's events assemble into ONE stream — resolved calls and call-space holes interleaved by authored call-site ordinal (`Hole.authored_site`, first-occurrence ordinal in call space), kind-tagged tie-break. A hole authored BEFORE a violating call now refuses first, exactly as the evaluator's walk encounters it; a hole after it no longer masks the earlier concrete violation. The 5a hides-mask rule applies to which holes enter the stream; the site ordinal decides where.

Byte-diagnostic + verdict parity green over the corpus, including fixtures where a hole and a violating edge share a fn.

rileyr:
Round 2 fixed at b94a3fa (rebased onto the corrected #483, whose publish-ordinal fix this consumes).

**The publish space is one ordered event stream.** With every publish site consuming an ordinal, `judge_only_edges` now merges known publish rows and PUBLISHES-hiding holes into a single stream sorted by authored site, mirroring the call space: a known ungranted publish authored before the computed one reports its violation diagnostics first and THEN the walk refuses; a computed publish authored first refuses without reporting the later crossing. Both event spaces also select holes by their retained hides-mask now.

Pins: `computed_publish_before_known_refuses_first` and `known_violation_before_computed_publish_reports_then_refuses` — both fixtures self-enroll in the corpus differential, which confirmed the evaluator's diagnostic order byte-for-byte in both directions.

rileyr:
Round 3 fixed at 0c1d37e (rebased onto the corrected #483).

**Boundary composition consults subject-grain subscriber holes.** The check now indexes holes anchored at `Subject(S)`; a publish row whose subject hides SUBSCRIBES refuses (Uncertified) at its authored position in the publish stream — the known subscriber rows are not the whole story, so exhausting them no longer certifies the absence of an ungranted edge. Pin: `subject_subscribes_hole_fails_only_edges_closed` (your counterexample: publish, no subscriber row, subject hole ⇒ Uncertified). The 5a frontier-truncation and authored-order fixes arrive through the rebase.

rileyr:
Round 4 fixed at 8558257 (rebased): the publish stream's subject-hole check uses the shared `subject_hole_covers` predicate — exact-or-wildcard over both the topic-name and wire-subject spellings — instead of an exact `SubjectId` lookup. Pin: `wildcard_subscriber_hole_fails_only_edges_closed` (`audit.**` hole refuses a boundary walk publishing `audit.event`).

rileyr:
Round 5 fixed at 8ba5673 (rebased): `judge_only_edges` consumes the shared `BusHoles` index — topic-grain holes now reach the boundary check identically to subject-grain ones. Pin: `topic_grain_subscriber_hole_fails_only_edges_closed`.

rileyr:
Round 6 fixed at 361e35a (rebased): the publish stream's subscriber-hole check consumes the typed `BusHoles` lookup and DEFERS instead of breaking — the known subscriber rows of every publish are still checked, so a known ungranted crossing stays Violated and an earlier call-edge violation can no longer be downgraded; the flag turns only a would-be Holds into Uncertified after the walk. Pins: `known_boundary_violation_survives_subscriber_hole`, `topic_hole_does_not_block_literal_publish_in_only_edges`.

rileyr:
Round 7 at 55acca8 (rebased): inherits the closed hole-shape matrix and validated absorbed-publish identity from #483; this branch's set-level pins now use the matrix-lawful `DynamicEndpoint` kind. No boundary-check changes were needed — its hole consumption already goes through the shared typed `BusHoles` index the matrix now guards.

rileyr:
Round 8 fixed at a2c74e8 (rebased): the boundary poisons on publisher incompleteness — a set-level PUBLISHES hole means an unknown publisher (possibly a src-group fn) may create an ungranted crossing, so a would-be Holds judges Uncertified while a known counterexample still wins. Pin: `publisher_hole_fails_only_edges_closed`. The site-partition, matrix, and capability laws arrive through the rebase.

rileyr:
Round 9 at ca0dd76 (rebased): inherits the required-mask matrix, the one-event-per-site partition, and the ViaStdlib–absorption agreement from #483. No boundary-side changes — the round-9 findings were all model-layer, and this engine's hole consumption already selects by the now-required family bits.

PR #485, opened 2026-08-17
feat(judgment): the endpoint/coverage/count family on the canonical model — Change 5c (GH #476)

rileyr:
Change 5c of the canonical-model epic (#476), stacked on #484 (5b): `require publishes/subscribes`, `require sealed`, `require attributed`, `cover`, and `count` migrated onto `ClaimIr` × `ApplicationModel` — same bar, same shared corpus differential, negative controls. Old evaluators stay authoritative until Change 9.

## What ships

- **`judgment::judge_endpoints`** — all five forms with the evaluator's exact spellings: declared-end existence joined by canonical spelling at the locus-decl grain; the sealed universal with its vacuity refusal; the attribution law with its opaque-call Uncertified fallback; seed coverage over `declared_in`; distinct-loci counting with the who-list rendering. Validation ports include unknown-topic did-you-mean, unresolvable import paths, non-attributable classes (validation and evaluation share the predicate, as the evaluator insists), and topicless seeds.
- **Two new model facts, computed by extracted evaluator rules**: `Function.attribution` (the `performs_directly` predicate per built-in class, moved to `claims::performs_directly_for` and *called* by the builder) and `Function.opaque_call` (`has_opaque_unresolved`) — the round-9 doctrine again: when the evaluator owns a rule, extract and call it.
- **Stream completion**: the differential now prepends the lowering's `issues` (adoption-collision diagnostics from the shared clause enumeration, preserved since round 15) to the expected diagnostic stream — the evaluator emits them before any validation, and with 5c families in scope they became observable.

## The gate

The one corpus differential now covers **seven claim forms across three families**, merged by ordinal. One divergence class surfaced and retired this round (the issues-stream position). Negative control: clearing `relations.declares_publish` flips a holding `require publishes` to Violated.

1262/1262 across the three crates; fmt clean; no user-visible behavior change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,24 @@ behavior.
 
 ## Unreleased
 
+### The endpoint/coverage/count judgment - family 5c (GH #476 Change 5c)
+
+`judgment::judge_endpoints` - `require publishes/subscribes`
+(declared-end existence over `declares_publish` and the
+subscription rows, joined by canonical spelling), `require sealed`
+(universal over group loci with the empty-projection vacuity
+refusal), `require attributed` (the direct-site predicate computed
+by the evaluator's own extracted rule into new model facts
+`Function.attribution` + `Function.opaque_call`; the opaque-call
+Uncertified fallback included), `cover` (seed topic domain via
+`declared_in`), and `count` (distinct declared-end loci with the
+full who-list rendering). Validation ports: unknown topics with
+did-you-mean, unresolvable import paths, non-attributable classes,
+topicless seeds. Adoption-collision diagnostics from the clause
+enumeration surface as lowering issues at the head of the
+diagnostic stream, exactly where the evaluator emits them. Same
+corpus differential; negative control drops declares_publish.
+
 ### The boundary-grant judgment - family 5b (GH #476 Change 5b)
 
 `judgment::judge_only_edges` - the `only edges` family on the
--- a/crates/hale-model/src/entity.rs
+++ b/crates/hale-model/src/entity.rs
@@ -49,6 +49,13 @@ pub struct Function {
     /// classified leaf would match, making the sink set nearly
     /// vacuous (GH #476 Change 5a).
     pub direct_effects: Vec<String>,
+    /// Built-in classes for which this fn PERFORMS a direct site
+    /// (`require attributed`'s predicate, computed by the
+    /// evaluator's own rule). Sorted.
+    pub attribution: Vec<String>,
+    /// Has an unresolved call that is not a frontier path — the
+    /// `require attributed` opaque fallback (GH #476 Change 5c).
+    pub opaque_call: bool,
     pub provenance: ProvenanceId,
 }
 
--- a/crates/hale-model/src/entity.rs
+++ b/crates/hale-model/src/entity.rs
@@ -56,6 +56,12 @@ pub struct Function {
     /// Has an unresolved call that is not a frontier path — the
     /// `require attributed` opaque fallback (GH #476 Change 5c).
     pub opaque_call: bool,
+    /// The author wrote `@effects(is: { <user class> })` on this fn
+    /// — AUTHORED classification, distinct from the expanded label
+    /// set (a composed class labels its atomic members, but the
+    /// purpose the author supplied is the composed NAME; review:
+    /// composed attribution classes).
+    pub carries_user_class: bool,
     pub provenance: ProvenanceId,
 }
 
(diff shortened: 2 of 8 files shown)


rileyr:
Fixed at 5648472 (rebased; inherits the #483 typed-hole and universe corrections).

**Composed classes count as authored attribution.** The model now carries `Function.carries_user_class` as a computed FACT (the builder derives it through `EffectClassDefinition::Composed` expansion), and `judge_endpoints` reads that instead of re-deriving attribution from atomic labels only. A fn annotated `@effects(is: { composed })` where `effect composed = { alloc, … }` is attributed for the member classes. Pinned by `composed_class_counts_as_authored_attribution` — the fixture's verdict is Violated for an unrelated root, and the pin asserts the composed-annotated fn is absent from the unattributed list (the previously-wrong half).

Corpus differential green.

rileyr:
Round 2 fixed at ae2870e (rebased).

**`require attributed` consults holes.** The fallback now unions `Function.opaque_call` with every fn carrying a hole whose hides-mask intersects EFFECTS (and lacking `carries_user_class`) — module-scoped fns and `on_failure` hooks arrive exactly that way, and the result is Uncertified, never a fail-open Holds over a body that may perform the class with no authored purpose.

This is a documented divergence from the evaluator (which never sees unanalyzed bodies); the corpus differential carves out precisely that shape — `RequireAttributed` + old Holds + new Uncertified + an EFFECTS-hiding hole in the model — so any other divergence still trips it. Pin: `attributed_fails_closed_on_unanalyzed_bodies`, using the real `on_failure` recipe end-to-end plus the Holds control without it.

rileyr:
No round-3 items of its own — rebased onto the corrected #484 at dac1f35; suite green (954/954 in hale-types at this head).

rileyr:
Round 4 fixed at 726b2c8 (rebased).

**Relation-aware endpoint holes.** `judge_endpoints` now indexes PUBLISHES / SUBSCRIBES / CARDINALITY holes at subject grain (with the shared wildcard-covering predicate) and topic grain, and treats known endpoint rows as a LOWER BOUND whenever a relevant hole covers the topic — with the monotone cases preserved exactly as you specified:

- `require publishes/subscribes`: a known witness still proves the existential; an absent witness plus a relevant hole judges Uncertified, never Violated.
- `cover`: an apparently-uncovered topic with an incomplete subscriber set cannot prove the violation (Uncertified when nothing is concretely uncovered), while a concretely uncovered topic still proves it.
- `count`: `>=` still holds from enough known rows, and a lower bound already above an upper bound still violates; every other comparison over an incomplete set judges Uncertified. Your example — `count subscribers(topic T) <= 0` with a hole at T's subject hiding `SUBSCRIBES | CARDINALITY` — is the pin's first assertion.

Pins: `count_and_require_fail_closed_on_endpoint_holes` (including the monotone controls: `>= 1` with one known publisher stays Holds even when the publisher set is also marked incomplete) and `cover_fails_closed_on_subscriber_holes`.

rileyr:
Round 5 at f01df91 (rebased): `endpoint_unknown` now delegates to the shared two-grain `BusHoles` index rather than its own parallel collection — one relation-hole lookup across the train, so the same canonical hole has the same reach in every judgment. Behavior is unchanged (covered by the round-4 pins); this closes the consistency half of the round-5 topic-grain finding from the #485 side.

rileyr:
Round 6 fixed at c933584 (rebased): `endpoint_unknown` projects through typed identity — the claim's `TopicId` matches topic-grain holes, and its WIRE pattern (never its name) is what subject-grain holes cover. Pin: `literal_collision_subject_hole_does_not_block_topic_count` — your `Orders` / `"wire.orders"` counterexample, with a real-wire control asserting the topic's actual wire subject still blocks.

rileyr:
Round 7 at f3ec76a (rebased): inherits the closed matrix from #483 — the locus-grain endpoint hole from your counterexample is now rejected at validation, so `require`/`count` can no longer meet a valid model carrying holes they ignore; the shapes they DO consume (subject/topic-grain `DynamicEndpoint` with PUBLISHES/SUBSCRIBES/CARDINALITY/DELIVERY) are the matrix rows. This branch's set-level pins switched to the lawful kind.

rileyr:
Round 8 at 42fe2cb (rebased): inherits the tightened matrix (fn-grain SUBSCRIBES shape rejected — your `require`/`count` counterexample can no longer inhabit a valid model), the site partition, and the absorption-aware capability law. No endpoint-side changes were needed: these judgments already consumed set-level PUBLISHES through `endpoint_unknown`.

rileyr:
Round 9 at 0b9c81e (rebased): inherits the round from #483. No endpoint-side changes needed.

PR #486, opened 2026-08-17
feat(judgment): the quantitative-bound family on the canonical model — Change 5d (GH #476)

rileyr:
Change 5d of the canonical-model epic (#476), stacked on #485 (5c): the `bound` family migrated onto `ClaimIr` × `ApplicationModel` — the evaluator's `site_count` DFS ported over both vertex kinds, corpus-parity green on the first differential run. Old evaluator authoritative until Change 9.

## What ships

- **`judgment::judge_bound`** — call-tree SUM of carrier sites with MAX over dispatch alternatives (model call rows sharing a site ordinal *are* the dispatch group, which is what makes the fold structural), memoization with the evaluator's cycle discipline, the full `Unbounded` vocabulary with its four-way `why()` renderings, the site-pointing related diagnostic under the bundle-span gate (interior stdlib sites carry no bundle span, exactly as the evaluator suppresses them), and the heaviest-path witness chain.
- **Count-grade model facts**: `AbsorbedNode.carries` (rendered carrier classes — `std::secret` bodies carry `secret_use` and must count), `in_loop` + dispatch `group` on interior call events, `in_loop` on interior publishes, and `Publish.in_loop` on the model's own send rows. The carrier test intersects labels (user fns) or node carries (interior) with the class's atomic expansion from the typed effect-class table.
- **Validation ports**: unknown group, the builtin rejection (`bound` takes a user class or `secret_use` — one contract, one spelling), undeclared classes with did-you-mean, projection vacuity.

## The gate

The shared corpus differential now judges **five families** (reachability, boundary grants, endpoint/coverage/count, bounds) merged by ordinal — and 5d went green on its first run, which is the count-grade absorption facts paying for themselves. Negative control: clearing `model.labels` zeroes the count and flips a Violated bound to Holds.

1263/1263 across the three crates; fmt clean; no user-visible behavior change.

Next: 5e (pointwise certificates — the `@effects`/`@no_panic`/`@phase_effects`/`@budget` annotation surfaces), the last family.


rileyr:
Fixed at 83d1ff9 + aee6bb5 (rebased; inherits the earlier corrections).

**Entry facts survive the user→stdlib edge.** The absorption entry now records `entry_dispatch`, `entry_in_loop`, `entry_group`, and `entry_provenance`; `judge_bound` builds the crossing edge from those, so a looped call into a carrier-bearing stdlib body is Unbounded with the loop attributed at the authored site. Pinned by `looped_stdlib_entry_with_carrier_is_unbounded` (hand-built model).

**True interleave, both spaces.** 83d1ff9's first cut still bucketed holes ahead of known calls — aee6bb5 is the honest fix: call space merges direct calls, absorbed entries, and unfollowable holes by site ordinal; publish space merges publishes and computed-subject holes by publish-site ordinal; `Ev::Truncated` lowers to `UnboundedIr::StepCeiling`. The walk consumes each fn's events in authored order, so first-refusal semantics match the evaluator exactly.

Corpus differential green (1270/1270 at the rebase point).

rileyr:
Round 2 fixed at 6821871 (rebased; consumes the #483 publish-ordinal fix for its event streams).

**Shared projection + FnKey root order.** `judge_bound` now admits member free fns unconditionally (the evaluator's `fn_set` inserts every named free fn — a declaration-only fn evaluates with contribution zero rather than making the group projection-vacuous) and iterates roots in FnKey order (free fns before methods), so first-unbounded-root selection matches the evaluator's reason and witness. Pin: `declaration_only_free_fn_bound_counts_zero` (self-enrolls in the differential — evaluator agrees).

**One authored dispatch = one group.** Every alternative introduced by one dispatch site now folds under the normalized local group `Some(site)`, whether its target is a user fn or an absorption entry — `entry_group` (summary-global) no longer splits one runtime dispatch into two groups whose contributions summed. `entry_in_loop` / `entry_provenance` still come from the real entry edge. Pin: `mixed_dispatch_alternatives_share_one_group` (hand-built model where max(1,1)=1 Holds and the old sum reported a phantom 2).

**Cyclic classes refuse.** `bound` now applies the shared InvalidCycle validation — `Invalid` before evaluation with the 5a wording, never Holds by counting zero through an empty atom set. Pin: `cyclic_bound_class_is_invalid`.

rileyr:
Round 3 fixed at a52a48b (rebased).

**Carrier counting consults EFFECTS.** A fn whose EFFECTS rows are hidden has an unknown own-count — the walk refuses at that vertex instead of counting zero and certifying. **Bus traversal consults subject-grain SUBSCRIBES:** a publish whose subject's subscriber set is incomplete makes the fan-out's contribution unknown. Both are UNKNOWN shapes rather than "unbounded", so they judge **Uncertified** with their own wording (there is no evaluator analog — the old engine cannot see these model facts), distinct from the Violated that Cycle/LoopCarrier/Unfollowable/StepCeiling produce.

One subtlety the pin caught: publish rows for declared topics carry the topic NAME, not the wire pattern, so the subject-hides text index covers both spellings (the same fix applies to the 5a index it mirrors).

Pins: `effects_hole_makes_bound_uncertified` (your second counterexample verbatim — EFFECTS-only hole on a root ⇒ Uncertified, not zero-count Holds) and `subject_subscribes_hole_makes_bound_uncertified`.

rileyr:
Round 4 fixed at 53564a0 (rebased): the count walk's subject-hole check uses the shared `subject_hole_covers` predicate over both the topic-name and wire-subject spellings, replacing the exact text lookup. Pin: `wildcard_subscriber_hole_makes_bound_uncertified` (`audit.**` hole makes fan-out through `audit.event` unknown ⇒ Uncertified).

rileyr:
Round 5 fixed at bc4105d (rebased).

**The known lower bound decides first.** Unknowns left the Err channel entirely: an EFFECTS-holed vertex still counts its known labels, and a subscriber-holed publish still traverses its known handlers — both set a side-channel `UnknownCause` instead of aborting. The verdict now follows exactly the monotonic rule the endpoint counts adopted in round 4:

- known count > limit → Violated (your example — `F` labeled `money`, EFFECTS hole, `bound money <= 0` — is the pin: still Violated with the hole present)
- known count ≤ limit and unknown exists → Uncertified
- otherwise → Holds

Pins: `known_violation_survives_effects_hole`, `known_fanout_violation_survives_subscriber_hole` (the fan-out equivalent: a known subscriber path over the limit stays Violated under the subject hole), with the round-3 Uncertified pins still green for the known-under-limit case.

**Topic-grain holes** reach the count walk through the shared `BusHoles` index. Pin: `topic_grain_subscriber_hole_makes_bound_uncertified`.

rileyr:
Round 6 fixed at e19a3a6 (rebased).

**Typed publish identity.** The count walk's publish events carry the declared `TopicId` and the wire pattern (user rows and stdlib interiors both, via the shared `AbsorbedEvent::Publish.declared_topic`), and the subscriber-hole check consumes the typed `BusHoles` lookup.

**The invariant behind the fold.** `ApplicationModel::validate` now enforces the `(from, site)` equivalence class: more than one member — counting non-ViaStdlib call rows and absorption entries together — requires interface alternatives (call rows of ONE interface; absorption entries with a dispatch rendering). Two direct calls at one site are rejected, so bound's per-site MAX can never absorb what the evaluator would SUM. ViaStdlib rows are excluded as the legacy contraction of the same authored call an absorption entry represents. Pin: `two_direct_calls_cannot_share_a_site` (asserts the builder's output is lawful, then forces the two-direct-one-site shape and requires refusal). I went with validation rather than an explicit dispatch-group field to keep the schema churn contained; happy to add the field in a follow-up if you'd prefer the fact to be represented rather than inferred.

rileyr:
Round 7 fixed at e6d1833 (rebased).

**Interior dispatch identity is validated.** Absorbed calls now satisfy `group.is_some() ⇔ dispatch.is_some()`, and within one node the same group id carries exactly ONE (interface, method) rendering — your two dispatch-less group-7 calls are refused, as is one group spanning two different dispatches, while genuine conformer alternatives stay lawful (`interior_dispatch_groups_are_validated`).

**Entry compatibility is identity agreement, not presence.** The `(from, site)` class check now resolves each user interface row's raw interface through the model's own `interfaces` table to its display, pairs it with the conformer's method name, and requires every absorption entry's dispatch rendering to equal that same pair (and each other) — the raw/display spelling divide is bridged by the table rather than guessed, so a stdlib entry can no longer ride an unrelated interface call at its site.

rileyr:
Round 8 fixed at 5082547 (rebased).

**Dispatch labels bind to their targets** (landed on #483 where the schema lives, composing with this branch's round-7 group laws): an absorbed call's rendered method must be its target's own method identity — User target: the function's name tail; Interior target: the node display's tail — and the entry label must be node zero's method. Your charge_a/charge_b-under-("Payer","pay") example is refused at validation, so `judge_bound`'s per-group MAX can never collect calls to unrelated functions under one label. The round-7 pin's lawful control now uses conformers whose displays spell the dispatched method, exercising both laws together.

**Bound poisons on publisher incompleteness**: a set-level PUBLISHES hole seeds the walk's `UnknownCause` — fan-out may be missing edges — so a within-limit count judges Uncertified while a known over-limit lower bound still violates. Pin: `publisher_hole_makes_bound_uncertified`.

On the schema note: I kept `AbsorbedNode.display`-tail as the canonical method identity rather than adding a typed field — the display is the demangled FnKey, whose tail IS the method by construction, and the validator now enforces the binding. Happy to lift it into a typed field in a follow-up if you'd prefer the identity stored rather than derived.

rileyr:
Round 9 at 278a65e (rebased): inherits the required-mask matrix, the one-event partition, and the ViaStdlib–absorption agreement — the count walk's absorption-only traversal is now validated-safe, since a contracted row without its realizing interior can no longer inhabit a valid model. One pin literal gained this branch's absorption fields in the rebase.

PR #487, opened 2026-08-17
feat(judgment): the pointwise-certificate family on the canonical model — Change 5e (GH #476)

rileyr:
Change 5e of the canonical-model epic (#476), stacked on #486 (5d): the pointwise certificates — `@effects(none/only/publish)`, `@no_panic`, `@phase_effects` — migrated onto `ClaimIr` × `ApplicationModel`, closing the 5a–5e wave. Old evaluators stay authoritative until Change 9.

## Architecture

The certificate engines are witness-driven analyses over the merged summary (per-class site probes, `witness_path`, trap analysis) — facts the model deliberately does not carry per-site. Rather than approximate them, the migration takes the evidence route the epic's Change 6 formalizes: the engines remain the one analysis authority, the builder runs them via a **grouped report** (`effect_report_grouped` / `effect_report_three_way` — the *same pass* `hale check` consumes; its flat diagnostic stream is reassembled from the strata, so the check path and the evidence cannot disagree), and each certificate's outcome + diagnostics land as typed **`CertificateEvidence`** rows on the model — the artifact's #392 `lowered` rows, now with their diagnostics. The judgment renders verdicts and diags from model data, and adds the undeclared-class validation from the typed effect-class table; undeclared `is:` classes surface as lowering issues (carries has no law row).

- **`Provenance::ForeignSpan`** — the evaluator's leaf spans inside stdlib bodies live in their own offset space; evidence preserves them verbatim so rendering is byte-identical (the differential caught this on its first run).
- **The 5e differential** compares the evaluator's law strata (undeclared-class pass + all per-certificate groups, phase contracts included) as a **multiset**: cross-certificate stream order legitimately differs between `hale check` (phase contracts ride the placement stream) and ClaimIr ordinal order; within-certificate diag order is the evidence's own. Documented in the test.
- Negative control: clearing `model.evidence` flips a holding certificate to Invalid — the judgment provably reads evidence.

## Scope note

`@effects(causes/depends)` evaluate in `check.rs`'s graph pass (no lowered rows exist for them today), and `@budget`/quantitative dims have no artifact rows — both are lowered to `ClaimIr` (since Change 4) and their judgment migration follows in-review as the 5e completion, on this branch.

1265/1265 across the three crates; fmt clean; no user-visible behavior change.

With this PR the 5a–5e chain is complete: **#483 (reachability) ← #484 (boundary grants) ← #485 (endpoint/coverage/count) ← #486 (bounds) ← #487 (certificates)**, one shared corpus differential across all five families.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,28 @@ behavior.
 
 ## Unreleased
 
+### The pointwise-certificate judgment - family 5e (GH #476 Change 5e)
+
+`judgment::judge_certificates` - `@effects(none/only/publish)`,
+`@no_panic`, and `@phase_effects` on the canonical model. The
+certificate engines remain the one analysis authority: the builder
+runs them (the grouped report shares the exact pass `hale check`
+uses, split into strata) and stores each certificate's outcome and
+diagnostics as typed `CertificateEvidence` on the model - the
+artifact's #392 lowered rows, with their diags; the judgment
+renders verdicts and diagnostics from model data, adds the
+undeclared-class validation over the typed effect-class table, and
+the lowering emits undeclared `is:` classes as issues (carries has
+no law row). `Provenance::ForeignSpan` preserves stdlib-space
+diagnostic offsets verbatim. Corpus differential compares the
+evaluator's law strata as a multiset (cross-certificate stream
+order differs by design between `hale check` and ClaimIr ordinal
+order; within-certificate order is the evidence's own). Negative
+control: clearing model.evidence invalidates certificates.
+Remaining 5e surfaces (`@effects(causes/depends)` which live in
+the check.rs graph pass, and `@budget`/quantitative dims) follow
+in-review.
+
 ### The quantitative-bound judgment - family 5d (GH #476 Change 5d)
 
 `judgment::judge_bound` - `bound C <= N on paths from G` on the
--- a/crates/hale-model/src/application.rs
+++ b/crates/hale-model/src/application.rs
@@ -338,21 +338,25 @@ impl EvidenceTable {
             });
         }
         // Source-snapshot tie: the sidecar's source units (path +
-        // content digest) must be exactly the judged model's — a
-        // source edit that merely shifts an offending site changes
-        // neither shape nor law, but its old local offsets would
-        // render against the new source (review round 3).
-        if self.provenance.sources.len()
-            != model.provenance.sources.len()
-            || self
-                .provenance
-                .sources
-                .iter()
-                .zip(model.provenance.sources.iter())
-                .any(|(a, b)| {
-                    a.path != b.path || a.digest != b.digest
-                })
-        {
+        // content digest) must exactly equal BOTH the judged
+        // model's and the law table's (review rounds 3–4) — the
+        // three artifacts must describe one snapshot, or stale
+        // offsets render against the wrong source bases.
+        let unit_mismatch =
+            |a: &[crate::provenance::SourceUnit],
+             b: &[crate::provenance::SourceUnit]| {
+                a.len() != b.len()
+                    || a.iter().zip(b.iter()).any(|(x, y)| {
+                        x.path != y.path || x.digest != y.digest
+                    })
+            };
+        if unit_mismatch(
+            &self.provenance.sources,
+            &model.provenance.sources,
+        ) || unit_mismatch(
+            &self.provenance.sources,
+            &table.provenance.sources,
+        ) {
             return Err(ClaimIrError::InvalidProvenanceRecord {
                 index: usize::MAX,
             });
--- a/crates/hale-model/src/application.rs
+++ b/crates/hale-model/src/application.rs
@@ -15,7 +15,7 @@ use crate::entity::{
 };
 use crate::hole::Hole;
 use crate::ids::{EntityRef, FunctionId, ProvenanceId, TopicId};
-use crate::claim_ir::ClaimIrError;
+use crate::claim_ir::{ClaimIrError, ClaimIrTable};
 use crate::provenance::{Provenance, ProvenanceTable};
 use crate::relation::{
     AffinedTo, Call, DeadInterfaceCall, DeclaredIn, DeclaresPublish,
@@ -298,26 +298,37 @@ pub struct EvidenceTable {
     /// with — a judgment refuses evidence whose shape disagrees
     /// with the model it is asked to judge.
     pub model_shape: u64,
+    /// [`ClaimIrTable::semantic_digest`] of the law table this
+    /// evidence answers. The topology shape does NOT hash
+    /// annotation laws — two programs with identical topology but
+    /// different `@effects` classes share a `model_shape`, so the
+    /// sidecar must also be tied to the LAW it certifies (review
+    /// round 2).
+    pub law_digest: u64,
     pub rows: Vec<EvidenceRow>,
     pub provenance: ProvenanceTable,
 }
 
 impl EvidenceTable {
     /// Structural laws against the judged pair: the shape must
-    /// match the model, ordinals must be unique and in the law
-    /// table's range, subjects must agree with the ClaimIr row's
-    /// resolution, and diagnostic provenance must resolve.
+    /// match the model, the law digest must match the table, and
+    /// ordinals must be unique and in the law table's range,
+    /// subjects must agree with the ClaimIr row's resolution, and
+    /// diagnostic provenance must resolve.
     pub fn validate(
         &self,
         model: &ApplicationModel,
         model_shape: u64,
-        law_rows: usize,
+        table: &ClaimIrTable,
     ) -> Result<(), ClaimIrError> {
-        if self.model_shape != model_shape {
+        if self.model_shape != model_shape
+            || self.law_digest != table.semantic_digest()
+        {
             return Err(ClaimIrError::InvalidProvenanceRecord {
                 index: usize::MAX,
             });
         }
+        let law_rows = table.rows.len();
         let mut seen = std::collections::BTreeSet::new();
         for (i, row) in self.rows.iter().enumerate() {
             if row.ordinal as usize >= law_rows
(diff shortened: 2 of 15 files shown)


rileyr:
All three architecture P1s corrected at f13cbf1 (full rework, workspace suite 2735/2735).

**Evidence is a sidecar, keyed by ordinal.** `ApplicationModel.evidence` is gone — a model no longer carries a cached prior judgment of itself. `hale_model::EvidenceTable` holds evidence rows keyed by ClaimIr ORDINAL with the typed subject (`Option<FunctionId>`), its own provenance table, and `model_shape` = the model's `TopologyShapeV1`. The producer (`evidence::derive_certificate_evidence`) runs the one grouped engine pass and performs the (subject, form) reconstruction exactly once; `judge_certificates(table, model, &evidence, bases)` consumes ordinals only — no string cursor. Stale evidence (shape mismatch) judges `Invalid` with a naming diagnostic; a missing, short, or subject-disagreeing row is `Invalid`. `EvidenceTable::validate` enforces the structural laws (shape tie, unique in-range ordinals, subject agreement, resolvable records). Canaries: `stale_evidence_is_refused`, `dropping_evidence_changes_the_verdict`, `derived_evidence_validates`.

**Engine completeness.** The four variants whose engines live elsewhere (`EffectCauses`, `DependsSet`, `AllocBudget`, `QuantBudget`) now each produce exactly one Judged row — at minimum `Uncertified` — and the corpus differential enforces judged-ordinals == certificate-family-rows per program, plus verdict parity against the evaluator's replayed severities. Pinned by `unmigrated_families_judge_uncertified`.

**Undeclared class ⇒ `Invalid`, one dedup authority.** A row asserting about a never-declared class judges `Invalid` (the evaluator's row held vacuously). The diagnostic moved to the LOWERING as a `ClaimIrTable` issue with the evaluator's per-annotated-root dedup across a root's `none:`/`causes:`/`is:` lists — it must live there because `is:` (carries) lowers to no row at all; this also closed a silent gap where the carries validation was threaded but never written. Pins: `undeclared_class_is_invalid_not_holds`, `undeclared_class_dedups_across_a_roots_lists`.

**ForeignSpan by explicit discriminator.** The stdlib parses at base 0, so its spans numerically overlap the first user file — containment was a guess in both directions. The certificate emitters now flag each diagnostic's origin AT THE EMIT SITE from the witness step's owning `FnKey` (`DiagSink` carries per-diag flags; `find_effect_site` returns the site's owner); the producer maps flagged diags to `ForeignSpan` verbatim (normalized non-inverted, and both model and evidence validation reject inverted records) and only user-flagged diags to `Source`. Pinned self-calibratingly by `stdlib_owned_diag_span_is_foreign_never_source`: the test measures the stdlib offset, pads the user file until that offset falls inside its numeric range, and requires the classification to survive.

Side effect worth noting: phase certificates now emit into the shared diag sink, so their group ranges are recorded directly in final-stream coordinates (the `phase_base` rebasing step is gone); the flat `hale check` stream is byte-identical, corpus-verified.

rileyr:
Round 2 fixed at aba9a4f.

**Evidence ties to the law it answers.** `ClaimIrTable::semantic_digest()` — a crate-owned FNV-1a over every row's `Hash` image (ordinal, name, origin, the full typed law with its references; deliberately not `DefaultHasher`, whose algorithm is unspecified across Rust releases) — is recorded on the sidecar as `law_digest` at derivation. The judgment refuses evidence whose digest disagrees with the table it is asked to judge. Your two-program example is the pin: `same_shape_different_law_evidence_is_refused` asserts the `none: {syscall}` / `none: {alloc}` fixtures SHARE a `TopologyShapeV1` (so the shape check alone cannot distinguish them) and that the first law's evidence judges the second law `Invalid`.

**Mandatory validation in the judgment.** `judge_certificates` now runs `EvidenceTable::validate` itself — shape tie, law-digest tie, unique in-range ordinals, subject agreement, resolvable diagnostic provenance (including inverted spans) — and a malformed sidecar invalidates every certificate row. Duplicate ordinals no longer collapse silently in the `BTreeMap`; dangling provenance no longer renders span zero; and the per-row certificate count is now exact (`== expected`, not `>=`), so extra certificates are refused rather than prefix-consumed. `CertificateEvidence.form` remains display data; agreement is enforced by the digest over the typed law, not by string comparison. Pin: `malformed_evidence_is_refused` (duplicate ordinal, extra certificate, dangling provenance — each flips Holds → Invalid).

**Imported subjects render their author-facing spelling.** The lowering's undeclared-class issue now formats the resolved `NameRef.display` instead of `raw`, so a library annotation names `kv::helper` / `kv::Store::bump` rather than `__lib_…`. Pin: `imported_undeclared_class_issue_renders_display` — a cross-seed bundle covering an imported free fn AND an imported method, plus a no-`__lib_` leak assertion.

Workspace suite 2749/2749 at this head.

rileyr:
Round 3 fixed at f5f3761.

**Per-row law binding.** `ClaimRow::certificate_forms()` (hale-model, dep-free) is now the one authority for the `(subject, form)` strings a row's engines produce — the producer matches against it, and `EvidenceTable::validate` requires each evidence row's stored certificate forms to agree positionally with its exact ClaimIr row. `form` is no longer unchecked machine data. Your swap scenario is the pin: `swapped_certs_between_same_subject_laws_are_refused` — two one-certificate laws on one fn exchange their `certs` vectors; digest, ordinals, subjects, counts, and provenance all survive the swap, and the per-row form binding is what refuses (every certificate row ⇒ Invalid). A SHORT row (an unresolved subject the engines never saw) deliberately remains a per-row judgment concern rather than a table-wide refusal; a longer row or any form disagreement is a malformed sidecar.

**Snapshot ties.** Validation now compares the sidecar's source units (path + content digest) against the judged model's exactly — pinned by `edited_source_evidence_is_refused`, which asserts the edited fixture keeps an identical `TopologyShapeV1` (motion-invariance is by design) and an identical law digest, so the source-digest comparison is demonstrably what refuses the stale offsets. And `EvidenceTable.inputs_digest` covers the certificate engines' inputs outside the model — the Hale-source stdlib the walks absorb plus the compiler version — recomputed by the judging toolchain at validation. Pin: `different_toolchain_evidence_is_refused`.

Workspace suite green at this head (2758/2758).

rileyr:
Round 4 fixed at 29bf524.

**The law table is bound to its snapshot.** `semantic_digest` now hashes each row's provenance id plus the table's full provenance store — source units AND records — and `EvidenceTable::validate` requires exact source-unit agreement among all three artifacts: evidence, model, and law table. Your mixed pairing (model + evidence from snapshot A, table re-lowered from snapshot B with identical semantic annotations but shifted locations) is refused twice over: the law digests differ and the table's source units disagree with the evidence's. Pin: `relowered_table_from_edited_source_is_refused`, which asserts the two tables' typed laws are identical so the provenance store is demonstrably what distinguishes the digests.

**Analysis identity beyond the package version.** `analysis_inputs_digest` gains `ANALYSIS_SEMANTICS_VERSION` — an explicit engine-semantics constant whose documented contract is to be bumped in the same change as any result-affecting modification to the certificate engines (effect/witness traversal, allocation-summary behavior, grouping, strata, wording) — and now hashes the static registries that are data rather than code: the stdlib-surface classification table (namespaces, fn names, effect bits, open prefixes) and the `PATH_RENAMES` table, alongside `AP_SOURCE` and the package version. Registry drift therefore invalidates evidence mechanically, without anyone remembering the constant; engine-code drift is the constant's documented obligation, kept adjacent to the code it governs.

rileyr:
Round 5 fixed at feb2dbf.

**Cyclic-class certificates refuse.** `judge_certificates` applies the shared class-validity rule reachability and bound already follow: every class-bearing certificate form naming an `InvalidCycle` class judges Invalid — Forbid/Only/Causes lists, phase allow-lists, and user-class quant budgets. Your exact two-class cycle is the pin (`cyclic_class_certificate_is_invalid_not_holds`); it self-enrolls in the corpus differential under the documented-divergence guard, since the evaluator's row holds vacuously while the machine verdict now refuses the denotation.

**Unrenderable provenance is rejected.** Both `ApplicationModel::validate` and `ClaimIrTable::validate` now reject an inverted `ForeignSpan` — an accepted-but-unrenderable record is the same defect as a dangling `SourceId`. The `StdlibAbsorption.entry_provenance` range check landed in the #483 commit this branch inherits (`c462e8e`), closing the other half of the finding where the field lives.

Workspace 2770/2770 at this head.

rileyr:
Round 6 fixed at 0eadc3e.

**Budget class validity.** A `QuantBudget` row with a `UserClass` dimension now feeds its class reference into the shared validity rules, giving exactly your three-line matrix: undeclared ⇒ Invalid, cyclic ⇒ Invalid (round 5's check already covered the dimension), valid-but-unmigrated ⇒ Uncertified. The quantitative evaluator's diagnostic — "budgets effect class `monye`, which is never declared", with its near-miss hint — is retained as a lowering issue beside the row. Pin: `undeclared_budget_class_is_invalid_not_uncertified` (Invalid for the typo'd class, an Uncertified control for the valid one, and the issue with its hint).

**`ANALYSIS_SEMANTICS_VERSION` = 2.** Bumped per its own contract: rounds 5–6 changed certificate verdict semantics (cyclic classes Invalid instead of a replayed vacuous Holds; undeclared budget dimensions Invalid instead of Uncertified), so v1-produced evidence is refused by a v2 judgment and vice versa. The version doc now records what v2 means, so the constant's history doubles as a semantics changelog.

Workspace 2777/2777 at this head.

rileyr:
Round 7 at the rebased head: inherits the full validation hardening from #483/#486 (closed hole matrix, absorbed-publish identity, interior dispatch identity). No certificate-side changes this round — the three blockers were all validation, no verdict semantics moved, so `ANALYSIS_SEMANTICS_VERSION` correctly stays at 2. Workspace 2780/2780 at this head.

rileyr:
Round 8 at the rebased head (workspace 2786/2786): inherits the full round — consumed-only hole shapes, the authored-site partition, target-bound dispatch labels, and absorption-aware capabilities. No certificate-side changes: validation-only round, no verdict semantics moved, `ANALYSIS_SEMANTICS_VERSION` stays 2.

rileyr:
Round 9 at the rebased head (workspace 2788/2788): inherits the model-layer round. No certificate-side changes; validation-only, `ANALYSIS_SEMANTICS_VERSION` stays 2.

PR #489, opened 2026-08-18
feat(artifact): typed law rows + adequacy — artifact generation projects the model (GH #476 Change 6)

rileyr:
GH #476 **Change 6**: the artifact's law rows are projected from the canonical model path, and the artifact gains typed `law` / `capabilities` / `adequacy` sections.

## What changed

**One render authority.** `ClaimRow::claims_form()` (hale-model, dep-free) reproduces the evaluator's form spelling for every claims-block `ClaimIr` variant; `ClaimRow::family()` names each row's judgment family, with `JudgmentFamily::required_relations()` as the adequacy contract — the same authority pattern `certificate_forms()` established in Change 5e.

**Artifact generation projects the model** (acceptance criterion 6, for the migrated families). `project_law_rows()` renders `claims` rows from ClaimIr with verdicts merged from the five Change-5 judgments, and the effects-family `lowered` rows from the evidence sidecar, preserving the legacy report's phase-first stream order. `@budget` rows keep their old producers until the quantitative engines migrate (`JudgmentFamily::Unmigrated`). The old evaluators remain the **check** authority until Change 9.

**The corpus differential** (`artifact_law_projection.rs`) holds the projection equal to the evaluator rows — name, form, result, source, and ORDER — over the whole corpus, with the two Change-5 documented divergences carved out explicitly. It passed on first contact: the form renderer and stream ordering reproduce the evaluator byte-for-byte.

**Schema 1.11** — three unhashed, digest-covered typed sections:
- `law`: every lowered ClaimIr row (ordinal, name, origin, family, machine verdict, source provenance) + `law_digest` + `inputs_digest` — the evidence-sidecar ties a consumer checks before trusting external evidence against this artifact;
- `capabilities`: the model's positive completeness account, typed;
- `adequacy`: per migrated family, `exact` when capabilities vouch every relation family that judgment consumes, else `degraded`.

`shape_hash` is **unchanged** — replay admission and recorded baselines survive (the whole-corpus projection differential still holds).

**MODEL_SEMANTICS = 2.** Machine verdicts are stricter in the two documented places (cyclic/undeclared-class certificates ⇒ `invalid`, never a vacuous `holds`; `require attributed` over an unanalyzable body ⇒ `uncertified`, never a fail-open `holds`), and the document `verdict` follows the machine. The three-level story is pinned: the legacy `lowered` row preserves the engine's replay, the `law` row carries the machine's word, and `verdict` follows the law. Consumers track the constant and refuse unrecognized semantics by design.

**Demand boundary**: artifact emission now legitimately derives the model. The demand canary's plain-check half is byte-unchanged (a no-claims check still never derives); its artifact half now asserts the demand exists.

## Tests

- `artifact_law_projection.rs`: the corpus differential + typed-section pins (`law` digests recompute against the canonical path; adequacy degrades per family — a computed publish degrades every bus-composing family while `certificate` stays exact; the semantics-2 end-to-end pin).
- Tamper/golden tests now track the schema/semantics constants instead of hardcoding versions.
- Spec (`spec/verification.md`) + book (`docs/src/claims.md`) updated in the same change; CHANGELOG entry included.
- Workspace 2792/2792.


rileyr:
Round 1 fixed at 2404307.

**Production projects the model.** `dump_topology` now emits `project_model_half(&vmodel)` as the hashed model half, and `shape_hash` is its hash (`debug_assert`ed equal to `project_shape_hash`) — byte-identical over the corpus, so existing hashes and replay admission are untouched. The legacy gathering no longer supplies any emitted row: it is returned only through `dump_topology_parts(..).1` for the corpus differential, whose direction is now inverted — the projection is production, the legacy arm is the comparison, retired at Change 9. One semantic authority; acceptance criterion 6 satisfied as stated.

**Typed law payload + evidence.** `ClaimRow::law_payload_json()` (hale-model, dep-free) renders one tagged object per `ClaimIr` variant with the law's operands — every reference as `{"name": raw, "display", "resolved"}`. Your two-class example now reads: `"law": {"kind": "effect_forbid", "at": {...}, "classes": [{syscall}, {publish}]}` with `"certs": [{"ordinal": 0, "form": …syscall…, "result": "holds"}, {"ordinal": 1, "form": …publish…, "result": "violated"}]` — which certificate failed is machine-addressable (pinned). The raw half of each reference is the machine join key; the cross-seed author-spelling test exempts `law` exactly as it exempts `topics`, with the duality documented. And **Track A's claim view now consumes the payload**: highlighting walks the typed operand tree (`law_display_names`) instead of substring-matching the form string — the "Track B upgrade" comment is gone because the upgrade is here.

**Trustworthy adequacy.** The certificate family requires PUBLISHES (`@effects(publish: {…})` cannot prove a computed subject in-set), and the builder now derives `exact_cardinality` from the closed-world enumeration — endpoint adequacy is no longer permanently degraded. Both of your controls are pinned: a computed-subject program degrades every family including `certificate`, and a fully known program with a `count` claim reads `exact` across the board with the count row `holds`.

**No unwitnessed passes.** `legacy_unmigrated_verdicts` carries the old engines' authoritative results into the unmigrated law rows: `@budget` aligns by position within each variant (counts asserted; a mismatch fails closed to `uncertified`), and `causes:`/`depends:` attribute the old passes' diagnostics by the exact span both sides anchor at (the fn-name span / the depends-declaration span — verified identical sources). `law_pass` then covers **every** application-tier row with no family exemption. Pinned two ways: `unmigrated_rows_carry_legacy_verdicts` (a passing `causes:` and `@budget` read `holds`; a violated budget reads `violated` and fails the document) and the corpus-wide invariant `clean_verdict_implies_every_law_row_holds`.

Spec, book, and CHANGELOG updated. Workspace 2795/2795; both corpus differentials green.

rileyr:
Round 2 fixed at 41098df.

**No legacy production authorities.** The unhashed `sources` / `provenance` / `topics` sections now project from `ApplicationModel` (`project_unhashed_tail`), byte-green over the whole corpus; the legacy gathering feeds only the differential, which was extended to compare the tail with the same inverted direction. Getting to byte parity surfaced three real model gaps, each fixed at the schema:
- unplaceable spans intern as `ForeignSpan` (offsets preserved — `Synthetic` dropped them, so sourceless bundles lost every provenance span);
- `SourceUnit.digest` keeps the producer's exact string (the numeric parse lost non-canonical digests and made `sources` unprojectable);
- the topics projection renders the observation identity with the subject-fused `topic_shape_hash`, and opaque cross-seed payloads keep the empty-shape spelling this section has always carried (the artifact/manifest fusion rule for those is a pre-existing question I've deliberately not changed here);
- `provenance.calls` applies the same v1-both-endpoints universe filter as the hashed half (a declaration-universe-recovered edge has no legacy span row).

**The bridge never manufactures `holds`.** `legacy_unmigrated_verdicts` bridges only rows the old engines demonstrably enumerated: an old-visible (nonrecursive) annotation walk gates budgets positionally per variant, and causes/depends attribution additionally requires a single assert per subject — two asserts share one anchor span, so a diagnostic cannot be attributed and those rows stay `uncertified`. Your module-scoped `causes:` example is the pin: lowered, never evaluated, `uncertified`, and the document verdict is `law_failed` — an unwitnessed law cannot leave the artifact clean.

**Typed selectors.** `law_payload_json` now takes the entities and provenance tables it renders from; `BusSelector` payloads serialize their resolved candidate sets — topic identities as raw+display pairs, wire-subject patterns — plus the selector's own source location. Pinned with a publish-set fixture asserting the candidate topic and the span.

**Relation-level adequacy.** `family_adequacy` derives from the new `ApplicationModel::unresolved_relation_mask()` (typed holes ∪ absorption residue — the exact mask the capability law reads), so publish and subscribe completeness count independently: your subscriber-only-hole case is the pin — `certificate` stays `exact` while the bus-composing families degrade. The coarse capability flags remain the public completeness account; adequacy no longer routes through their coupling.

**1.11 admission.** Track A requires and structurally validates `law` (digests, row shapes incl. known families/verdicts, tagged payloads), `capabilities`, `adequacy`, and the claims-to-law join; the tamper suite gained the restamped-deleted-`law` case, refused with the section named.

Workspace 2798/2798; both corpus differentials byte-green.

rileyr:
Round 3 fixed at 5387e53.

**Structural opacity.** `PayloadContract` gains `opaque: bool`, set at the one decision point (whether `canonical_type_shape` resolved) and carried through the payload dedup keys — so two contracts whose strings collide but whose opacity differs stay distinct rows. The projection reads the flag, never the string. Your fixture is the pin: a struct field literally named `opaque` yields `"shape": "opaque:i"` in the topics section with the fused hash byte-equal to the shared `topic_shape_hash` implementation the binary registers with — the artifact/manifest/recording join survives. (I used a struct discriminant rather than the enum sketch to keep the shape string's ~30 value-agnostic readers intact; the variant is structural either way, never string-inferred.)

**One endpoint renderer.** `v1_display_map` + `v1_endpoint_subject` are extracted and shared by the hashed half and the provenance tail. Your colliding-literal program is the pin, both forms: `relations.publishes[].subject == provenance.publishes[].subject == "events::Changed"`, same for subscribes — the provenance row joins the relation row it locates.

**Positive adequacy.** `exact_bus_endpoints` splits into independent `exact_publishes` / `exact_subscribes` (the relation-level positive account you asked for), and `family_adequacy` requires `required ⊆ vouched` — holes revert to being the validation cross-check via the existing `CapabilityContradiction` law. Your exact model state (`exact_effects = false`, no holes, validates Ok) is the pin: `certificate` reads `degraded`, because absence of recorded unknowns is not proof. The round-2 subscriber-only pin now passes through the positive account: publishes stay vouched, certificate stays `exact`.

**Closed-vocabulary admission.** Legacy `claims` rows now carry the law `ordinal` they project from; the renderer joins by ordinal, so a same-name row cannot masquerade. Admission enforces: recognized law kinds with per-kind complete payload shapes (a bare `{"kind": "forbid_reaches"}` is refused by name), resolved group/topic references existing in the artifact's own `groups`/`topics` sections, unique contiguous ordinals from zero, typed `capabilities`/`adequacy` contents, and the exact one-to-one claims-to-law projection with name and verdict agreement. Both requested tamper controls are in the restamped suite: the gutted payload (refusal names "incomplete law payload") and the duplicated same-name row (refusal names the broken contiguous sequence).

Workspace 2801/2801; both corpus differentials byte-green.

rileyr:
Round 4 fixed at 4b1ea34.

**Lossless payload (P1-1).** `during` and `seed` serialize as typed references with resolution status and provenance — the unresolved-`during boot` case is pinned (`resolved: false` survives). `QuantDimIr` renders as a typed dimension: `{"builtin": <closed tag>}` or `{"user_class": <full class reference>}` (pinned with a resolved `money` budget). Every operand reference now carries its own provenance via the shared `loc` renderer (annotation subjects anchor at the row's provenance, which the row already serializes — documented at the render site). The four fleet variants get distinct tagged payloads carrying their operands.

**Evidence, not bare verdicts (P1-2).** `ProjectedLawRow` keeps each judgment's ordered diagnostics, resolved to `(message, file, span)` and serialized as `evidence` on the row; certificate `certs` keep their per-certificate diagnostics from the evidence sidecar. The pin shows a violated reachability law shipping its countermodel diagnostics with locations, and a violated publish certificate keeping its root/leaf diagnostics. (Family-typed witness structures — path steps, count chains — are a natural Change-7/9 refinement; the round's minimum bar, ordered diagnostics with provenance, is met in full.)

**Semantic binding (P1-3).** The new `topology_law` module decodes every payload into a typed `Law` enum; admission then canonically **re-renders the claims-tier form and requires byte agreement** — your operand-swap tamper (groups replaced, form kept, digest restamped) is control 8, refused with the form binding named. It also binds `source ↔ origin` and `kind ↔ family`, joins claims ↔ law in **both** directions (ordinal sets must be identical), and **recomputes the document verdict** — your clean-over-violated tamper is control 9, refused with the recompute named.

**Real closed decoder (P1-4).** Exact vocabulary throughout: `via` elements ∈ {calls, bus} and non-empty, verbs ∈ {publish, subscribe}, comparators ∈ {==, <=, >=}, dimensions closed-or-user-class, `during`/`avoiding` validated when present, selector elements typed, class references' `builtin` flags checked against the language's builtin set (`is_builtin_effect_class`) with builtin ⇒ resolved, and resolved fn/locus/group/topic references checked against the artifact's own `sorts`/`groups`/`topics`. `capabilities` must carry exactly the 10 known boolean flags — no extras, no absences — and `adequacy` is **recomputed from the positive capability account** rather than trusted (exact key set of the 5 migrated families).

Workspace 2803/2803; both corpus differentials byte-green. The Track A CLI suite round-trips real artifacts through the full decoder — emitter and decoder agree end-to-end.

rileyr:
One addendum while you re-review: the re-issued round-4 text contained a thread the round-4 commit had not covered — **fleet admission trusted `verdict: "clean"` as a string**. Closed at the new head: `fleet.rs` now recomputes the verdict from the component's own claims/lowered results and non-fleet law rows (the same rule Track A admission applies), and refuses a schema-1.11 artifact with no law rows as an unverifiable law account. Control: `a_restamped_lying_verdict_is_refused` — a restamped component carrying an injected violated law row under a lying `clean` is refused past the integrity gate with the recompute named. Workspace 2804/2804.

Everything else from the re-issued list was already in at 4b1ea34 (typed `during`/`seed`/dim refs with provenance + distinct fleet payloads; `evidence` on law rows and certs; the `topology_law` closed decoder with form re-render, both-direction join, and verdict recompute; the exact capability set + recomputed adequacy) — the pins are `payload_refs_are_lossless`, `law_rows_carry_evidence`, and tamper controls 8/9.

rileyr:
Round 5 addressed in 6090d46 — all five P1s, plus one the warnings surfaced.

**P1-1 — module-scoped law subjects.** The `law` section now carries two catalogs: `fn_universe` (every function display the model knows — wider than the legacy `sorts.fns` universe) and `effect_classes` (with declared/cyclic status). Annotation subjects resolve against `fn_universe`, so a `module billing { @effects(...) }` artifact admits through Track A. Pinned end-to-end (`module_scoped_annotation_subject_admits` sends a compiler-emitted artifact with a module-scoped subject through render).

**P1-2 — annotation laws bound to their evidence.** Both named attacks are now tamper controls that refuse under a *recomputed* digest:
- `annotation_class_swap_is_refused`: swapping the forbidden class while keeping the old certs fails because expected cert forms are re-rendered from the typed operands — the refusal names the recomputed form (``expected form `forbid reaches({pure_math}, effects(alloc))` ``).
- `budget_operand_mutation_is_refused`: `per_call` 4→0 fails because a budget verdict is only admissible with a `lowered` evidence row matching the form re-rendered from the typed payload; the stale passing row no longer matches.
Certificate verdicts also recompute per-row from the evidence (max severity, class-catalog `invalid` justification).

**P1-3 — impossible refs.** Decoder v2: exact per-kind object shapes with unknown-field rejection, full operand retention on every variant, resolution against the artifact's own catalogs (loci, groups, topics, phases incl. lifecycle names, seeds, effect classes, `fn_universe`), and `holds` over any unresolved operand refused — flipping `resolved: false` cannot rescue a verdict (pinned). The `name` half is now load-bearing too: raw↔display must stay one-to-one across the whole law section, so a `name` swap under an unchanged `display` refuses (`raw_name_swap_under_unchanged_display_is_refused`). Residual, stated plainly: *absolute* raw-name validation isn't possible from inside the artifact because the surface deliberately renders in author spelling and carries no raw catalog for fns/groups; the consistency ledger is the maximal in-artifact check. If you want raw catalogs serialized, that's a surface-widening call I left to you.

**P1-4 — foreign spans never guessed.** `Judged` carries a per-diagnostic source-space discriminator, filled by `judge_certificates` from the evidence sidecar's `ForeignSpan` records; `project_law_rows` never re-resolves a foreign diag against bundle sources. The pin is self-calibrating (`foreign_cert_evidence_gets_no_guessed_location`): it measures the stdlib alloc-leaf offset, pads the user file until that offset falls numerically *inside* it — where a containment guess would misfile stdlib evidence as application code — and requires the artifact evidence row to carry no location while keeping its message.

**P1-5 — one shared admission.** `validate_law_account` (new `topology_law` module) is called identically by Track A rendering and `hale fleet check`, after the digest/semantics gates: closed decode, evidence binding, both-direction claims↔law join, capability/adequacy recompute, document-verdict recompute. Fleet consequences pinned:
- `deleted_law_rows_are_refused`: a claims-bearing component with `law.rows` gutted and restamped refuses on the two-way join ("claims[0] does not project one-to-one from law ordinal 0"), not vacuously accepted.
- `a_restamped_lying_verdict_is_refused` now documents that the deep decode bites *before* the verdict recompute — the injected bare payload is refused as incomplete; either way the law account speaks and the test asserts the digest is not what refused.

Docs: spec/verification.md (catalogs, shared admission, foreign-span honesty), docs/src/claims.md (artifact shape + the "digest is only the integrity gate" paragraph), CHANGELOG round-5 entry.

Workspace 2810/2810, both corpus differentials byte-green, zero build warnings.

rileyr:
Round 6 addressed in 6adaa04 — all five P1s.

**P1 — annotation-only law deletion.** Every compatibility `lowered` row now carries its law `ordinal` (and `cert` ordinal for certificate rows; the legacy budget/quant producers are keyed by `ClaimRow::budget_lowered_form` consumed in table order). Admission builds the exact expectation multiset from the typed rows — forms, results, *and subjects* re-rendered from the operands — and requires the lowered section to project one-to-one from it: no orphan row on either side. The named attack is pinned (`annotation_only_law_deletion_is_refused`): gutting `law.rows` under a restamped digest on an `@effects`-only artifact refuses with "is not claimed by any typed law row".

**P1 — raw identities anchored.** The law catalogs are now canonical `{name, display}` pairs — `fn_universe`, `loci`, `groups`, `topics` (which also carry their wire `subject`), plus the wire-subject universe — and every resolved reference must match one **exact pair**. The singleton swap you constructed is exactly what this closes: `raw_name_swap_under_unchanged_display_is_refused` now runs against a single-occurrence reference and refuses on "does not match any canonical pair" (the round-5 ledger stays as a secondary consistency check). The catalogs cross-tie to the `topics`/`groups`/`sorts` sections so they can't be extended as a side channel.

**P1 — selector candidates bound.** The decoded `Selector` retains its full candidate sets; each topic candidate must be a canonical pair, each subject candidate must be in the subject universe, and — the binding — admission **recomputes** both sets from the catalogs with the compiler's own matching rule (`hale_model::bus_ref_matches`, the same single definition the lowering delegates to, so this is not a re-derivation of a second rule). The Allowed→Denied swap under an unchanged selector name is pinned (`selector_candidate_swap_is_refused`).

**P1 — causes/depends verdicts keyed to their law.** New `law.legacy` report: `{ordinal, form, result}` per unmigrated non-budget row that carries an imported verdict, where `form` is the fingerprint re-rendered from the typed operands (`ClaimRow::legacy_form` in the emitter, `expected_legacy_form` in admission — a byte-agreeing pair like the cert/budget renderers). Admission requires the report to project one-to-one from the causes/depends rows: an operand mutation orphans the entry (pinned: `causes_operand_swap_is_refused`, money→spare with both classes declared), and orphan entries are refused in the other direction. The class-validity account now covers `causes:` too — a row naming an undeclared or cyclic class cannot hold. (The compiler refuses to emit such an artifact at all, so this rule only ever bites tampered documents.)

**P1 — fleet rows refused.** Taken as recommended: until Change 7 owns a typed fleet account, application-artifact admission refuses fleet rows **by name** — family or origin `fleet`, and all four `fleet_*` kinds in the decoder (`Law::Fleet` no longer exists). They can never be excluded from the document verdict because they never admit (pinned: `fleet_row_injection_is_refused`, your exact row shape with `verdict: violated` under `clean`). Since fleet composition shares `validate_law_account`, the same refusal holds there.

Docs: spec/verification.md (catalog pairs, lowered bijection, legacy report, fleet refusal), docs/src/claims.md (artifact shape + admission paragraph), CHANGELOG round-6 entry.

Workspace 2814/2814, both corpus differentials byte-green, zero build warnings.

rileyr:
Round 7 addressed in 0a0b0f0 — all four P1s.

**P1 — static invalidity dominates.** The rule is now exactly as you stated it: `has_unresolved || law_class_invalid` (a full traversal over every class-ref operand, checked against the artifact's own class catalog) forces the verdict to be **exactly `invalid`** — for every family, so `violated`/`uncertified` can't dress up a malformed law either — and otherwise a certificate verdict must equal **exactly** the recomputed evidence severity. Your semantics-2 counterexample is the pin (`cyclic_class_verdict_flip_is_refused`): the honest cyclic artifact admits, and flipping `invalid`→`holds` + `clean` refuses even though the recomputed cert severity *is* holds. The two adjacent manifestations closed with it: the empty-certs escape is gone — inside the analyzable universe certs always bind, so a resolved `@no_panic` asserting `invalid` with discarded certs refuses on the count (pinned); outside the analyzable universe (`sorts.fns`, which module-scoped fns are deliberately absent from) a row must carry *no* certs and only `invalid` — matching what the emitter actually produces for module subjects. Non-certificate rows may only judge `invalid` for a decodable reason.

**P1 — cyclic artifacts self-admit.** Confirmed both of your compiler-produced cases failed their own admission at 6adaa04; both now admit (pinned: `cyclic_unmigrated_artifacts_self_admit`, plus the `none:` shape in the flip pin). Expectation basis is now what the old engine produced, not the machine verdict: an `invalid` unmigrated row's `law.legacy` entry / keyed budget `lowered` row becomes an **optional** expectation — fingerprint and subject still bind (an operand mutation still orphans it), the preserved result is constrained to the old engine's vocabulary, and it never overrides the machine verdict. `holds`/`violated` rows still *require* their entry; `uncertified` rows still admit none.

**P1 — catalogs closed.** All the closure you listed: every catalog unique in both halves (raw and display; class names; subject patterns), `law.topics` ↔ topics section and `law.groups` ↔ groups section and `law.loci` ↔ `sorts.loci` in exact bijection (count + membership), `fn_universe` ⊇ `sorts.fns`, and `law.subjects` tied to the model — every entry must be a declared topic's wire subject or a relation-endpoint subject, and every declared subject must be present. Your exact construction is the pin (`ghost_catalog_widening_is_refused`): `Ghost::Allowed` + its wire pattern appended to the catalogs refuses on the closed-bijection check before recomputation can be fooled.

**P1 — digests recomputed, evidence validated.** `law_digest` is redefined as a *recomputable external fingerprint*: the canonical-JSON rendering of the law rows (serde_json's deterministic BTreeMap serialization), fnv1a64 — the emitter computes it from its own serialized rows, admission recomputes it from the parsed document, and the projection test proves both sides agree byte-for-byte. A row edit under a stale `law_digest` refuses even with a fresh `artifact_digest` (pinned). `inputs_digest` must equal the consuming binary's own `analysis_inputs_digest()` — evidence produced under a different stdlib/analysis snapshot is refused with a re-dump message (pinned), per the evidence engine's stated contract. Evidence objects are structurally validated everywhere (non-empty message, paired file/span resolving to a known source); migrated `violated`/`uncertified` rows must retain their judgment's ordered evidence, and violated certificates their diagnostics — with the pleasant consequence that the round-4 lying-verdict tamper now fails one step earlier, for lacking the countermodel it never had. Tamper controls restamp *both* digests so each still exercises the deep binding it targets.

Docs: spec/verification.md, docs/src/claims.md admission paragraph, CHANGELOG round-7 entry. serde_json promoted from dev-dependency to dependency in hale-types for the canonical rendering.

Workspace 2820/2820, both corpus differentials byte-green, zero build warnings.

rileyr:
Round 8 addressed in 2c83673 — all three P1s, each verified against your exact reproducer first.

**P1 — every compiler-emitted Invalid admits.** Confirmed your `require attributed(all purpose)` artifact was refused at 0a0b0f0. On the two options you offered: reproducing all of the evaluator's validation rules at admission would be a second implementation of ~20 judgment rules that would drift, and a parallel reason-enum would be a channel admission cannot independently verify for judgment-internal rules — so the serialized reason is the judgment's own retained explanation: an `invalid` row must carry either a *statically decodable* invalidity (unresolved operand, undeclared/cyclic class — where dominance still forces exactly `invalid`) or its judgment's ordered evidence, which every legitimate evaluator Invalid produces and which round 7 already validates structurally. Your artifact admits (pinned, `operand_domain_invalid_admits`), and the same pin strips the explanation and shows the bare `invalid` still refuses. This holds for operand-domain errors, vacuity, empty `during`, and `avoiding` overlap uniformly.

**P1 — PhaseEffects matches the evidence producer exactly.** Both of your options, split by case: the evidence layer now synthesizes a `Holds` certificate for an implicit lifecycle phase on a **top-level** locus (no hook body performs no effects — your check-clean `@phase_effects(birth: {})` artifact now judges `holds` and admits, pinned), and a report-less subject — a module-scoped locus, and uniformly any subject the engines never analyzed — judges **`uncertified` with its residue on the row**, never Invalid (`judge_certificates`' missing-report branch now distinguishes absent/empty report = residue from partial disagreement = still Invalid; static class invalidity still dominates both). To keep admission's analyzability rule exactly aligned with the producer, `law.loci` rows carry an `analyzable` flag (module loci appear in every sort, so the discriminator has to be explicit) — which also closes the delete-certs-and-claim-uncertified downgrade for loci. The fn-subject side moved with it for consistency: module-fn certificate rows are now `uncertified`-with-residue too. The corpus differential carves out the synthetic rows as an explicit documented divergence and still requires the legacy subsequence to match in order and in full.

**P1 — the subject universe validated against its own typed projection.** New unhashed **`endpoints`** section: every bus endpoint at wire-subject grain — `{verb, subject, via: site | declaration}` — projected from the model's `publishes`, `declares_publish`, and `subscribes` relations, so a declared publisher with no send site is a visible row. Admission requires `law.subjects` to equal **exactly** the subjects the endpoint and topic sections carry, in both directions; the relations-based reconstruction is gone. Your `bus { publish "unused.address" of type Msg; }` artifact admits (pinned, including the presence of the `via: "declaration"` row), and the reverse gap you named is pinned too: deleting the subject from `law.subjects` while the endpoint row remains refuses on the equality (`declared_endpoint_without_site_admits_and_closes`). One note for the cross-seed case: the endpoints section is wire-subject grain and therefore deliberately raw, like `topics` — a subject-less imported topic's default subject *is* its mangled identity — so the xseed author-spelling test excludes it alongside the two existing raw sections.

Docs: spec/verification.md (analyzability + synthetic certificates + evidence-backed invalidity + endpoint universe), docs/src/claims.md artifact shape, CHANGELOG round-8 entry.

Workspace 2823/2823, both corpus differentials byte-green, zero build warnings.

rileyr:
Round 9 addressed in d3a0560 — all three P1s.

**P1 — table-level failures reach the artifact.** `project_law_rows` now returns the law-selection account: `ClaimIrTable::issues` plus the judgment pre-pass it previously discarded as `_pre`, each with its source location. It's serialized as `law.issues`, participates in `law_digest` (the canonical fingerprint now covers `{issues, rows}`) and in the document verdict (a non-empty account is `law_failed`, recomputed on both the emitter and admission sides), and admission validates each entry like any other diagnostic. Your exact scenario is the pin (`duplicate_claim_names_cannot_be_clean`): two individually-holding duplicate-name claims fail `hale check`, emit two `holds` law rows *plus* the issue, get `law_failed` (not `clean`), and admit as an honest failing artifact — while emptying the issues and flipping the verdict refuses, because the duplicate-name case is **recomputed from the rows themselves** ("duplicate claim name `iso` with an empty law-selection account"). Constitution/adoption/collision failures flow through the same account; those aren't row-recomputable, but deleting them now also has to defeat the digest-covered verdict recompute, which the issues participate in.

**P1 — one canonical authority for both facts.**
- *Endpoints*: the section must now project **exactly** from the artifact's own relations — site rows from `relations.publishes`/`relations.subscribes` (V1 rows name declared topics by display; the wire subject comes from the topics section, so the mapping is recomputable), declaration rows from the new typed **`declares_publish`** relation section (locus + subject, locus validated against the catalog). Your `audit.log` construction is the pin (`endpoint_narrowing_against_relations_is_refused`): deleting the endpoint row while the actual publish relation remains refuses on "does not project from the artifact's relations".
- *Analyzability*: `LocusDecl::analyzable` is set once by the **model builder's** declaration walk; the evidence layer and the emitter both read the model — the two downstream AST walks are deleted, per the architecture rule you cited. And admission doesn't trust the Boolean either: it **recomputes** it against the hashed function universe — a locus's members appear in `law.fn_universe` while the legacy summary universe (`sorts.fns`) contains exactly the analyzed ones, so for any locus with members the flag must agree with its member account. Your flip attack is the pin (`analyzable_flip_is_refused`). The residual honesty note: a *member-less* locus has no independent member evidence — but for such a locus both shapes are vacuously truthful (no code performs no effects in any phase), so the flip cannot manufacture a false `holds`.

**P1 — semantics bumped.** `ANALYSIS_SEMANTICS_VERSION` is 3, with the reason documented at the constant: round 8's synthetic implicit-phase certificates, report-less subjects judging `uncertified`, and the changed completeness rule are result-affecting producer/judgment changes. The version is folded into `analysis_inputs_digest`, so pre- and post-round-8 evidence cannot share an `inputs_digest` — and round 7's admission check (`inputs_digest` must equal the consuming binary's own digest) makes the mismatch a refusal with a re-dump message.

Corpus differentials gained the two documented divergences this round introduces (the residue diagnostic the old engine never produced), and the tamper restamps track the `{issues, rows}` canon.

Docs: spec/verification.md, docs/src/claims.md artifact shape (`law.issues`, `declares_publish`), CHANGELOG round-9 entry.

Workspace 2826/2826, both corpus differentials byte-green, zero build warnings.

rileyr:
Round 10 addressed in 29fcb96 — both P1s, each verified against your exact reproducer first.

**P1 — typed endpoint identity.** Confirmed your colliding artifact (topic `Orders` with subject `"wire.orders"` plus a literal `"Orders"` end, both verbs) was rejected by its own admission at d3a0560. Endpoint rows and `declares_publish` rows now carry `declared_topic` — projected from the model's syntactic fact on `Publish`/`Subscribe`/`DeclaresPublish`, so the literal stays a literal even under a display-name collision. Admission compares the typed identities directly: a topic-typed row's `subject` must equal that topic's wire subject from the catalog; site rows project onto the V1 relations under their **own** declaredness (a covered end by its topic display, a literal end by its text); declaration rows must equal the typed `declares_publish` relation on `(subject, topic)`. The `wire_of` display-string inference is deleted. Your artifact admits (pinned), the retag attack (marking the literal as topic-covered) refuses on the wire-subject disagreement (pinned), and `require publishes(some G, topic Orders)`'s distinction between `publish Orders` and `publish "wire.orders"` is now externally visible on the declares rows.

**P1 — coverage at function grain, in evidence identity.**
- *Grain*: `Function::analyzed` is the model's fact, set once by the model builder from its existing unanalyzed tracking (module bodies **and `on_failure` handlers** — your first case). `law.fn_universe` rows carry the bit, and admission requires the analyzed subset to equal `sorts.fns` **exactly** — the hashed summary universe is precisely the walked set, so the anchor is the hash, not name prefixes. The locus flag recomputes from member coverage (all non-`on_failure` members analyzed); your on_failure-only locus admits (pinned), and the membered module-locus flip still refuses.
- *The memberless case you caught*: rather than leave the gap, the semantics changed — a memberless locus is **vacuously analyzable** (there is no body to walk, so the walk completed trivially and every phase contract holds by absence). The honest memberless module-locus artifact now judges `holds`/`clean`, and the flag is recomputable in *both* directions: marking it `analyzable: false` to manufacture `uncertified` refuses (pinned both ways; the round-8 pin updated, with a new membered-module case keeping the `uncertified` shape covered). Implicit-phase certificates are additionally bound: exactly the synthetic `holds` with no diagnostics.
- *Identity*: `EvidenceTable` carries a coverage digest (`ApplicationModel::analysis_coverage_digest` over the locus `analyzable` and function `analyzed` bits); `EvidenceTable::validate` — the tri-party tie the judgment consults — refuses a sidecar derived beside different coverage as stale. Pinned: flipping one coverage bit on an otherwise-identical model invalidates a previously-holding judgment. `TopologyShapeV1` is untouched, per your note on recording compatibility.

One consequence worth flagging: Track A's older referential check required `phases`/`effects` keys to be in `sorts.fns`, which made any membered module locus inadmissible for an unrelated reason — it now resolves against the full function universe.

Docs: spec/verification.md, docs/src/claims.md artifact shape, CHANGELOG round-10 entry.

Workspace 2830/2830, both corpus differentials byte-green, zero build warnings.

rileyr:
Round 11 addressed in 1f3ed08 — both P1s.

**P1 — the typed site projection is lossless.** Site endpoint rows now carry their owning fn/handler and authored site ordinal (from `Publish.function`/`site` and `Subscribe.handler`/`site`), so the two rows in your collision fixture are `("Emitter::emit", 0, "wire.orders", Some("Orders"))` and `("Emitter::emit", 1, "Orders", None)` — distinct typed facts that no longer collapse. Admission validates owners against the function universe, enforces row uniqueness, projects the image onto the V1 relations at **(owner, name)** grain, and — the piece that makes it lossless against the dedup'd legacy rows — requires the per-owner row **counts** to equal the span-grained provenance section, which keeps one row per authored site. I ran your full narrowing attack (literal site row + declaration row + `declares_publish` row + subject + candidates removed, topic end retained, both digests restamped) and it refuses at every depth: the declares tie catches the shallow variant, the owner-grain subscribe tie and the provenance publish-count catch the deep one. Pinned (`colliding_literal_narrowing_is_refused`). Your framing was exactly right — the image comparison proved compatibility, not survival; the counts and owner grain are what prove survival.

**P1 — the three coverage states are typed and distinct.** `Function::summarized` (a behavior-summary row exists) joins `analyzed` (the body was walked), and "a certificate engine emitted a report" remains the evidence account — three states, never conflated. Failure handlers carry the typed `FunctionKind::FailureHandler`; the display-prefix test is deleted, so your `on_failure_helper` module method is a real unanalyzed member, the locus is honestly `analyzable: false`, and its artifact admits (pinned, including the typed `kind: "method"` assertion). The coverage laws — `summarized ⇒ analyzed`, failure handlers never analyzed, the legacy fn sort equals the summarized set — are validated in `ApplicationModel::validate` (new `ModelError::CoverageLaw`), re-checked at admission, and folded into the evidence coverage digest along with the kind. The admission tie moves to the right universe: the **summarized** subset of `law.fn_universe` must equal `sorts.fns` exactly, and certificate analyzability keys on the walked bit.

One factual note from verification: your `fn noop() {}` case does not reproduce at this codebase state — I probed it before changing anything, and `noop` appears in `sorts.fns` (the legacy sort is built from the summary key set, and the summary does carry rows for empty free fns; the artifact admitted under the old tie too). The new coverage law states that identity explicitly (`topology_v1_fns == summarized set`, validated in the model), so if the universes ever do drift the way you described, it becomes a model-validation failure at derive time rather than a silent admission mismatch downstream.

Docs: spec/verification.md, docs/src/claims.md artifact shape, CHANGELOG round-11 entry.

Workspace 2832/2832, both corpus differentials byte-green, zero build warnings.

rileyr:
Round 12 addressed in 11fd01a — all three P1s, with one residual surfaced as a design decision rather than silently absorbed.

**P1 — exact typed endpoint equality.** All three structural asks implemented: site rows enforce exactly one occupant per **(verb, owner, site)**; declaration rows keep the **owning locus** in the compared identity on both sides (`endpoints` declaration rows and the `declares_publish` relation now serialize and compare `(locus, subject, topic)`, so the declaration-move attack against `require publishes(some G, topic T)` refuses on the tuple); and every site/declaration row is **anchored to the typed provenance account** — rows carry their authored span, and the per-owner span *multisets* must equal the span-grained provenance section one-to-one, replacing the count tie (deleting or duplicating one of two colliding rows leaves the spans behind).

The substitution variant you described — replacing the literal site-1 row's `(subject, topic)` with a copy of site-0's identity at the *same span* — I verified end-to-end and it still admits. Root cause, stated plainly: the wire identity of a display-colliding literal exists only in unhashed sections (the model half's relations spell both ends `"Orders"`, and the legacy gathering cannot reconstruct per-span wire identity, which is exactly the lossiness this thread has been unwinding). Every unhashed fact in this artifact is held by cross-section consistency, and this edit is consistent across all of them. Closing it requires putting endpoint wire identity into the **hashed half** — a shape-identity change for every bus-carrying program, breaking the "shape_hash unchanged" property this PR has preserved (and the recorded replay baselines with it). That's a maintainer decision I didn't want to take unilaterally in round 12; if you want it, I'd propose a dedicated hashed `endpoint identity` line under the existing MODEL_SEMANTICS=2 umbrella as the follow-up. The spec documents the residual with this exact framing.

**P1 — one typed failure-handler rule.** The builder no longer counts `Failure` members toward module-locus unanalyzability — a failure handler is never analyzed anywhere, so it cannot make a locus unanalyzable — matching admission's typed-kind exemption exactly. Your module-scoped failure-only `Guard` now emits `analyzable: true` (symmetric with its top-level equivalent) and admits (pinned).

**P1 — the coverage account is closed.** Three layers: (1) `ApplicationModel::validate` — an unanalyzed body must retain its `UnanalyzedBody` residue and an analyzed body carries none, so `analyzed` derives from the hole account rather than floating free. (2) The upgrade path is anchored to the hash: a walked body is summarized (corpus-verified — this is the converse of round 11's law, and together they pin `analyzed` to `sorts.fns`), so your exact flip — module `f` marked `analyzed: true` with a manufactured Holds certificate and removed residue — refuses on "analyzed but not summarized — the walked set is the summary set" against the hashed universe (pinned). (3) `EvidenceTable::validate` categorically refuses certificate payloads for subjects or phases whose typed coverage says no report exists, resolved through the claim rows' typed subject ids — pinned: a full certificate for a subject flipped to unanalyzed, even with a *recomputed matching coverage digest*, judges Invalid, never Holds — your closing sentence verbatim: the digest proves the sidecar repeated the model's bits; the eligibility check proves its evidence obeys them.

Docs: spec/verification.md (span anchoring, residual, closed coverage), docs/src/claims.md artifact shape, CHANGELOG round-12 entry.

Workspace 2835/2835, both corpus differentials byte-green, zero build warnings.

rileyr:
Round 13 addressed in 473a87a — both P1s, each verified against your reproducer first.

**P1 — declared-end identity includes the typed topic.** The canonical key for `declares_publish` is now **`(locus, subject, declared_topic)`** in all three places you named: the builder's collection map (the `or_insert` first-writer collapse is gone — the map key itself carries the typed distinction, so both ends survive structurally), `ApplicationModel::validate`'s sorted-key law, and the schema documentation table. I built your exact fixture in **both declaration orders** and confirmed both typed rows reach the artifact each time — the literal `("wire.orders", topic: none)` and the typed `("wire.orders", topic: Orders)` as separate `declares_publish` rows with their own spans — and both artifacts admit. `require publishes(some G, topic Orders)` no longer depends on which line the author wrote first. The pin (`colliding_declarations_both_survive`) runs both orders and asserts both rows plus admission; since the corpus provider harvests `.hl` from test files, the shape is now exercised by both corpus differentials as you asked.

**P1 — one closure membership rule.** Before choosing between your two options I probed the decisive fact: a **top-level** closure-only locus with `@phase_effects(birth: {})` produces no function entity for the closure and certifies synthetically — the certificate machinery never walks closure bodies at *any* scope (neither `effects.rs` nor the allocation summary reference closures at all). So the honest, consistent rule is your second option: closures don't participate in analyzability anywhere. The builder's module-executable set now counts exactly the members that produce function entities the engines could walk — `Fn | Lifecycle | Mode` — which is precisely admission's member-coverage universe minus the typed failure exemption. Your closure-only `Hidden` now emits `analyzable: true` (symmetric with its top-level twin) and admits (pinned). Worth stating the consequence plainly: a phase contract on a closure-carrying locus is blind to closure effects — but that is a pre-existing property of the certificate engines at top level, not something this rule introduces; if closure bodies ever become walkable, they'll need function entities and the coverage account will carry them like any other member.

Docs: spec/verification.md (canonical identity + closure rule), CHANGELOG round-13 entry.

Workspace 2837/2837, both corpus differentials byte-green, zero build warnings. The hashed-wire-identity question from round 12 remains open on your side.

rileyr:
Round 14 addressed in 1c3dcd4.

**The missing model laws are in.** `ApplicationModel::validate` now enforces both laws admission already had, exactly as specified:

- **`Function.analyzed ⇒ Function.summarized`** — joining the existing converse, so `analyzed ⇔ summarized` holds at the model and the bit is pinned to the hashed legacy fn sort there too. Your function-grain path — `analyzed = true`, `summarized = false`, `UnanalyzedBody` hole removed so the residue law is silent — now fails validation (`function_coverage_upgrade_fails_model_validation`).
- **`LocusDecl.analyzable == all members except FailureHandler are analyzed`** (empty relevant set ⇒ true) — derived from the **typed `member_of` relation and `FunctionKind`**, per your note, not display prefixes. Your locus-grain path — flip `analyzable` only, unanalyzed method and its hole intact — now fails validation (`locus_coverage_upgrade_fails_model_validation`).

**The sidecar no longer trusts the unverified bits either.** Since the internal API can be handed a model that was never validated, `EvidenceTable::validate`'s eligibility hardened to match rather than assuming lawfulness: fn subjects require `analyzed && summarized` (the hashed anchor — an upgraded bit without a summary row is not a reportable subject), and locus subjects' member coverage is **recomputed from `member_of`** instead of trusting `LocusDecl::analyzable`. The two manufacture tests you asked for do the strongest version of the attempt — the sidecar is *derived against the upgraded model*, so shape and coverage digests match by construction and a well-formed certificate exists — and both still fail to produce Holds (`function_coverage_upgrade_cannot_manufacture_holds`, `locus_coverage_upgrade_cannot_manufacture_holds`).

The three-layer agreement you described is now literal: the same coverage state is lawful or refused identically at the model validator, the evidence/judgment API, and artifact admission — with the model laws, the sidecar recompute, and the artifact checks all reading the same typed facts (`member_of`, `FunctionKind`, the summarized anchor).

Docs: spec/verification.md (both laws + hardened eligibility), CHANGELOG round-14 entry.

Workspace 2841/2841, both corpus differentials byte-green, zero build warnings. Still open on your side: the round-12 hashed-wire-identity design question.

rileyr:
Round 15 addressed in 7ef5698.

**Ownership is closed.** Taken as your first option, with the second folded in: `Function` gains its canonical **`owner: Option<LocusDeclId>`** (set by the model builder's single declaration walk — `Some` for methods, hooks, modes, and failure handlers; `None` for free functions), and `member_of` must be a **total, exclusive partition agreeing with it exactly** — a free function appears in no row; every other kind in exactly one row, at its canonical owner. All four structural laws you listed are enforced. Ownership is coverage-bearing: the owner is folded into the evidence coverage digest, since moving a member changes whose coverage it counts toward — and, as you noted, group projection hangs off the same relation, so the partition law protects reachability's locus-to-function projection too, not just phase evidence.

**One shared validator.** Everything from rounds 10–15 — the partition, `analyzed ⇔ summarized`, failure handlers never analyzed, the `UnanalyzedBody` residue agreement, owner-derived locus analyzability (via the shared `locus_members_analyzed`, which reads the closed owner account), and the summarized/legacy-sort identity — now lives in **`ApplicationModel::validate_coverage`**, called by `ApplicationModel::validate` (last, so purely structural defects still report as themselves) and by **`EvidenceTable::validate` before any digest comparison**. The sidecar's weaker member scan is deleted; per your closing line, a model whose ownership account is corrupted cannot certify anything, digests notwithstanding.

**Both negative controls, at both layers.** `deleted_membership_is_refused_everywhere`: removing `member_of(Hidden::poke, Hidden)` and flipping `Hidden.analyzable = true` — your exact counterexample — now fails model validation on the partition law (poke's canonical owner has no matching row), and a sidecar derived *against* the mutated model (digests matching by construction) still cannot produce Holds. `moved_membership_is_refused_everywhere` does the stronger consistent move — row *and* owner repointed to `App` together, since a row-only move dies immediately on the agreement law — and fails on the destination's coverage law (`App` now owns an unanalyzed member while marked analyzable), with the sidecar refusing identically.

Docs: spec/verification.md (closed ownership account + shared validator), CHANGELOG round-15 entry.

Workspace 2843/2843, both corpus differentials byte-green, zero build warnings. Still open on your side: the round-12 hashed-wire-identity question.

rileyr:
Round 16 addressed in the latest commit.

**Ownership is anchored to the entity identity.** `validate_coverage` now enforces exactly the law you specified: for `owner = Some(l)`, the owner id is range-checked directly (its own `DanglingId`, not an indexing accident), the function's **raw name** must be prefixed by `loci[l].name + "::"`, and its **display** by `loci[l].display + "::"` — with `owner = None ⇔ Free` already in place from round 15. Your fully coordinated laundering — owner and `member_of` row repointed to `App`, `App.analyzable = false` agreeing with its new unanalyzed member, `Hidden.analyzable = true` agreeing with its emptied owner set, every relational law satisfied — is refused precisely because `Hidden::poke` cannot canonically be owned by `App`. The negative control (`coordinated_ownership_laundering_is_refused`) asserts the refusal is the *identity anchor specifically* (it matches the law text, not just any `CoverageLaw`), and the sidecar arm derives evidence against the mutated model and still cannot manufacture Holds. This also re-closes the group-projection path you flagged: a member cannot leave its canonical locus's groups by relational edits, because the relation must mirror an identity-anchored owner.

**The artifact carries the typed owner.** `law.fn_universe` rows gain `"owner"` — present iff the kind is non-free, required to resolve to a cataloged locus, and display-anchored with the same refusal ("cannot canonically be owned by") — and admission's locus member coverage now recomputes from the **typed owner account**; the display-prefix recovery is deleted. The artifact-layer control (`artifact_owner_swap_is_refused`) performs the same coordinated edit on the serialized document (owner retag plus both flag updates, both digests restamped) and refuses on the identity anchor.

Docs: spec/verification.md (identity anchoring + typed artifact owner), docs/src/claims.md artifact shape, CHANGELOG round-16 entry.

Workspace 2845/2845, both corpus differentials byte-green, zero build warnings. Still open on your side: the round-12 hashed-wire-identity question.

PR #490, opened 2026-08-20
feat(fleet+artifact): Change 7 — schema 1.12 hashed endpoint identity + typed FleetModel (GH #476)

rileyr:
GH #476 Change 7, in two commits.

**Commit 1 — schema 1.12: canonical endpoint identity joins the hashed shape.** The versioned transition discussed at the end of #489: the V1 relations render a topic-covered end and a display-colliding literal as one spelling, so `shape_hash` could not distinguish two systems that talk to different wire addresses — which is exactly what the shape exists to distinguish, and what replay admission relies on. The hashed model half gains an `endpoint_identity` section (verb, owner, source-order site ordinal, byte-exact wire subject, declared topic; emitted only when endpoints exist, so bus-free programs keep their identity), and admission requires the unhashed endpoint sections to agree with it exactly. The round-12 substitution residual flips from admitted to refused (pinned): substituting a colliding literal's content now contradicts the hashed half, and rewriting the hashed half changes the program's identity.

Costs, taken deliberately and documented in CHANGELOG + spec: **shape hashes change for every bus-carrying program** — shape baselines and `.halerec` recordings need one re-record. `TOPOLOGY_SCHEMA` bumps to 1.12 as the explicit marker. The corpus differential keeps its V1 charter (the new section is stripped before the byte comparison — it is content the legacy gathering never produced, and cannot, which was the whole finding). Goldens regenerated.

**Commit 2 — the typed `FleetModel`.** `fleet_model::ComponentModel` decodes each admitted component artifact ONCE — after signature, integrity, semantics, verdict, and the shared law-account admission — into exactly the typed rows composition consumes: the vertex universe, both call relations unioned, the V1 publish/subscribe rows, the topics join surface, the component's unknowns, decl provenance for witness locations, and the hashed endpoint identity. `Component.artifact: serde_json::Value` is gone; every crawl site and all five `Value`-taking helpers are replaced by typed accessors. The only `serde_json::Value` left in fleet.rs is the parse at the decode boundary — the #476 architecture canary ("fleet code cannot walk generic JSON for a modeled fact after cutover") holds.

One behavioral strengthening rides the cutover: `has_endpoint` ("declaring a topic is not using it") now reads the hashed endpoint identity at byte-exact wire grain — a publisher end needs a send-site row, and a display-colliding literal can no longer satisfy a check aimed at the topic.

Workspace 2846/2846, corpus differentials green on their charter, zero build warnings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,16 @@ behavior.
 
 ### Typed FleetModel + the versioned shape transition (GH #476 Change 7)
 
+Fleet composition is TYPED end to end: component artifacts are
+decoded ONCE into a typed ComponentModel (fns, unioned call
+relations, V1 publish/subscribe rows, the topics join surface,
+unknowns, decl provenance, and the hashed endpoint identity)
+immediately after the shared admission - the rest of composition
+never walks generic JSON for a modeled fact (the #476
+architecture canary). Route endpoint checks now read the
+byte-exact hashed wire identity: "declares but never sends" is
+judged from typed site rows, collision-proof.
+
 Schema 1.12: canonical ENDPOINT IDENTITY joins the hashed model
 half (`endpoint_identity`: verb, owner, source-order site
 ordinal, byte-exact wire subject, declared topic). The V1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,16 @@ behavior.
 
 ### Typed FleetModel + the versioned shape transition (GH #476 Change 7)
 
+Round 3: the typed decode is strict, and keys are unambiguous.
+ComponentModel::decode refuses malformed semantic rows instead of
+filtering them - a number-typed call endpoint or unknown reason
+is an error, never a silently dropped edge or erased residue
+(both pinned against components that genuinely carry the shapes,
+with asserted premises). Duplicate object keys are rejected
+before parsing at both consumption boundaries: serde's last-wins
+map parse would otherwise let a second shape_hash shadow the
+raw-verified one (pinned).
+
 Round 2: route identity is one grain, and the declared identity
 is verified. Route role checks and edge construction both read
 the typed topic-identity endpoint rows (a literal end on the same
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,15 @@ behavior.
 
 ### Typed FleetModel + the versioned shape transition (GH #476 Change 7)
 
+Round 2: route identity is one grain, and the declared identity
+is verified. Route role checks and edge construction both read
+the typed topic-identity endpoint rows (a literal end on the same
+wire satisfies neither; wire-grain accessors remain for the fleet
+claims that quantify over wire subjects). verify_shape_hash
+recomputes the model-half hash from the raw artifact at both
+consumption boundaries before any decoding - a coordinated
+hashed+unhashed endpoint edit under a stale shape_hash refuses.
+
 Fleet composition is TYPED end to end: component artifacts are
 decoded ONCE into a typed ComponentModel (fns, unioned call
 relations, V1 publish/subscribe rows, the topics join surface,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,16 @@ behavior.
 
 ### Typed FleetModel + the versioned shape transition (GH #476 Change 7)
 
+Round 4: the raw admission pass is JSON-semantic and
+path-aware. One structural walk (scan_top_level) drives
+everything: duplicate keys compare DECODED names (an escaped
+spelling cannot smuggle a second shape_hash past the scanner into
+serde's last-wins map), and verify_shape_hash /
+verify_artifact_digest locate their fields at the TOP LEVEL - a
+nested decoy is data, never the verified value (pinned both ways:
+the decoy neither confuses the verifier on an honest artifact nor
+rescues a drifted one).
+
 Round 3: the typed decode is strict, and keys are unambiguous.
 ComponentModel::decode refuses malformed semantic rows instead of
 filtering them - a number-typed call endpoint or unknown reason
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,21 @@ behavior.
 
 ### Typed FleetModel + the versioned shape transition (GH #476 Change 7)
 
+Round 6: completeness with the polarity of the law. The blanket
+holds->uncertified rule now applies only to the absence-certifying
+families (forbid_reaches/only_edges); require_* witnesses are
+positive facts no incomplete endpoint set can erase, and their
+route-backed failures are definite (the plan's route table is
+complete). Counts evaluate over a [known, known+hidden] interval
+with per-verb relevant flags (publish+cardinality /
+subscribe+cardinality): min met by known rows holds, max/eq
+exceeded by known rows violates, the undecided interval is
+uncertified - including the conjunctive eq/min/max form. The
+claim row is serialized after the final verdict (it previously
+recorded the pre-rewrite result). Track A is capped at exactly
+the current schema; the canonical layout table is versioned by
+it.
+
 Round 5: layout is law, and the completeness account is
 honored. The top-level key sequence is canonical and closed -
 order defines the verified hash ranges, so moving a model-half
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,20 @@ behavior.
 
 ### Typed FleetModel + the versioned shape transition (GH #476 Change 7)
 
+Round 5: layout is law, and the completeness account is
+honored. The top-level key sequence is canonical and closed -
+order defines the verified hash ranges, so moving a model-half
+section outside the shape_hash interval, introducing an unknown
+top-level key, or demoting artifact_digest from final position
+all refuse (pinned). The typed component decode carries
+capabilities + adequacy, and fleet claims fail closed over
+admitted degradation: a family a component admits as `degraded`
+cannot certify holds through it - `uncertified`, naming the
+instances and withdrawn flags - scoped like the
+unreachable-unknown rule so unrelated degradation does not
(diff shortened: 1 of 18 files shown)


rileyr:
Round 2 addressed in a1bdc99 — both P1s.

**P1 — one identity grain for routes.** Taken as your "simplest rule", exactly: since a plan endpoint explicitly names a local topic, both operations now read the typed endpoint rows carrying **that topic identity** — the role check via `has_topic_endpoint` (`endpoint.topic == Some(plan_endpoint.topic)`; a publisher role additionally needs a send-site row, preserving "declaring a topic is not using it"), and edge construction via `topic_publishers`/`topic_subscribers`, whose owners come from the same rows the role check admitted. Your counterexample — a component with only a literal send on `"wire.orders"`, never publishing through `Orders` — now refuses at route admission instead of forming an admitted-but-edgeless route that a fleet `forbid_reaches` could vacuously hold over (pinned: `literal_only_publisher_cannot_satisfy_a_topic_route`, which rebuilds the prober as a literal-only publisher and asserts the plan is refused). The reverse failure is structurally gone: the V1-name-grain `publishers_of`/`subscribers_of` lookups are deleted, so a literal end whose *text* is `"Orders"` has nothing left to collide with. The wire-grain `has_endpoint` accessor survives for exactly the consumers you carved out — the fleet claims (`require_*`, count) that deliberately quantify over wire subjects.

**P1 — the declared identity is verified.** New `hale_types::topology::verify_shape_hash`: it extracts the model half from the raw artifact text (the substring between the `shape_hash` header line and the `sources` tail — JSON strings cannot contain raw newlines, so the line-anchored boundary is unambiguous), recomputes the FNV, and compares with the declared value. Both Track A and fleet run it **before law admission or typed decoding**, per your ordering note. Your stronger control is the pin (`stale_shape_hash_is_refused`): the literal wire is rewritten *consistently* in the hashed `endpoint_identity` and every unhashed mirror, only `artifact_digest` is restamped, and the document refuses with "shape_hash does not recompute from the model half" — so fleet can no longer record a stale `app_shape_hash` for a component whose wire identity drifted. The tamper helpers now restamp `shape_hash` alongside `law_digest`, keeping every existing deep pin pointed at its own target rather than the identity gate.

Docs: spec/verification.md (route grain + verified identity), CHANGELOG round-2 entry.

Workspace 2848/2848, corpus differentials green on their charter, zero build warnings.

rileyr:
Round 3 addressed in 732da39 — both P1s.

**P1 — the decoder is strict and fallible.** Every semantic row `ComponentModel::decode` consumes now refuses on malformation instead of filtering: call endpoints in both relations, publish/subscribe fields, topic rows, the unknown's `fn` and **every** reason entry, and every sources row. Your framing was exactly right and I've kept it in the code comment: "already admitted" was no defense, since the law-account admission doesn't strictly decode the call relations or unknowns. The optionality boundary is explicit: provenance, unplaceable (negative-source) decl rows, and a bus-free program's *absent* `endpoint_identity` section are legitimate absences — but a present-and-mistyped section refuses, and a decl row naming a source id outside the sources table now refuses too rather than dropping the location.

Both of your negative controls are pinned — with one honest process note: my first draft of the test skipped silently, because the stock fleet fixtures carry **no** call rows or unknowns at all (I probed the artifacts before trusting the green run). The pinned version rebuilds the oms component to genuinely carry both shapes (`call_it(helper, …)` gives it a `calls` row *and* an `indirect_call` unknown), asserts the needle premise so it can never no-op again, then flips `calls[0].to` to a number (refuses with "relations.calls[0].to must be a string") and `unknowns[0].reasons[0]` to a number (refuses with the reason-indexed message), both under fully restamped digests.

**P1 — one unambiguous value per key.** New `hale_types::topology::find_duplicate_key`: a raw pre-parse scan — string- and escape-aware, every object at every depth — run at both consumption boundaries (Track A and fleet) before `serde_json` ever sees the document. Your exact construction is the pin (`duplicate_shape_hash_is_refused`): a second `shape_hash` appended after the hashed half with `artifact_digest` restamped — where the raw verifier would check the first, correct value and the parsed map would consume the stale one — now refuses with "duplicate object key `shape_hash`". Since the scan covers all keys at all depths, the same shadowing is closed for every other section (a duplicated `endpoint_identity` or `law` can't split the verified and consumed views either).

Docs: spec/verification.md (strict decode + key unambiguity), CHANGELOG round-3 entry.

Workspace 2850/2850, zero build warnings.

rileyr:
Round 4 addressed in the latest commit.

**One structural walk replaces the textual helpers.** `scan_top_level` does a single raw pass — string- and escape-aware — that (a) **decodes key names before duplicate comparison**, so your `"shape_hash"` spelling is the same scanner key as `"shape_hash"` and refuses with the decoded name (pinned: `escaped_duplicate_key_is_refused`); the unescaper covers the full JSON escape set including surrogate pairs, and a malformed escape refuses the document outright. And (b) it returns the top-level entries with byte spans, which both verifiers now use **structurally**:

- `verify_shape_hash` reads the *top-level* `shape_hash` entry and hashes exactly the bytes between it and the top-level `sources` entry — a nested `shape_hash` placed before the real field is data, never the checked value. Pinned in both directions (`nested_shape_hash_decoy_does_not_confuse_the_verifier`): the decoy on an honest artifact doesn't break admission (the verifier reads the top level and the artifact still admits), and the same decoy cannot rescue a consistently drifted model under a stale top-level value — it still refuses on the top-level staleness.
- `verify_artifact_digest` gets the analogous fix you flagged: the digest is the top-level entry, the covered body ends at the preceding top-level entry's terminator (byte-identical to what the emitter hashed), and the `rfind` that a late nested `artifact_digest` could have captured is gone.

Boundary ordering: the structural scan runs *first* at both consumption boundaries, so a duplicate key refuses by name rather than surfacing as a confusing downstream verifier miss.

Docs: spec/verification.md, CHANGELOG round-4 entry.

Workspace 2852/2852, zero build warnings.

rileyr:
Round 5 addressed in the latest commit — both P1s.

**P1 — layout is law.** `verify_top_level_order` runs on the scanned document at both consumption boundaries, right after the duplicate scan: known keys must appear in the emitter's canonical sequence (absences allowed — a bus-free program legitimately has no `endpoint_identity`), unknown top-level keys are refused by name, and `artifact_digest` must be the **final** entry. Your two bypasses are the pins (`top_level_reordering_is_refused`): moving `relations` before `shape_hash` refuses on "out of canonical order" — the edit that would have parked the call graph outside the identity's coverage — and a demoted `artifact_digest` (moved before `verdict`) refuses likewise. Taking the closed-list route also strengthened round 4's decoy result: an unknown top-level key now refuses outright, so that pin was rewritten to assert the by-name refusal.

**P1 — the completeness account is honored.** `ComponentModel` now decodes `capabilities` and `adequacy` (strictly), and a fleet claim of a family that an involved component admits as `degraded` cannot certify `holds`: it reports `uncertified`, naming the instances *and their withdrawn capability flags* (the actionable half of the account). Violations stand — a witness is a witness; only absence-certification fails closed. One design note worth surfacing: my first cut degraded *fleet-wide*, and the existing `fleet_model_soundness` pin caught it — the codebase already settled that "an unknown the claim's source cannot reach is not evidence about that claim," so the adequacy scoping mirrors that rule exactly: a degraded instance counts only if the claim's sources can reach it (under the composed graph) or it hosts a quantified endpoint; count claims, which quantify fleet-wide exposure, treat every degraded-endpoint component as involved. Your minimal control is the pin (`degraded_component_blocks_absence_certification`): an admitted component with `exact_calls: false`, consistently degraded adequacy (withdrawing CALLS degrades every CALLS-consuming family, so the account is flipped coherently), and **no legacy unknowns** — a reachability prohibition routed through it reports `uncertified` naming `oms-0` and `exact_calls`, never `holds`.

Docs: spec/verification.md (layout law + honored account with its scoping), CHANGELOG round-5 entry.

Workspace 2854/2854, zero build warnings.

rileyr:
Round 6 addressed in the latest commit.

**Polarity.** The blanket `holds → uncertified` rule now applies only to `forbid_reaches`/`only_edges`, exactly as you specified. The endpoint and count forms moved their completeness handling inside the evaluators:

- **`require_*` is completeness-independent in both directions.** A known routed witness stays `holds` — and following your "route-backed structural failure remains violated" note through: since route admission only ever admits routes over *known* typed endpoints, and the plan's route table is complete by construction, a hidden endpoint can never be the missing half of a route the plan doesn't carry — so both verdicts are definite and `require_*` is fully exempt from the degradation pass.
- **Counts follow the canonical monotone rule** over a `[known, known+hidden]` interval: known rows are the lower bound; an uncounted component that withdraws the *relevant* completeness (publisher counts consult `exact_publishes` + `exact_cardinality`, subscriber counts the subscribe pair) raises only the upper bound. `min` holds at the lower bound, `max`/`eq` definitely violate once the lower bound exceeds them, the undecided interval is `uncertified`, and the conjunctive eq/min/max form is judged over the interval (any definite violation → violated; all definite passes → holds; else uncertified). The witness names both the counted instances and the hidden candidates.

**Row serialization** — good catch, confirmed: the round-5 post-pass rewrote the pass/fail tracking but `rows.push` ran first, so the serialized row kept the pre-rewrite result. The final verdict is now chosen before the row is pushed.

**All five controls** live in one pin (`completeness_polarity_is_respected`, with `gw-0` withdrawing `exact_publishes` coherently — both flags you'd touch degrade all five families, so the account flips uniformly): the routed `require_subscribes` witness stays `holds` under unrelated degradation; `min: 1` with one known publisher stays `holds`; `min: 2` inside `[1, 2]` is `uncertified`, not `violated`; `max: 0` against a known publisher stays `violated`; and `eq` distinguishes known-too-high (`exactly_zero` → violated) from undecided (`exactly_two` → uncertified).

**Compatibility note** — taken with your "no backwards compat" ruling: Track A now consumes exactly `TOPOLOGY_SCHEMA` (1.12), so the stated contract and the closed layout table agree; the layout table is versioned by the schema it describes.

Docs: spec/verification.md, CHANGELOG round-6 entry.

Workspace 2855/2855, zero build warnings.

Issue #464, opened 2026-08-13
Placement-aware devirtualization: extend closed-world dispatch from towers to thread domains


rileyr:
## The ask

Extend the closed-world dispatch optimization beyond its current same-thread-by-construction scope: use **placement** to prove that a publisher and all subscribers of a subject share one consumer thread, and devirtualize their traffic accordingly. Today we devirt within a locus tower because a tower is trivially same-thread; after placement is known, "same thread" is a computable equivalence class over the whole instance graph.

## What exists today (the three tiers)

1. **Intra-tower AST desugar** (P23, downstream handoff 2026-08-11): a closed-world intra-locus/tower publish rewrites to a direct handler call at the AST level. Decided purely from tree structure — no placement knowledge needed.
2. **Static per-subject buckets** (build #1b): the BusGraph gate in hale-types proves a subject closed-world (no transport binding, no wildcard, no cross-seed, no routing key) → compile-time subject id + `lotus_bus_dispatch_static` bucket walk. Dispatch is still queued; only the lookup is devirtualized. Includes the `flat` and `no_pinned` (`_st` enqueue) fast paths.
3. **Full direct-call devirt** (build #1b slice-2, `lotus_bus_dispatch_static_direct`): the bucket walk becomes synchronous direct calls, gated on (a) every subscriber cooperative on the owner thread — no mailbox, no coop_pool; (b) every handler provably **quiet** — self-only mutation with pure expressions, no calls, no I/O, no republish; (c) payload flat (pointer-free POD). This is the ~1.7ns path.

Tier 3's "same-thread" condition is the narrowest possible one. Placement widens exactly that condition and nothing else.

## The key insight: placement defines single-consumer equivalence classes

The invariant the runtime already rests on — one cooperative pool is **one worker thread**, one pinned tower is one thread, main is one thread (`spec/runtime.md` § Scheduler; the same invariant the existing devirtualization and `spec/testing.md` § Determinism stand on) — means placement partitions instances into **thread domains**: main, pool-k, pinned-instance-i.

Two facts make this compile-time-usable:

- Placement is **source-declared** (`placement { p: pinned(core=0); }`, `cooperative(pool=X)`), so it is in the model, not discovered at runtime. Placement stays semantics-free (#462's rule); this proposal uses it only as a *proof input* for an optimization, never as a semantic knob.
[...]

PR #491, opened 2026-08-21
feat(model): deployment consumers — the arrangement, the dispatch plan, canonical obs ids (GH #476 Change 8)

rileyr:
Change 8 of the canonical-model epic (GH #476): the deployment and optimization consumers. Three commits — the arrangement, the dispatch plan, the manifest entity ids.

## The arrangement (commit 1)

Change 2 left `locus_instances` / `realizes` / `owns` / `placed_in` / `thread_domains` / `bindings` / `binds` empty with a note that no consumer exported them. Their consumers land here, so they are populated: the params-default tree rooted at the main locus, `pinned(replicas = K)` fanned to K instances, thread domains by the runtime's own rule (pinned owns a domain, `cooperative(pool = X)` runs on X's worker, everything else inherits its owner, root is `main`, a binding reader is its own domain).

What the arrangement cannot see stays a typed hole rather than silence: instances born outside it hide `OWNS|PLACED` at the born locus, adapter transports hide `BINDS|DELIVERY` at the topic, and the capability account gains `exact_ownership` / `exact_placement` / `exact_routes`.

That closed a fail-open. The ownership graph walks locus member bodies only — its question is which owner a child bubbles to, and a free function has no owner — so `fn main() { W { }; P { }; }`, the corpus's most common program shape, produced zero modeled instances *and* claimed exact placement over them. `free_fn_birth_sites` closes it; the arrangement root itself is exempt, since `fn main() { App { }; }` is how the arrangement is entered.

## The dispatch plan (commit 2)

`DispatchPlan::derive(&ApplicationModel)` owns the lowering decision — a conclusion derived from the model, never a model row. One row per subject: flavor (`dynamic` / `static_bucket` / `static_direct`), the gate's reason, the subscriber list, publisher and subscriber thread domains, and `same_domain` — #464's stage-0 survey question ("how much queued bus traffic is same-thread-domain?") as a model query instead of a bespoke topology walk. `hale model dump` prints the plan and the survey count.

Codegen no longer decides. It still builds its own gate facts, over the merged and desugared program its emission uses — the source of facts must stay local or the lowering could disagree with itself — but the ladder is `DispatchFlavor::of` in hale-model, and the ids, direct set, and direct-subscriber lists come from `plan.static_subjects()`. Same ids, same order, same lowering.

The plan is part of what a build **is**. `exec_digest` frames `DispatchPlan::digest()`, so two builds of byte-identical sources that lower dispatch differently — notably the `LOTUS_NO_BUS_DEVIRT=1` control arm — cannot share a recording identity, and replay across that boundary is refused by name.

**Defect found by the differential.** A subject named by a `bindings { }` entry is exempt from devirtualization: the adapter's peer is (or may be) the real counterparty. The exemption was keyed only by the topic DECL name, but codegen builds its graph after topic desugaring, where subjects are wire strings — so a bound topic could be lowered into a static bucket its own adapter is not part of. The walk now records both grains. One of the 21 bus-carrying corpus programs disagreed; that is what surfaced it.

## Canonical entity ids for iris (commit 3)

Manifest rows name entities and number them in registration order — a fact about the run, not the program — so a consumer joining a live process to the source-derived model had to match on strings. Now the compiler answers: `hale_model::obs_ids` projects the model to `(manifest kind, name, canonical id)`, the CLI hands the rows to codegen with the model hash and exec digest, the prelude stamps them, and the runtime writes the id into the entry's `aux_b` when the row is lazily created.

`aux_b` has been in the entry layout since v0 and written as 0 by every path, so no consumer's layout moves and no protocol version changes. Ids are `index + 1`, keeping 0 as "unstamped" for harness builds and for entities the model does not name. MK_TOPIC carries the `SubjectId` (the manifest fuses publishers by wire subject, so the address is the identity), MK_LOCUS_TYPE the `LocusDeclId`, MK_BINDING the `BindingId` — indices into the model whose `shape_hash` the header already carries.

## Tests

- 5 plan laws in hale-types: wire grain (a plan keyed by decl name could not be compared to codegen's at all), the ladder, a pooled subscriber is not same-domain, a transport-bound subject stays dynamic *and says why*, unarranged loci forfeit the survey field rather than guess, the digest tracks the plan.
- A corpus differential through the real binaries: `hale model dump` vs `HALE_DISPATCH_TRACE=1 hale build`, requiring agreement on every subject the model names, over every bus-carrying corpus program. The converse is deliberately not asserted — codegen additionally carries stdlib subjects no user model names.
- An identity test: the all-dynamic control arm stamps a different exec digest, and replaying its recording against the default build is refused by name.
- A manifest test that reads the live segment and requires every row's `aux_b` to equal the id `hale model dump` reports, over a program whose birth order deliberately differs from the model's table order — otherwise it could not tell a canonical id from a registration-order one.
- A codegen-side control proving a harness build still publishes all-zero `aux_b`.

Full workspace suite green (2864 tests), zero warnings.

## Docs

`spec/runtime.md` gains the dispatch-plan section (flavors as a derived conclusion, the shared ladder, `same_domain`, the plan in execution identity, the transport-bound exemption at both grains) and the canonical-entity-id paragraph. The book's bus chapter tells authors how to read the plan and that a bound topic keeps the real bus path. CHANGELOG carries the Change-8 entry including the devirt fix.

## What Change 8 does not do

Instance-level canonical ids stay out: the runtime's instance table is birth-ordered and the model's instance ids are static-arrangement ids, so joining them needs the arranged path at the birth site — a separate ask, not a prerequisite for the ID join this delivers. Flavors A and B from #464 (same-domain local queue, widened quiet direct) are unchanged; this ships the stage-0 survey they were waiting on.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,58 @@ behavior.
 
 ### Deployment consumers: the arrangement + the dispatch plan (GH #476 Change 8)
 
+Review round 1 (PR #491), five blockers plus two identity bugs they
+uncovered:
+
+- **Replica rows carry their index, not the count.** `replicas = 3`
+  produced `[3, 3, 3]` where the runtime has replicas 0, 1, 2 — the
+  keyed-delivery model the arrangement exists to feed would have
+  named a population no process has. `validate` now enforces that a
+  replicated field's instances form a contiguous 0-based set and
+  that the index in the path and the index in the row agree.
+- **Binding rows are decoded, not defaulted.** Every unix binding
+  was modeled `role: Listen, loss: WaitCapable` regardless of what
+  was authored, so an explicit (or inferred) `connect` binding was
+  its own opposite in the model. Roles now come from ONE rule shared
+  with the desugar, loss follows the role, and rows are canonically
+  sorted BEFORE ids are assigned — two bindings authored in reverse
+  subject order used to produce a model that failed its own
+  canonical-order law.
+- **A partly dynamic population is not same-domain.** The plan built
+  its domain map from arranged instances only, so a locus with one
+  arranged instance and one dynamic birth answered "main" for the
+  whole population — a false stage-0 optimization opportunity, and
+  an unsafe precondition for the placement-driven lowering it is
+  meant to inform. A placement hole now deletes the locus's answer.
+- **`hale build` artifacts carry an execution identity.** The
+  identity plumbing reached `hale run` and `hale replay` but not the
+  path that produces the binaries users ship, which set a model hash
+  and no digest at all. It now stamps the digest from the FINAL
+  build options plus the plan.
+- **Canonical ids are identity-bound and total.** `model_hash` does
+  not cover every table these ids index (binding rows are not in the
+  artifact), so the header publishes `entity_id_digest` at `0x88`
+  (proto 0.3) over the exact stamped rows: a consumer recomputes it
+  from its own model and joins only on a match. The mapping table
+  grows with the program instead of silently dropping entities past
+  a fixed 512 — a dropped row read back as "unstamped", which is
+  what a build with no ids at all looks like. Out of memory now
+  withholds the whole channel, loudly, rather than half of it.
+
+**Fix (observation identity).** A program with a `bindings { }`
+block published `model_hash 0` for its whole life: registering a
+binding creates the observation segment, segment creation snapshots
+the identity fields, and the prelude stamped them afterwards. The
+identity setters now run before anything that can register. Eager
+recording/replay init deliberately stays after binding realization,
+so a backend with no replay class still refuses at its own seam.
+
+**Fix (binding role inference).** The documented inference
+(publish-only → `connect`, subscribe-only → `listen`) was dead code:
+the desugar ran it after rewriting topic references to literal
+subjects, so it saw no topic ends and filled in nothing, and codegen
+refused every binding that did not spell `role:` explicitly.
+
 The model's arrangement tables are populated and their consumers
 land on them. `locus_instances` / `realizes` / `owns` /
 `placed_in` / `thread_domains` / `bindings` / `binds` now carry
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,64 @@ behavior.
 
 ## Unreleased
 
+### Deployment consumers: the arrangement + the dispatch plan (GH #476 Change 8)
+
+The model's arrangement tables are populated and their consumers
+land on them. `locus_instances` / `realizes` / `owns` /
+`placed_in` / `thread_domains` / `bindings` / `binds` now carry
+the static deployment: the params-default tree rooted at the main
+locus, `pinned(replicas = K)` fanned to K instances, thread domains
+by the runtime's own rule (pinned owns a domain, `cooperative(pool
+= X)` runs on X's worker, everything else inherits its owner, the
+root is `main`, a binding reader is its own domain). What the
+arrangement cannot see is a typed hole, not silence: instances born
+outside it (method bodies AND free functions - `fn main() { W { };
+}` is the corpus's most common shape) hide OWNS|PLACED at the born
+locus, adapter transports hide BINDS|DELIVERY at the topic, and the
+capability account gains `exact_ownership` / `exact_placement` /
+`exact_routes` accordingly. This closed a fail-open: a program whose
+loci are all born in `fn main` used to claim exact placement over
+zero modeled instances.
+
+`DispatchPlan::derive(&ApplicationModel)` owns the lowering
+decision: one row per subject with the flavor (dynamic /
+static_bucket / static_direct), the gate's reason, the subscriber
+list, publisher and subscriber THREAD DOMAINS, and `same_domain` -
+GH #464's stage-0 survey question as a model query rather than a
+bespoke topology walk. `hale model dump` prints the plan and the
+same-domain count. Codegen no longer decides: it still computes its
+own gate facts over the merged, desugared program it emits from,
+but the ladder is `DispatchFlavor::of` in hale-model and its ids /
+direct set / direct-subscriber lists come from the plan. A corpus
+differential runs both fact sources through the real binaries and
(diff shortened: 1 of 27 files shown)


rileyr:
All five fixed at `da1afd7`, each with the control you asked for. Two further identity bugs fell out of the work; both are in this commit.

**1 — replica index.** Instances now store `Some(i)`, and it is a model law rather than builder trust: `validate` requires a replicated field's instances to form a contiguous 0-based set AND the index in the path to equal the index in the row (they were two spellings of one fact, so a `base[2]` claiming replica 0 is now refused too). Controls: a three-replica derivation asserting exactly `[("App.workers[0]", 0), ("App.workers[1]", 1), ("App.workers[2]", 2)]` with non-replica instances claiming no index, plus a negative that stamps the count into every row and requires `ReplicaIndicesNotContiguous`. The pre-existing `architecture.rs` fixture carried `path: "App.w", replica: Some(0)` — an incoherent pair the law now catches — and is corrected to `App.w[0]`.

**2 — binding projection.** Rows are decoded first, canonically sorted second, ids assigned third, and `binds` points at post-sort ids. The role comes from one rule shared with the desugar (`hale_syntax::desugar::binding_role_for`: explicit wins, publish-only is connect, subscribe-only is listen, ambiguous/unused is a diagnostic and yields a hole rather than a guess). Loss follows the role, per what the runtime actually does — connect is the publish side, where a send failure marks the entry lost and `or wait` parks through the reconnect window; listen re-arms on peer EOF and an unservable link is structural. Controls: explicit connect stays Connect with WaitCapable; inferred listen stays Listen with Fail; two bindings authored in reverse subject order validate, sort to `["a.sub", "z.sub"]`, and every `binds` row joins a topic to the binding for its own subject.

Fixing this surfaced that the documented role inference was **dead code**: `desugar_topics` ran `desugar_binding_roles` *after* `rewrite_items` had turned every `BusSubject::Topic` into a literal, so the publish/subscribe sets were always empty and codegen refused any binding without an explicit `role:`. Reordering makes the inference live, which is what the book already claims and what makes the model's rule and the desugar's rule the same rule rather than the same rule with one branch dead.

**3 — mixed populations.** `derive` now subtracts every locus carrying an OWNS/PLACED hole from the domain map, so an incomplete population has no answer rather than a partial one. Control: one arranged `Sub` under `App` plus a second `Sub` born inside a `cooperative(pool = io)` locus, asserting both premises (the arranged instance exists; `exact_placement` is false) and then `subscriber_domains == []`, `same_domain == false`, `same_domain_queued().0 == 0`.

**4 — `hale build` identity.** The build path now derives plan digest and entity ids from one model derivation and stamps `exec_digest` last, from the finalized options; `options_fingerprint` is shared so build and run fingerprint identically. Control: the same source built twice (default, then `LOTUS_NO_BUS_DEVIRT=1`), both run under recording, requiring both digests nonzero and different, and requiring replay across them to refuse on execution identity.

**5 — observation ids.**

*Identity binding.* You are right that `model_hash` cannot carry these ids — I confirmed the binding case directly (adding a `bindings { }` entry leaves `shape_hash` byte-identical). Rather than narrow what is stamped, the ids now carry their own identity: the header publishes `entity_id_digest` at `0x88` (proto 0.3) over the exact `(kind, name, id)` rows, computed in Rust by `hale_model::obs_ids::digest` and carried by the runtime, so there is one implementation. The contract is now "recompute from the model you hold and join only on a match" — a detectable refusal instead of a silent misattribution. Control: two builds where `model_hash` is identical (premise asserted) but a binding row appears, requiring the digests to differ AND each to equal what a consumer recomputes from the derived model.

*Capacity.* The table grows with the program; there is no cap. Allocation failure sets a broken flag that withholds every id and zeroes the digest, with a named stderr line — all-or-nothing, since a partial table is indistinguishable from an unstamped build. Control: a 600-entity program with every registered row stamped. I verified it actually catches the old behavior by temporarily reintroducing a 512 stop: 78 registered locus types came back unstamped.

**Bonus bug.** Chasing (5) turned up that any program with a `bindings { }` block published `model_hash 0` for its entire life: registering a binding creates the observation segment, segment creation snapshots the identity fields, and the prelude stamped them afterwards. The identity setters now run ahead of the bindings prelude and the config loader. Eager recording/replay init deliberately stays *after* binding realization — moving it broke `replay_ingress::shm_ring_refuses_replay_and_touches_no_shared_memory`, correctly: a backend with no replay class must refuse at its own seam naming itself, not be pre-empted by a generic identity refusal. That ordering is now spelled out in the code and in `spec/runtime.md`.

Spec and CHANGELOG updated for all of it (proto 0.3 header field and the join protocol, the role/loss ruling, the replica-index law, the same-domain completeness rule, the stamp-before-registration ordering). Full workspace suite green at 2873 tests, zero warnings.

rileyr:
Fixed at `7f3e4aa`. Your reproducer was exact — I ran it first and got `internal error: derived model violates a model law: ReplicaIndicesNotContiguous { base: "App.workers" }`, so the law was refusing a perfectly ordinary nested arrangement.

The validator now decides replica-ness from the final path component only: it takes the substring after the last `.`, and treats the row as a replica row only when *that* component ends in `[N]`. Brackets in ancestor components describe the ancestor, which is the whole point — the arrangement walks into each replica, and what a replica owns is an ordinary child.

Positive control derives your example and requires all five instances with the leaves carrying no index:

```
("App",                 None),
("App.workers[0]",      Some(0)),
("App.workers[0].leaf", None),
("App.workers[1]",      Some(1)),
("App.workers[1].leaf", None),
```

It also pins the placement consequence, since that is the fact the nesting exists to carry: each leaf lands in its replica parent's own domain (`pinned:App.workers[0]`, `pinned:App.workers[1]`), not on main.

I also strengthened the existing negative rather than only keeping it. It set the count into every row, which trips the path-vs-field agreement clause before contiguity is ever consulted — so the contiguity clause itself was never exercised. It now additionally builds `[0, 1, 7]` with path and field agreeing, and requires the refusal to name `App.workers`.

Spec and CHANGELOG note the last-component rule. Full workspace suite green at 2874 tests, zero warnings.

PR #492, opened 2026-08-22
feat: one authority per question — retire the duplicate evaluators and the legacy artifact arm (GH #476 Change 9)

rileyr:
The last change of the canonical-model epic (GH #476). Three commits: claim verdicts, the artifact's second serialization, and the evaluator's demotion.

## What a duplicate authority was costing

Two of them were live. `hale check` reported reachability, boundary, endpoint and bound verdicts from an evaluator that re-derived them from source, while the artifact reported the same four families from the judgment engines over the canonical model. And `dump_topology_parts` still ran the legacy gathering — relation rows, labels, unknowns, groups, supervision, the whole unhashed provenance tail — beside the projection that actually gets emitted.

A corpus differential can hold two implementations equal. It cannot make them the same answer: the guarantee stops exactly at the edge of the corpus, and every program a user writes is outside it.

## 1 — claim verdicts (commit 1)

`judgment::claim_law_diags` runs the migrated engines over (ClaimIr × ApplicationModel × evidence), and check calls it. The judgment merge had been inline inside `project_law_rows`, which made the artifact its only possible consumer; extracted as `topology_projection::judge_all`, it is the judgment entry point and both consumers call it.

Scope is deliberate. Law SELECTION — which laws exist at all: clause enumeration, constitution adoption, group resolution, the library/world tier rule — stays with the claim surface, because that is a question about the surface rather than about what a law says, and it has one implementation. Certificates are not emitted from the judgment path either: `judge_certificates` judges rows against evidence the effects engine produced, and that engine emits its own diagnostics in check, so re-emitting them would duplicate the thing being deleted. Unmigrated families (`causes`, `depends`, `@budget`) keep their existing single authorities — they are single, not duplicated, and migrating them is Change 5's remaining work, not this one's.

`LoweringIssue` gains a `family` so the two consumers of the table filter coherently instead of pattern-matching message text: the artifact carries every issue, check reports only those no other engine owns.

**The demand rule survives, sharpened.** A claim-free program still never derives the model in check — the LSP's cost contract, and `claim_law_diags` early-outs on a structural AST scan. A program *with* claims now does, because judging it means reading the model. The Change-2 canary is updated to exactly that contract.

**Two defects found while proving parity:**

- `judge_only_edges` never validated grant topics. `only edges A -> B { publish Metrix; }` with no such topic silently dropped the grant and evaluated a **weaker claim than the one written**, reporting its violations as if the author had chosen them. The evaluator refused it as invalid with a did-you-mean hint; the engine now does too. The corpus differential could not have caught this — the fixture covering it is a `format!` template, which the corpus provider skips.
- One documented divergence remains and is now bounded rather than latent: `require attributed` over a body with an indirect or opaque call is `uncertified` where the evaluator fail-open held (Change 5c, recorded by the artifact's `semantics` bump to 2). Check gains that diagnostic, so **a build that passed on the old fail-open now fails**.

The differential holds the two authorities byte-equal — kind, span, message, related notes, order — over every corpus program, with the divergence carved out precisely: the model may only ADD lines, and only `uncertified` ones. It asserts the carve-out is still exercised, so a regression back to fail-open fails the test rather than passing it quietly.

## 2 — the artifact's second serialization (commit 2)

Deleted: −752 lines from `topology.rs`, plus the collectors that fed only it. `dump_topology_parts` returns one String.

That leaves a question the deletion has to answer. With no rival implementation, what stops a model-builder change from silently moving an artifact hash and re-keying `.halerec` replay admission for every existing recording? A committed baseline, same shape as the effects-manifest gate: `topology_shape_baseline.txt` records `origin -> shape_hash` for 1398 corpus programs, and the gate reports the **moved rows** and how to regenerate deliberately. A hash move becomes a decision in a diff.

The corpus test also keeps the self-consistency the emitter still owes with one authority: the artifact it wrote must hash to what the projection says, and its emitted model half must BE the projection. An emitter stamping a hash of something other than what it wrote would still be a silent replay-admission bug.

The per-fixture tests keep their charters (authored-order interface ties, declaration-order labels, supervision ties, raw-order sealed rendering, the full retry literal, exact-renames demangling) and assert on the projected bytes. Two were named for the comparison rather than the property and are renamed to what they pin.

## 3 — the evaluator's demotion (commit 3)

Its last two production callers wanted something it computed *on the way* to judging: the artifact's constitution section and `hale fleet`'s matrix each took the identities and discarded the diagnostics and outcomes, so every artifact dump ran a full claim evaluation for a value it threw away. `constitution_identities` stops at selection.

The evaluator survives as the comparison arm of three corpus differentials — an independent implementation disagreeing is evidence a baseline cannot give. What does not survive is the ambiguity: `legacy_oracle_is_test_only.rs` scans every `src/` file in the workspace and fails if any calls the evaluation entry points, and asserts the oracle still exists so that deleting it later takes the canary with it rather than passing vacuously.

## User-visible behavior changes

Both are fail-closed corrections, both already true of artifacts since Change 6:

1. `require attributed` over an indirect or opaque call is `uncertified` at check time. Builds that passed on the fail-open now fail.
2. An `only edges` grant naming an undeclared topic is an invalid law with a hint, instead of a silently dropped grant.

## Tests

Full workspace suite green at 2876, zero warnings. New: the claim-diagnostics differential (corpus-wide, bounded carve-out), the artifact baseline gate (1398 rows), the oracle-boundary canary. Updated: the demand canary, to the post-cutover contract.

## What Change 9 does not do

It does not migrate `causes` / `depends` / `@budget` — those have one authority each already; they are Change 5's remaining families. It does not rename `LegacyProjection`, whose fields (the V1 fn universe, the stdlib absorption sidecar, the dispatch gates) are load-bearing model data under a name that has stopped describing them; that is a mechanical rename worth doing on its own, not folded into a behavior change.


rileyr:
All three fixed at `b0ab4f6`. Two of them turned out to have more behind them than the symptom.

**1 — selection is now one result with two consumers.** You are right that the split was the real defect, and worse than having two implementations: the checker and the document gave opposite machine-readable answers about one program. `claims::select` returns the universe and every selection diagnostic together; `selection_diags` and `lower_claims` both consume it, so a refusal reaches `law.issues` and the document verdict.

That alone still left the row reading `holds`, which is what you asked me not to accept. So the judgment now refuses to hold over a domain selection rejected: `empty_refused_domain` marks the row `invalid` across every group-quantified family (reachability src and dst, boundary src and dst, endpoint, bound). No diagnostic is emitted there on purpose — selection owns the message, and saying it twice is the duplication this change removes.

`may_be_empty` is excluded deliberately. An author who declared the group may be empty has opted into vacuous truth and the evaluator has always granted it; there is a control asserting that case still holds, so the fix cannot quietly become "empty groups are always invalid".

Your reproducer now yields `verdict: law_failed`, both selection issues in `law.issues`, and `isolation` as `invalid`. Three end-to-end controls drive the real binary: unknown member, unannounced empty group, and the `may_be_empty` counter-case.

**2 — judgment now requires a program that denotes a model.** Moved to `check_bundle_opts`, after the resolver and checker agree. Claim errors deliberately do not gate it: a program whose only failures are broken laws still has a valid model, and hiding its other laws behind one failure would be a fail-open of its own.

The regression test asserts both premises you implied — a non-claim error *and* a genuinely unlawful model (`validate()` is `Err`) — because without the second the test would prove nothing about the hazard. Its mirror proves the gate is not a blanket refusal: fix the type error and the same claim is judged, and a violated claim on a checked program is still reported.

Worth flagging: **the gate found an existing test fixture that had been ill-typed all along.** `model_provenance`'s through-stdlib fixture calls `Router.get`, which does not exist; it only ever "passed" because the old evaluator judged claims regardless of type errors. Fixed to the real `add("GET", …)` API, so that test now proves what its name says.

**3 — claim spans keep their offsets.** I took the lowering fix as the primary one, since it covers every caller rather than every caller I remembered: unplaceable spans now keep their offsets as `ForeignSpan`, exactly as the model builder already did, and `span_of` already reconstructed those correctly. I also installed the source map in the LSP, since it had the bases, paths and text the whole time.

The `check_program` control asserts the claim's own span, not merely non-zero — and I verified it fails under the old synthetic collapse before keeping it. The two-file LSP test initially failed for a reason worth reporting: my first matcher was too loose and matched the *secondary* notes, which correctly live in the other file. The test now pins the primary violation to `b_app.hl` at the claim's line, and the secondary notes landing in `a_domain.hl` is the multi-file placement working.

**On the differentials.** Both gain a bounded `Holds → Invalid` carve-out for refused domains, in the same shape as the existing 5c one. And a fact worth having on the record: **no existing corpus artifact hash moved.** Selection issues live in the unhashed law account, so replay admission is untouched by all of this; the only baseline change is the row for a new fixture.

Full workspace suite green at 2883, zero warnings. Spec and CHANGELOG carry all three, including the two user-visible rules (`invalid` over a refused domain, and judgment requiring a checked program).

rileyr:
Fixed at `a574a90`. You were right about the shape of the mistake, not just the cases: I had the judgment *ask the model* whether selection accepted a group, and "is it empty, and did it declare `may_be_empty`?" is a different question with — as you showed — at least three wrong answers.

Selection now records what it decided, as it decides it. `GroupSelection` has the four states you specified: `Resolved`, `IntentionallyEmpty`, `SelectorFailed`, `Refused`. `ClaimIrTable` carries the map by raw group name, `claims::select` produces it, and the judgment reads `is_judgable()` across every group-quantified family. Only `IntentionallyEmpty` may hold vacuously.

The detection point matters for one of your cases: an unresolvable selector is invisible in the model *by construction*, so there is nothing to observe after the fact. Selection therefore watches whether `resolve_member` reported anything for that member — the one place the failure is known. The vacuity diagnostic itself is unchanged, so evaluator parity stays byte-equal.

All three controls are in, and I verified each **fails under the old member-count guard** before keeping it (I stubbed the guard back in behind an env var and re-ran):

- `{ MissingWorker } may_be_empty` → selection issue + invalid row.
- `{ Worker, MissingWorker }` → invalid, not a verdict over `Worker` alone.
- duplicate declarations → invalid, not a verdict against whichever definition survived the map insert.

The `may_be_empty` positive control still holds vacuously, so the fix cannot drift into "every empty group is refused".

Both differentials' carve-outs now key on the carried status as well — they were keyed on the same member-count guess, which would have kept passing for the wrong reason.

Suite green at 2886, zero warnings. Spec and CHANGELOG record the four states and, explicitly, why the status is carried rather than re-derived.

rileyr:
Fixed at `ec915b2`. Your reproducer reproduced exactly — `holds` before, `invalid` after, and I verified that direction by stubbing the guard back out behind an env var before keeping the control.

`avoiding` is now guarded alongside the endpoints, positioned as you suggested: after the family's validation, so the shape diagnostics still fire, and before roots, mask, or any verdict is derived from a refused group.

**On the pattern rather than the instance.** This is the third round where a refused domain slipped through a position nobody had enumerated, so I stopped fixing operands one at a time. `every_group_operand_position_refuses_a_refused_group` drives all eight group-operand positions across the five families — reaches src/dst/avoiding, only-edges src/dst, bound from, require sealed, require endpoint — each with an unresolvable member in exactly one position, and asserts its own coverage count. Adding a family or an operand fails that test until its case is added, which is the property I actually want after three rounds of the same class.

Two mechanical corrections fell out of building it:

- The endpoint and bound engines refused *before* running the arm, which suppressed validation diagnostics the evaluator emits (`require sealed` over an empty group, for one). A refused domain forbids a **verdict**, not the messages — both engines now run the arm and rewrite the verdict afterwards, collected by ordinal since the arms `continue` from several points. The claim-diagnostics differential caught this, which is what it is for.
- The new fixtures carry an unsubstituted `{PRELUDE}` placeholder so the corpus provider skips them, the same way it skips every `format!`-shaped template. Without it, single-family fixtures land in the family differentials, whose diag filters key on claim name and compare one engine's output against the evaluator's whole stream — a false failure about my own test data.

Suite green at 2888, zero warnings. One added baseline row for a new fixture; **no existing corpus artifact hash moved**, so replay admission is untouched again. Spec now states the rule over operands rather than endpoints, and the CHANGELOG records the round.

PR #493, opened 2026-08-22
refactor(model): dissolve LegacyProjection — delete the copies, name the facts

rileyr:
`LegacyProjection` held two unlike things under one apologetic name: data no other table holds, and copies kept only so serialized bytes would not move. With no compatibility to preserve, the copies go and the facts get an honest home. Net −111 lines.

## Deleted — duplicate authorities inside the model

**`topology_v1_fns`** was the artifact's fn universe, stored beside `Function.summarized`, which is that same set. The builder computed both from one line of summary keys, and the model carried a coverage law whose entire job was asserting the two agreed — a law that exists only because the fact is written twice. It is derived now (`ApplicationModel::summarized_fns`), so there is nothing left to disagree, and the law is gone with the table.

**`topology_v1_calls_via_stdlib`** was the pre-model contraction walk's output — one Boolean, no revisit — carried verbatim because the model's revisit-on-strengthen lattice can set a loop bit that walk left false, and that bit sits inside the hashed model half. The artifact now projects the model's own `ViaStdlib` call rows, merged to endpoint grain with the loop bit OR'd.

Worth stating plainly: **the committed shape-hash baseline did not move.** Over every corpus program the lattice and the old walk produce identical bytes, so the compatibility carve-out was guarding a divergence no real program exhibits. The walk and its shared helper are deleted, along with a dead copy of it still standing in `topology.rs` after Change 9.

## Kept, renamed, documented

`Analyses` carries the bus graph's per-subject `dispatch_gates` (what `DispatchPlan::derive` reads) and `stdlib_absorption` (the stdlib interiors a reachability walk crosses). Neither bridges to anything legacy — they are facts about the program that no other table holds, and they were only in that compartment because the compartment was where analysis products went. `ApplicationModel.legacy` is now `.analyses`.

## Tests

Charters unchanged; two pins now say what they mean. The universe-restriction test flips `summarized` rather than a parallel table — the point it was always making — and the via-stdlib test is retitled to what it proves. One architecture test that asserted the deleted table was sorted and in-range is replaced by one asserting the derived set is the flag's view, in canonical order.

I tried to build a fixture where the lattice and the old walk actually disagree on a loop bit, to pin the behavior change directly. I could not produce one, which is consistent with the baseline not moving; the honest statement is the corpus-wide one above rather than a fabricated case.

Full suite green at 2888, zero warnings.


rileyr:
Taken, with a correction to the diagnosis that does not change what you asked for.

**The transition is in: schema 1.12 → 1.13**, at `997dc82`. You are right that this is a change of interpretation shipping under an unchanged schema and an unchanged hash-kind name, and that a corpus baseline cannot establish compatibility for a class the corpus does not contain. I took your second option: adopt the model's reading explicitly, with the schema bump, the spec rule, and the re-record note the 1.12 transition set the precedent for.

**Your reproducer does not actually diverge, and the reason is worth having on the record.** I built `main` in a worktree and ran both binaries against your `Gate::probe`:

```
main:   schema 1.12  shape 16638d003fd75bff  [{"from":"Gate::probe","to":"Hello::handle","loop":true}]
branch: schema 1.12  shape 16638d003fd75bff  [{"from":"Gate::probe","to":"Hello::handle","loop":true}]
```

The old walk's `seen` is per-caller, so your second (looped) entry to `Router::dispatch` is discarded exactly as you describe — but the contracted row's bit is set anyway, by the router's *own* entry loop when it re-emerges into `Hello::handle`. The caller-side distinction is masked. That holds for every callback today's stdlib exposes (`Router::dispatch` and the middleware chain both iterate; `__http_handle_one_conn` reaches `h.handle` loop-free but is private and only reachable through `Server`'s accept loop), which is why no committed baseline hash moved and why I could not construct a divergence when I tried.

That is a fact about today's stdlib, not a guarantee — a loop-free public callback would expose the class immediately. So the schema bumps regardless. `schema` sits outside the hashed range, so the declaration moves no hash by itself; a program in the distinguishing class takes a new `shape_hash` and one re-record, exactly as 1.12 did for bus programs.

**Control:** `the_via_stdlib_loop_bit_is_the_models` uses your fixture shape, pins the bit both in `relations.calls` and in the emitted artifact, and records why this variant does not diverge — so if a stdlib callback ever becomes loop-free, the test's premise changes rather than passing silently.

Goldens and the shape-hash baseline are regenerated; the only diffs are the schema label in the mermaid/SVG headers and the new fixture's baseline row. Suite green at 2889, zero warnings.

PR #494, opened 2026-08-22
feat(claims): the causes: judgment over the model (GH #476 Change 5f — engine + differential)

rileyr:
Stage 1 of migrating the three remaining claim families and retiring `claims.rs`. This lands the `causes:` **engine and its proof**, not the cutover — see "Why the cutover is not here".

## The engine

`judge_causes` computes the causal surface from model rows: transitive publish sites over `calls`, each subject's subscribers from `subscribes`, and what those handlers perform from `Function.effects`. It subtracts the fn's own direct effects, because a fn's own behavior is the `none:` form's business — subtracting it is what makes this the *causal* surface rather than a restatement of the fn's effects. A class that resolves to nothing (a cyclic definition) makes the contract vacuous, so the row is `Invalid` before evaluation, matching the rule the certificate family already applies.

## The proof

A corpus differential holds it byte-equal to `frontier::causes_diags` — message, span, order — on every corpus program carrying the family. **Zero divergences.**

One divergence is anticipated and bounded in the differential rather than waited for: the evaluator infers effects from the **user-only** summary, so a subscriber whose `syscall` arrives through a stdlib call reads as pure to it and its causal contribution disappears. The model's effect sets are stdlib-merged, so the class is seen. That is the fail-closed direction, and an undeclared causal effect is precisely what this law exists to catch — not seeing it through a stdlib body was an accident of which summary the old engine happened to hold. The differential permits the model to *add* a violation and nothing else.

## Why the cutover is not here

I implemented it, ran it, and backed it out, because it surfaced something that deserves its own decision.

Promoting `causes` out of `JudgmentFamily::Unmigrated` means the row stops importing an old-engine verdict, which means it stops carrying a `law.legacy` report entry. That entry is what `causes_operand_swap_is_refused` relies on: swap one declared effect class for another declared one in the typed law payload, restamp the digests, and the legacy fingerprint no longer re-renders, so admission refuses. Remove the import and that tamper is admitted.

The uncomfortable part is that this is not special to `causes`. That defense exists *because* the row imports an outside verdict, so every already-migrated family looks like it has the same gap for `holds` rows — which are unbacked by evidence across all of them, by design (Change 6's rule is that non-holds verdicts retain evidence). I did not finish proving that on a reachability row, and I would rather say so than assert it.

Three possible answers, none of which should be bolted onto a migration commit:

1. give `holds` rows evidence for the families where a payload operand is otherwise unanchored;
2. recompute the operand at admission against the artifact's own tables (catches substitution of a *declared* class, which the catalog rule does not);
3. accept a uniform posture — `law_digest` plus catalogs plus evidence-for-non-holds — and retire the family-specific fingerprint knowingly, updating the test to pin what actually defends migrated rows.

## Remaining stages

- **5f cutover** — after the question above is settled.
- **5g `depends:`** — same shape; the backward dual.
- **5h `@budget`** — alloc + quantitative, over the model's `weights`.
- **Delete `claims.rs`** — note this is larger than "delete the evaluator": the module still owns law SELECTION (clause enumeration, constitution adoption and identities, group resolution) which check and lowering both consume, plus vocabulary helpers the model builder calls. Those move; the evaluator and the differentials that use it as an oracle go.

Full suite green at 2890, zero warnings.

### DIFF
--- a/crates/hale-model/src/entity.rs
+++ b/crates/hale-model/src/entity.rs
@@ -45,7 +45,30 @@ pub struct Function {
     /// DERIVED effect classes in declaration order (the stdlib-merged
     /// transitive walk; order is semantic in the existing artifact
     /// and is preserved).
+    ///
+    /// RENDERED, and therefore lossy: when the walk reaches
+    /// something it cannot name, this collapses to the single token
+    /// `unclassified` and every simultaneously-known class is
+    /// discarded. That is fine for the artifact, which reports what
+    /// the old engine reported; it is NOT a basis for judgment. Use
+    /// [`Function::effect_lower_bound`] and
+    /// [`Function::effects_unknown`] for that.
     pub effects: Vec<String>,
+    /// The classes this function is KNOWN to reach, whether or not
+    /// the walk also hit something unnameable — the sound lower
+    /// bound, in canonical order (GH #476 Change 5f review).
+    ///
+    /// Judging over `effects` alone cannot express "definitely
+    /// syscall, and possibly more": a handler that performs a
+    /// syscall AND makes an indirect call renders as exactly
+    /// `unclassified`, so a law that a known effect already
+    /// violates reads as merely uncertain. The pair is the model's
+    /// job to keep, not the judgment's to reconstruct.
+    pub effect_lower_bound: Vec<String>,
+    /// The derived walk reached something it could not name, so
+    /// [`Function::effect_lower_bound`] is a lower bound and not the
+    /// whole answer.
+    pub effects_unknown: bool,
     /// DIRECT effect classes, sorted — what this body itself
     /// performs (carriers, own alloc/publish/spawn sites, ffi and
     /// classified stdlib callees), before propagation. The
--- a/crates/hale-model/src/entity.rs
+++ b/crates/hale-model/src/entity.rs
@@ -258,6 +258,16 @@ pub struct Phase {
 #[derive(Clone, PartialEq, Eq, Debug)]
 pub struct EffectClassDecl {
     pub name: String,
+    /// The class's position in the seed's declaration order.
+    ///
+    /// The table is name-sorted (canonical identity), but rendering
+    /// a set of classes is order-SENSITIVE: the language renders
+    /// built-ins in a fixed order and user classes in the order they
+    /// were declared, and a diagnostic naming two of them is
+    /// byte-compared. Sorting the table threw that fact away, so it
+    /// is carried explicitly rather than reconstructed from a side
+    /// channel (GH #476 Change 5f).
+    pub declaration_index: u32,
     /// `effect NAME;` exists (false = bare reference only).
     pub declared: bool,
     pub definition: EffectClassDefinition,
--- a/crates/hale-model/tests/architecture.rs
+++ b/crates/hale-model/tests/architecture.rs
@@ -65,6 +65,8 @@ fn tiny_model() -> ApplicationModel {
             effect_classes: Vec::new(),
             functions: vec![
                 Function {
+                    effect_lower_bound: Vec::new(),
+                    effects_unknown: false,
                     analyzed: true,
                     summarized: true,
                     owner: Some(LocusDeclId(0)),
@@ -79,6 +81,8 @@ fn tiny_model() -> ApplicationModel {
                     provenance: p,
                 },
                 Function {
+                    effect_lower_bound: Vec::new(),
+                    effects_unknown: false,
                     analyzed: true,
                     summarized: true,
                     owner: Some(LocusDeclId(1)),
--- a/crates/hale-types/src/frontier.rs
+++ b/crates/hale-types/src/frontier.rs
@@ -356,12 +356,49 @@ fn collect_published_subjects(
     }
 }
 
+/// One `causes:` assertion's outcome, keyed so a caller can join it
+/// to the lowered law row it came from.
+///
+/// A function may carry SEVERAL `causes:` clauses, and every one of
+/// them anchors its diagnostic at the same fn-name span — so a span
+/// is not an identity (GH #476 Change 5f, review round 4). The
+/// ordinal is the assertion's position among that function's
+/// `causes:` clauses, in source order, which is exactly the order
+/// the lowering emits rows in.
+pub struct CausesReport {
+    /// The annotated function, as the evaluator displays it.
+    pub function: String,
+    /// 0-based position among this function's `causes:` clauses.
+    pub ordinal: usize,
+    /// `None` when the assertion holds.
+    pub diag: Option<Diag>,
+}
+
+/// Per-assertion `causes:` outcomes — the oracle a differential
+/// joins one law to one law by.
+pub fn causes_reports(
+    programs: &[&Program],
+    graph: &BusGraph,
+) -> Vec<CausesReport> {
+    causes_inner(programs, graph)
+}
+
 /// `@effects(causes: {…})` — check the declared causal set against
 /// what the fn can actually cause through bus edges.
 pub fn causes_diags(
     programs: &[&Program],
     graph: &BusGraph,
 ) -> Vec<Diag> {
+    causes_inner(programs, graph)
+        .into_iter()
+        .filter_map(|r| r.diag)
+        .collect()
+}
+
+fn causes_inner(
+    programs: &[&Program],
+    graph: &BusGraph,
+) -> Vec<CausesReport> {
     // The seed's user effect-class table, so an excess class renders
     // as `money` rather than as nothing at all.
     let names: Vec<String> = programs
@@ -370,13 +407,24 @@ pub fn causes_diags(
         .find(|n| !n.is_empty())
         .cloned()
         .unwrap_or_default();
-    let mut roots: Vec<(FnKey, Vec<EffectClass>, Span)> = Vec::new();
+    // (fn, declared classes, span, per-fn assertion ordinal). A fn
+    // may carry several `causes:` clauses; the ordinal is what makes
+    // each one identifiable, since they share a span.
+    let mut roots: Vec<(FnKey, Vec<EffectClass>, Span, usize)> =
+        Vec::new();
     for p in programs {
         for item in &p.items {
             let mut push = |key: FnKey, fd: &FnDecl| {
+                let mut ordinal = 0usize;
                 for a in &fd.effects {
                     if let EffectAssert::Causes(cs) = a {
-                        roots.push((key.clone(), cs.clone(), fd.name.span));
+                        roots.push((
+                            key.clone(),
+                            cs.clone(),
+                            fd.name.span,
+                            ordinal,
+                        ));
+                        ordinal += 1;
                     }
                 }
             };
@@ -419,8 +467,8 @@ pub fn causes_diags(
         .find(|d| !d.is_empty())
         .cloned()
         .unwrap_or_default();
-    let mut diags = Vec::new();
-    for (key, declared, span) in &roots {
+    let mut reports: Vec<CausesReport> = Vec::new();
+    for (key, declared, span, ordinal) in &roots {
         let (actual, via) = causal_effects(&summary, graph, key, &ffi);
         let mut allowed = EffectSet::PURE;
         for c in declared {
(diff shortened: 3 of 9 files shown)


rileyr:
All four accepted — the engine is unsound as written, and I'd rather you not spend another review on it in this shape. Moving to draft.

Taking them in the order that matters:

**1 — incompleteness must survive.** This is the real defect: I wrote `filter(|c| c != "unclassified")` and a `continue` on unclassified handlers, which converts "unknown" into "nothing". Your fixture is exactly right, and the same hole exists on three more channels I never consulted: computed-subject publish holes, `analyses.stdlib_absorption` events (both interior `Publish` and `PublishHole`/`Truncated`), and call holes hit while collecting transitive publish sites. The fix is a two-part accumulator — a *known* causal set and an *uncertain* flag — with the verdict rule: known excess ⇒ `Violated`, else relevant uncertainty ⇒ `Uncertified`, else `Holds`. Uncertainty has to be tracked on both halves (discovering possible publishes, and each possible handler's effect set), as you say.

**2 — the join must be the model's wire identity.** Comparing rendered text was me inventing a fourth endpoint projection. `SubjectId` is the delivery identity; `declared_topic` is a link, not the key. Text equality also misses wildcard coverage, keyed delivery (`key_domain` × `key_predicate`), absorption publishes, and second bus hops. This wants one shared possible-delivery query over the typed rows that `judge_causes` consumes rather than reimplements — and it is the largest of the four, since 5a's own subscriber index keys by topic name and would want to move onto the same query.

**3 — set subtraction cannot express provenance.** Correct, and your example is unanswerable: one `syscall` member, subtracted once, erases the downstream occurrence. `caused` must accumulate only from possibly-delivered-to handlers, never seeded with the root's effects, with `excess = caused − allowed`. Noting your point that the legacy evaluator has the same bug — that makes it a documented fail-closed correction, not something to copy into the canonical authority.

**4 — the differential blesses too much.** Also correct, and worse than "loose": it compares only rendered diagnostics, never `Judged.verdict`, and `old_only.is_empty() && !new.is_empty()` will bless any new-only output for any reason. Your `ffi` example is the concrete one — `class_mask` folds `Ffi` into the `syscall` bit, while my `class_atoms` treats every built-in as an atom named by its spelling, so `@effects(causes: {ffi})` gets falsely rejected and the carve-out swallows it. The rendering-order defect is real too: the evaluator renders in fixed class order and I collect from a `BTreeSet`.

One thing that fix needs and the model does not currently carry: `effect_classes` rows are **name-sorted** in the builder, so declaration order — which the evaluator's rendering uses for user classes — is not recoverable. A two-user-class excess cannot render in evaluator order without adding a declaration index to the model. I'd add it as a typed field rather than reconstruct it, on the same principle as the group-selection status.

Plan for the rewrite: shared typed delivery query first (it constrains everything else), then the two-part accumulator with uncertainty from holes and absorption, then `ffi`/composed-class normalization through one shared mask helper, then the differential comparing verdicts with a premise-identified carve-out, plus the negative controls per relation you listed (calls, publishes, delivery, effects, absorption, holes).

Main is unaffected — nothing here is merged.

rileyr:
Rewritten at `324ed3e` — all four, replaced rather than patched. Your three fixtures are in verbatim as controls, and each one failed before the corresponding fix.

**1 — incompleteness survives.** Known set plus an uncertainty flag, fed by every channel I had ignored: holes hiding CALLS/PUBLISHES at a reachable fn, delivery holes at a topic or subject, stdlib interiors (`PublishHole`, `CallHole`, `Truncated`, and an interior subject with no model row), and an unclassified handler on a delivered path. Known excess ⇒ `Violated`; otherwise relevant uncertainty ⇒ `Uncertified`; only a complete clean walk ⇒ `Holds`.

**2 — the join is `SubjectId`.** `possibly_delivers` joins on the typed id and widens for wildcards; interior stdlib publishes resolve to the same id. Key filters are deliberately not consulted — a predicate can only remove deliveries, so ignoring it over-approximates reach, which is the sound direction here; that's stated in the doc comment rather than left implicit. The witness path still renders the authored spelling (`Orders`, not `orders`): join by identity, report by what the author wrote.

**3 — accumulated, never subtracted.** `caused` holds only what downstream handlers do. I also made the walk follow further hops, since "cause ANYWHERE in the system" plainly means the closure and a one-level walk fails open on a relay — that's a second control.

**4 — the differential is strict.** Verdicts compared, byte-equal messages required at equal strength, no weakening admitted, and strengthening allowed only against a premise computed from the model. `ffi` folds into the `syscall` bit through one shared `effect_class_atoms`, so your false-rejection case is now a passing control.

**Two things the strict differential found on its first run**, which is the best argument for your blocker #4:

- The path spelling regression from #2 — I had the witness printing the wire while the evaluator printed the topic name.
- The evaluator does not render the word "unclassified" at all: it **saturates**, unioning the unclassified bit as every class and asserting `can transitively cause syscall, block, time, entropy, env, secret_use`. So my first attempt to detect that case by message text was wrong, and the rule now keys off the model premise, as you asked. Worth flagging as a semantic decision rather than a mechanical one: the evaluator makes a definite claim about classes nobody observed, and the model answers `Uncertified` — the vocabulary Change 6 settled on for exactly that evidence. Neither certifies; they differ in what they will assert. The differential treats that as a permitted divergence only when an unclassified body actually sits on a delivered-to path.

**One model addition:** `effect_classes` rows are name-sorted for identity, so user-class declaration order — which the message renders by — was unrecoverable. It is carried now as a typed `declaration_index` rather than reconstructed from a side channel.

Suite green at 2899, zero warnings. Baseline diff is six added rows for the new fixtures; nothing moved. Still a draft, and the cutover is still out of scope pending the `law.legacy` tamper-defense question.

rileyr:
All three at `5f5ec0d`. The first needed a model change; the other two were the engine asking the wrong question.

**1 — the lower bound is now a modeled fact.** You were right that interpreting the rendered token as a pair cannot work, and it's worse than lossy: `UNCLASSIFIED` is `u64::MAX`, saturation rather than a bit, so the known classes are gone at the mask level too — there was nothing to mask out. The model carries `Function::effect_lower_bound` + `Function::effects_unknown`, computed by `infer_effects_lower_bound`: the same walk with the same union rules, except an unnameable edge sets a flag instead of swallowing the set. `effects` keeps its rendered form for the artifact, documented as unfit for judgment.

Your mixed fixture is in as a control, with its anti-control: known syscall + indirect call ⇒ `Violated`; the same shape without the known excess ⇒ `Uncertified`.

**2 — uncertainty is scoped to the endpoint.** The global hole scan was indefensible, and your adapter example is now a passing control. The check is scoped to the traversed subject/topic and asks `SUBSCRIBES`/`KEY_FILTERS` — is the possible handler set complete — rather than `DELIVERY`, which as you say is the must-deliver family answering a different question.

**3 — keyed delivery is decided.** `may_deliver` rules an edge out when an exact domain cannot meet a literal filter, and widens on everything unknown.

**Two things worth flagging, because they change what your examples prove:**

Your keyed fixture assumed the builder infers `Exact([1])` from `Msg { shard: 1 }`. It records `AnyOfType(Int)` — the literal is not read into a domain — so that program's sound answer is the conservative edge, and I've pinned it that way as your "unknown domain still creates the possible edge" control. The disjointness rule is pinned directly on `may_deliver`, where a lawful exact domain can be constructed, rather than through a program that cannot currently reach that state. If you'd rather the builder infer exact domains from literal sends, that's a separate model change and I'd want it argued on its own.

And the differential's byte-parity rule needed the saturation case: when the evaluator saturates it lists every class at once while the model names only what it knows, so at equal strength they're compared as a **subset** — the model may name fewer classes, never one the evaluator did not. That check caught nothing, which is the point of writing it as a subset rather than an exemption.

On your point that the premises are whole-program: they are, and I've left them that way for now rather than half-fix it. Tying each authorization to the specific row's causal closure means computing that closure inside the differential, which is re-implementing the engine in the test. The honest version is for the engine to return its witness — the reached handlers and the holes it counted — and for the differential to check premises against that. That's a change to `Judged` and I'd rather do it deliberately than bolt it on.

Suite green at 2905, zero warnings; one added baseline row, nothing moved. Still draft, cutover still out of scope.

rileyr:
All three at `4e78c92`.

**1 — publish sites survive.** The dedup key was `(subject, spelling)`, so the row carrying the delivering key domain could be the one dropped. Collapsing now happens on handler facts after every site is evaluated. Your control is in, with the exact domains set on the derived model — the builder records `AnyOfType` for a literal send, so that is the only way to reach the shape, and `validate` accepts it.

**2 — the residue is wider and properly typed.** You are right that omitting DELIVERY failed open on the case that matters most: a binding on the *published* topic means the publish leaves the program and the peer's behaviour is not in the graph at all. DELIVERY is included, subject-grained residue matches by pattern so a hole on `orders.**` covers a publish on `orders.created`, and the topic identity comes from the publish row instead of `position(|t| t.subject == subject)` — which, as you say, merely picks the first topic sharing an address.

**3 — one law against one law.** Both sides anchor at the annotated fn's span, which is also the row's provenance, so rows and evaluator outcomes join on it. Premises come from the row's own traversal: `judge_causes_witnessed` returns a `CausesWitness` and the test reads premises off it, rather than recomputing the closure — which would be re-implementing the engine inside its own test.

**The per-row differential earned itself immediately.** On its first run it failed with an `Uncertified` carrying an *empty* witness: the discovery-side hole set the uncertainty flag without recording why. A whole-program premise list cannot see that class of gap — it would have found some unrelated hole and authorized the result. That is the strongest argument for your blocker, and it caught my own bug rather than a hypothetical.

Two notes on what I did not do. The witness is a `causes`-specific type rather than a field on `Judged`; when 5g and 5h need the same treatment I expect it to generalize, but inventing the general shape from one instance seemed worse than waiting for the second. And the anti-controls you listed for "unrelated stdlib interior / second hop authorizes nothing" are implicitly covered now — premises are read off the row's traversal, so an unrelated fact cannot appear — but there is no test that would catch a regression back to whole-program premises. That is worth adding and I would rather flag it than claim coverage I have not written.

Suite green at 2907, zero warnings; one added baseline row, nothing moved. Still draft, cutover still out of scope.

rileyr:
Both fixed at `458f065`.

**1 — typed outbound routes end the closure.** You're right that scanning `holes` only catches the case where the model *doesn't* understand the transport. A `unix(..., role: connect)` route is fully modeled — typed `Binding` + `TopicBinding`, no hole — and the publish still leaves the application. The endpoint query reads `relations.binds` and `entities.bindings` alongside the holes now, and a Connect-role route on the published topic makes the endpoint incomplete. Your fixture is the control, mirroring the unrelated-binding one that must stay `Holds`.

**2 — the join key is (function, assertion ordinal).** The oracle reports per assertion: `causes_reports` returns the function, the clause's position among that function's `causes:` clauses, and its outcome. `causes_diags` is now a wrapper over it, so nothing in the production path moved. Your two-clause fixture is a control, and it asserts the shared span explicitly — otherwise the fixture could stop testing the thing it exists for without anyone noticing.

**Making the join honest surfaced a third fact.** The evaluator's root walk doesn't recurse into modules, so a module-scoped `causes:` annotation yields *no* outcome from it at all — which is the documented Change-6 rule that unmigrated rows bridge to the old engines only where the old walk demonstrably enumerated them. My first cut treated a missing outcome as a bad join key and failed; the differential now skips those rows explicitly and counts them, with a control pinning the reason (the evaluator reports nothing; the model lowers and judges the row). Reading that silence as `holds` would have let anything through on module-scoped code.

Suite green at 2910, zero warnings; one added baseline row, nothing moved.

On the anti-controls for whole-program premise regression: still not written, still flagged. Given four rounds of this file, I'd rather add them when the witness generalizes across 5g/5h than write a bespoke one now — but it's a real gap and I'm not claiming otherwise.

rileyr:
Fixed at `6083187`, and this one is worth naming for what it is: the same defect you caught two rounds ago for the delivery join, left behind in the route query. I fixed wire-vs-syntax in one place and not the other.

Both now join on the subject:

- an outbound route is a Connect-role binding whose **subject** is the published wire, regardless of the send's spelling or which topic the binding names;
- a topic-anchored hole is relevant when that topic **addresses** the published wire.

Your literal-publish fixture is in, plus its adapter twin, and I verified both fail against the declaration-link key before the fix. The unrelated-binding control still holds.

You also predicted a consequence I hadn't traced: the absorbed stdlib publishes. Those synthetic rows carry `declared_topic: None` by construction, so under the old key they could never match a topic-anchored route or hole — an interior publish into a bound wire was silently unreachable by the residue check. The subject join fixes that class too, though I haven't written a fixture that reaches an interior publish into a bound topic, so I'm claiming it as a consequence of the change rather than as tested behaviour.

Suite green at 2912, zero warnings; two added baseline rows, nothing moved.

Issue #525, opened 2026-09-04
DNA Track 0: unblockers — duplicate-accept diagnostic, perspective ctor override, aux_b resolution, iris hygiene (#521)


rileyr:
Part of #521 (Hale DNA). This is Track 0: four small unblockers that every later track leans on. Each ships on its own, none depends on the others, and none changes the DNA domain design.

## 1. Reject a second `accept` clause on one locus

`accept_param` is a singular `Option<(String, Ty)>` (`crates/hale-types/src/symbol.rs:190`), and each `LifecycleKind::Accept` member in `crates/hale-types/src/resolve.rs:879` assigns it, so a locus with two `accept` declarations silently keeps the last one. Codegen mirrors the singular shape (`crates/hale-codegen/src/locus/decl.rs:1233`, `instantiation.rs:2926`). There is no diagnostic anywhere in `crates/`.

Spec is explicit that single-accept-type per parent is the v1 design (`spec/types.md:255`). Until that changes, a second clause must be an error naming both spans, not a silent overwrite. The DNA spike (Track A) will hit this on the first fixture, and it must fail loudly there.

- Error diagnostic in resolve with both spans.
- Rust test in `hale-types` asserting the message.
- One sentence in `spec/semantics.md` under `accept`.

## 2. Perspective designation at a construction site

Comment 1 on #521 claims a containing constructor cannot override a `perspective(P)` field with a serving locus. Confirmed at the typechecker. The param-default path (`crates/hale-types/src/check.rs:8737`) has a `serves`-aware `conforms` escape; the literal-override path in `check_literal_fields` (`check.rs:13398`) only tests `TopSymbol::Interface`, so `App { gw: Gateway { router: RouterV2 { } } }` fails with:

```
locus `Gateway`: field `router` expects `Router`, got `RouterV2`
```

Codegen already resolves the override at `crates/hale-codegen/src/locus/instantiation.rs:2093` and the designation branch at `:2211` fires for whichever source produced the value. The fix is one `serves` branch in `check_literal_fields` mirroring `check.rs:8748`.

Design note that has to land with it: the slot is program-global and 1-1 (`spec/semantics.md:2855`). An override at a construction site designates the whole program's slot, exactly as a default does. Two holders designating different impls is last-wins today with defaults too. Document, do not invent a new rule.

- Checker branch plus fixture `crates/hale-codegen/tests/fixtures/examples/65-perspective-ctor-override/`.
- `spec/semantics.md` designation paragraph names both sites.
[...]

Issue #521, opened 2026-09-04
RFC Hale DNA: generic application harness for organism-native development, operations, and governed self-evolution


rileyr:
## Summary

This issue proposes **Hale DNA**: a Hale library plus a first-class `hale dna` toolchain surface for building Hale systems that can observe themselves and their environment, turn pressure into recursive work, propose changes to their own source/process/deployment, collect evidence, route those changes through explicit authority and human review, apply accepted mutations, and observe the result.

The intended product is not a hosted agent-orchestration SaaS rewritten in Hale. It is repository- and organism-native:

- the DNA domain is implemented in Hale and imported into a new or existing Hale project;
- the application/fleet remains the thing that is alive and acting;
- `hale dna` is the local toolchain/control surface through which humans introduce intent and govern it;
- Iris becomes a first-class part of the `hale` toolchain, joining the checked semantic model to the running system and DNA history;
- source remains truth: there is no proprietary workflow/organism graph which later translates into Hale.

The shortest formulation is:

> **A DNA-enabled Hale project is a governed software organism whose genome, expression, experience, work, and evolution are all readable through the same recursive hypergraph.**

“Organism” and “sentience” are operational language here, not a metaphysical claim. The concrete bar is that the system can sense, remember, model its own state, recognize pressure, form and compare perspectives, propose changes to itself, act within bounded authority, ask a superior for judgment, and measure what happened afterward.

This is an umbrella design issue. It should split into smaller implementation issues only after the domain shape is accepted.

## Why this belongs in Hale

Hale already has nearly every substrate required for this, and they are unusually composable:

- the recursive hypergraph and locus ownership give the organism its actual structure;
- vertical-only flow gives delegation, escalation, and supervision one direction at every depth (`H9`, `I4`);
- types carry facts without inventing a second structural primitive;
- interfaces and perspectives provide structural contracts and replaceable implementations;
- `accept` / `release`, lifecycle, capacity, displacement, and failure-up provide bounded recursive work;
- the typed bus carries facts across real ownership, process, placement, and binary boundaries;
- claims and constitutions express non-negotiable law over the checked program;
- `ApplicationModel` / `FleetModel` provide one typed, provenance-bearing compiler account of the checked system (#476);
- topology artifacts, canonical observation entity IDs, and the Iris protocol join that account to running locus instances;
- record/replay provides unusually strong execution evidence (#296, #519);
[...]

PR #530, opened 2026-09-04
fix(types): reject a second accept clause on one locus instead of silently overwriting the first

rileyr:
Track 0 item 1 of #525 (DNA epic #521).

A locus accepts exactly one child type (`spec/types.md` F.11, single-accept-type per parent). `resolve.rs` stored that type in a singular `Option` and every `LifecycleKind::Accept` member assigned it, so a second `accept(c: U)` clause replaced the first with no diagnostic. A parent written to own two child types quietly owned one, and the child literal inside it still found an owner, so nothing downstream noticed either.

It is now a typecheck error naming both clauses (the second carries the first as a related span), and the first clause stays the locus's accept type so the rest of the check sees the same program it always did.

Surfaced by the DNA Phase 0 design: its first fixture needs a Step to own both Work and delegated Tasks, and that has to fail loudly rather than compile into something else.

- `crates/hale-types/src/resolve.rs`: track the first clause's span, diagnose the second.
- `crates/hale-types/tests/duplicate_accept.rs`: two clauses error with related span pointing at the earlier one; one clause is clean.
- `spec/semantics.md` `accept(c)` section, `docs/src/services/parents-children.md`, CHANGELOG.

Verified: `cargo nextest run --release -p hale-types` (1159 passed).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,20 @@ behavior.
 
 ---
 
+## Unreleased
+
+### A second `accept` clause is an error, not a silent overwrite (GH #525)
+
+A locus accepts exactly one child type (`spec/types.md` F.11,
+single-accept-type per parent). The checker stored that type in a
+singular slot and a second `accept(c: U)` clause simply replaced
+the first — no diagnostic, and the parent written to own two child
+types quietly owned one. It is now a typecheck error naming both
+clauses (the second carries the first as a related span), and the
+first clause stays the locus's accept type. Surfaced by the DNA
+Phase 0 design (#521), whose first fixture needs a Step to own
+both Work and delegated Tasks and must fail loudly there.
+
 ## v0.19.2 — helpers get their speed back (2026-09-04)
 
 ### The caller-arena publish is gated on allocation, not on syntax (GH #522)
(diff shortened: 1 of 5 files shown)


rileyr (commented):
Reviewed commit `51d5c43d0cf1d6d0a35cea4b97ca8a7dcea7af3a` against `2a36bfb3535c8b81fd5b5ae6e997bd99398d524f`.

No correctness findings or merge-blocking concerns. The resolver rejects each later `accept` as a type error, attaches the first clause's span, and preserves the first resolved child type. The guard is local to each locus and is reached for module/imported declarations. The diagnostic prevents canonical model derivation for the invalid program. The spec and docs match the existing single-accept rule.

Validation in an isolated checkout of this commit:

- `cargo nextest run --offline --release -p hale-types`: 1,159 passed.
- Four additional scratch regression checks passed: repeated same-type clauses produce errors at the exact primary/related locations; the resolver preserves the first child type and parameter name; separate one-accept parents remain error-free; a duplicate in a second file preserves both shifted spans.
- `git diff --check` passed.

Nonblocking test suggestion: `one_accept_clause_is_clean` should assert that there are no error diagnostics, and the duplicate test should explicitly inspect the surviving `accept_param`. Its current comment describes preservation without asserting it. The implementation passes both stronger checks.

Recommendation: approve. Submitted as a comment review because this authenticated account is also the PR author.

PR #531, opened 2026-09-04
feat(types): perspective designation at a construction site (serves-aware literal override)

rileyr:
Track 0 item 2 of #525 (DNA epic #521). Confirms and closes the gap named in the first comment on #521.

`App { gw: Gateway { router: RouterV2 { } } }`, where `router` is `perspective(Router)`, was refused with `locus Gateway: field router expects Router, got RouterV2`, while the same override of an interface-typed field was fine. The param-default path (`conforms` in the params check) already consulted the impl's `serves` list; `check_literal_fields` only knew about interfaces, so a `perspective(P)` field, a plain `Ty::Named(P)` whose symbol is a Perspective, fell through to `assignable_from` and `P != Impl`. Codegen's designation branch in `lower_locus_instantiation` already fires for an override value, so the checker was the only thing in the way.

`check_literal_fields` now mirrors the param path. An override naming a locus that serves the perspective designates the slot. One that does not is refused with the `serves` clause it needs, and the generic mismatch is suppressed so the message is not doubled.

The slot stays program-global and 1-1. An override designates the program's impl exactly as a default does; two holders constructed with different impls are last-wins, the same as two holders with different defaults. Spec and docs say so rather than inventing a per-instance rule.

- `crates/hale-types/src/check.rs`: the `serves`-aware branch.
- `crates/hale-codegen/tests/fixtures/examples/65-perspective-ctor-override/`: runs and prints the V2 result (241, 2).
- `tests/hale/perspective_ctor_override_test.hl`: reads the routed value back through the holder.
- `crates/hale-types/tests/perspective_serves.rs`: override clean; non-serving override names the missing `serves` and suppresses the generic mismatch.
- `crates/hale-types/tests/fixtures/topology_shape_baseline.txt`: one added row for the new fixture.
- `spec/semantics.md` designation paragraph, `docs/src/services/perspectives.md`, CHANGELOG.

Verified: hale-types suite (1159), corpus oracle, syntax examples, hale-native suite, `hale run` on the fixture.

### DIFF
--- a/crates/hale-types/src/check.rs
+++ b/crates/hale-types/src/check.rs
@@ -13423,7 +13423,69 @@ impl<'a> Checker<'a> {
                     } else {
                         false
                     };
-                    if !interface_satisfied && !want.assignable_from(&got) {
+                    // GH #525 item 2 (2026-09-04): perspective
+                    // designation at a CONSTRUCTION site —
+                    // `App { gw: Gateway { router: RouterV2 { } } }`
+                    // where `router: perspective(Router)`. The
+                    // param-default path already asks `serves`
+                    // (the `conforms` computation in the params
+                    // check); this path only knew about
+                    // interfaces, so `perspective(P)` — a plain
+                    // `Ty::Named(P)` whose symbol is a Perspective,
+                    // not an Interface — fell through to
+                    // `assignable_from` and `P != Impl`. Codegen's
+                    // designation branch already fires for an
+                    // override value, so the checker was the only
+                    // thing refusing it. Note the slot is
+                    // program-global (1-1): an override designates
+                    // the whole program's slot, exactly as a
+                    // default does.
+                    let perspective_designated = if let (
+                        Ty::Named(pname),
+                        Ty::Named(arg_name),
+                    ) = (want, &got)
+                    {
+                        if matches!(
+                            self.top.lookup(pname),
+                            Some(TopSymbol::Perspective(_))
+                        ) {
+                            match self.top.symbols.get(arg_name) {
+                                Some(TopSymbol::Locus(li)) => {
+                                    if !li.serves.iter().any(|sv| sv == pname) {
+                                        self.diags.push(Diag::ty(
+                                            init.value.span(),
+                                            format!(
+                                                "{} `{}`: field `{}` is \
+                                                 `perspective({})`, but `{}` \
+                                                 does not serve it — declare \
+                                                 `locus {} : serves {}`",
+                                                kind_label,
+                                                name,
+                                                init.name.name,
+                                                pname,
+                                                arg_name,
+                                                arg_name,
+                                                pname
+                                            ),
+                                        ));
+                                    }
+                                    // Reported (or fine) — either way
+                                    // the generic mismatch below would
+                                    // only repeat it.
+                                    true
+                                }
+                                _ => false,
+                            }
+                        } else {
+                            false
+                        }
+                    } else {
+                        false
+                    };
+                    if !interface_satisfied
+                        && !perspective_designated
+                        && !want.assignable_from(&got)
+                    {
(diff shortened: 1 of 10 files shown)


crates/hale-types/src/check.rs
```diff
@@ -13423,7 +13423,69 @@ impl<'a> Checker<'a> {
...
+                    // the whole program's slot, exactly as a
+                    // default does.
+                    let perspective_designated = if let (
+                        Ty::Named(pname),
+                        Ty::Named(arg_name),
+                    ) = (want, &got)
+                    {
+                        if matches!(
+                            self.top.lookup(pname),
+                            Some(TopSymbol::Perspective(_))
```
rileyr, line 13459: [P2] Restrict the new designation exemption to supported constructor fields

`check_literal_fields` is shared by locus literals and pure data-type literals, but the new `perspective_designated` branch accepts both. Codegen only implements this designation in locus instantiation; `populate_user_type_fields` still rejects `Perspective` <- `LocusRef`. Consequently this PR changes the following program from a type error on the base commit into a successful `hale check` followed by a failed `hale run`/build:

```hale
perspective Router { fn route(code: Int) -> Int; }
locus RouterV2 : serves Router {
    fn route(code: Int) -> Int { return code + 200; }
}
type Holder { router: perspective(Router); }
fn main() {
    let holder = Holder { router: RouterV2 { } };
    println(holder.router.route(1));
}
```

At current head `47c36e33`, checking prints `ok: 1 file(s) typechecked`, then building fails with `type Holder field router type mismatch: declared Perspective("Router"), got LocusRef("RouterV2")`. The exact base rejects the initializer with `field router expects Router, got RouterV2`. Gate this exemption on the supported locus-parameter construction case, or supply matching lowering for data-type fields if that is intended, and add a check/build-agreement regression test.

rileyr (commented):
Reviewed current head `47c36e33a5703e7a1a5f0a22cbf684b5e74207eb` against `2a36bfb3535c8b81fd5b5ae6e997bd99398d524f`.

Recommendation: changes requested for the P2 checker/codegen disagreement detailed inline. The intended locus-constructor override works, and non-serving implementations are rejected, but the shared checker also accepts data-record field initializers whose lowering does not support perspective designation.

Validation:

- All 1,159 `hale-types` tests passed in the prescribed release profile at `0b55ba26`; the subsequent `47c36e33` commit only changes the effects baseline.
- Five supplemental checker checks passed: error-free serving override, a required slot, exactly one missing-serves error, rejection of a malformed serving implementation, and same-seed cross-file resolution.
- Built the PR CLI; the new fixture produced `241` and `2`, and the new native test passed.
- `cargo test --offline --release -p hale-cli --test hale_native_suite` passed at current head.
- The effects-baseline gate passes after `47c36e33`. The earlier CI failures were caused by the now-fixed missing fixture baseline and are not an outstanding finding.
- Independently reproduced the inline check/build mismatch at current head and verified that the exact base rejects the same program during type checking.

Submitted as a comment review because this authenticated account is also the PR author.

crates/hale-types/src/check.rs
rileyr: Confirmed and fixed in c29e57aa. The exemption is now gated on the locus literal path (`kind_label == "locus"`); a data-type field keeps the base's plain mismatch, so your `Holder` program is refused by `hale check` again with `type Holder: field router expects Router, got RouterV2`. Regression coverage: `ctor_override_on_a_data_type_field_is_still_rejected` in `perspective_serves.rs`, and a new `perspective_ctor_override_agreement` test in hale-codegen that builds and runs the locus case (routes 201 through the override) and asserts the data-type case is refused by the checker rather than the build. I did not add a data-type lowering: a `type` is a pure value and designating a program-global slot from a value literal has no owner for the impl's lifetime, so refusal is the right answer there.

PR #532, opened 2026-09-04
fix(obs): aux_b carries the canonical entity id; retire the v0 binding/scheduler meaning, proto 0.4

rileyr:
Track 0 item 3 of #525 (DNA epic #521). Answers iris handoff-14 P31 (`HANDOFF-iris-14-aux-b-collision.md`).

Iris's PROTOCOL §4 and its reference emitters had used a manifest row's `aux_b` since v0 as binding → owning topic id and scheduler → cpu index. hale's native emitter, from proto 0.3, wrote the canonical model entity id there for every kind, and the rationale in `lotus_obs.c` said the field had "been written as 0 by every path", which was true of this emitter and false of the protocol's. The two meanings are not distinguishable from the value. Nothing was broken only because no consumer read the field yet. Handoff-14 asked hale to pick: give the entity id its own field, or retire v0's meaning.

**Retired.** The entity id is the #476 join and what every consumer arriving next (Track B's model-diff perspective, Track C's experience view) will read; v0's use was one binding row and the scheduler rows in `synth.c`. `proto_minor` is now 4 with no layout change, because the meaning moved and a consumer depends on the meaning, not the offset. Every emitter writes the entity id or 0, the scheduler cpu index moves to `aux_a`, and the binding → topic pairing is dropped (the counter table and the binding name carry it). A consumer still gates on `entity_id_digest != 0`; at minor ≥ 4 it may additionally trust a nonzero `aux_b` was never anything else.

- `crates/hale-codegen/runtime/lotus_obs.c`: `proto_minor = 4`, rationale corrected and credited.
- `crates/hale-codegen/tests/obs_entity_ids_unstamped.rs`: pins `proto_minor == 4` in the header so a future bump is a protocol decision, not a test fix.
- `spec/runtime.md` § Native observation emission, CHANGELOG.

**Paired iris change**, not in this PR because the iris worktree carries uncommitted inspector work (item 4 of #525): `UPSTREAM-NOTE-2026-09-04-aux-b.md` plus `aux_b_proto04.patch` are dropped untracked in the iris tree, cut against its current working tree. The patch touches `emitter/protocol.h` (`OBS_PROTO_MINOR 4`, history line, `aux_b` comment), `emitter/synth.c`, `observe/glue.c`, and PROTOCOL §4/§13. Verified in a scratch copy: `git apply --check` clean, `emitter/` and `consumer/` build, `observe/glue.c` compiles with `-Wall`.

Verified here: all 19 `obs_*` integration tests in hale-codegen pass at 0.4.

With this in, PROTOCOL §13 has no open item blocking the v0 freeze on the hale side.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,28 @@ behavior.
 
 ## Unreleased
 
+### Observation proto 0.4: `aux_b` means one thing (GH #525, iris handoff-14 P31)
+
+Iris's PROTOCOL §4 and its reference emitters had used a manifest
+row's `aux_b` since v0 as *binding → owning topic id, scheduler →
+cpu index*. hale's native emitter, from proto 0.3, wrote the
+canonical model entity id there for every kind — and this file's
+rationale said the field had "been written as 0 by every path",
+which was true of hale's emitter and false of the protocol's. The
+two meanings are not distinguishable from the value; nothing was
+broken only because no consumer read the field yet.
+
+Resolved by retiring the v0 meaning rather than moving the entity
+id: `proto_minor` is now 4, every emitter writes the entity id or
+0, the scheduler cpu index moves to `aux_a`, and the binding →
+topic pairing is dropped (the counter table and the binding name
+carry it). No layout change. A consumer still gates on
+`entity_id_digest != 0`; at minor ≥ 4 it may additionally trust
+that a nonzero `aux_b` was never anything else. The paired iris
+change (protocol.h, synth.c, observe/glue.c, PROTOCOL §4/§13) is
+handed over as `UPSTREAM-NOTE-2026-09-04-aux-b.md` in the iris
+tree. Last open item before the protocol freezes at v0.
+
 ### A second `accept` clause is an error, not a silent overwrite (GH #525)
 
 A locus accepts exactly one child type (`spec/types.md` F.11,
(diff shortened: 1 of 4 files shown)


rileyr (commented):
Reviewed `6d45a61f4d299b07db04874ff09f0bdd983f37d8` against `2a36bfb3535c8b81fd5b5ae6e997bd99398d524f`.

No merge-blocking correctness findings in the Hale-side change. The native emitter already publishes canonical entity IDs in `aux_b`; the minor-version bump makes the revised meaning explicit without changing the binary layout. Retaining `entity_id_digest` as the guard for canonical identity is correct; version 0.4 alone must not imply that a segment carries stamped IDs.

Validation at this commit:

- Compiled the full native emitter and exercised stamped and unstamped topic/locus/binding manifests: version 0.4, expected IDs/digest, and unchanged layout.
- The current unpatched Iris consumer attached to and read both generated segments successfully.
- Applied the paired Iris patch only in a scratch copy; its emitter, consumer tools, and observation glue compiled successfully.
- All currently reported PR CI checks are green, including all four test shards, macOS, sanitizers, TSAN, and GenMC.

Release coordination note: the paired Iris update remains an untracked patch in that repository. This PR can establish the Hale-side 0.4 contract, but Track 0 item 3's cross-emitter acceptance and the protocol freeze should remain open until the Iris changes are committed and landed. The PR description already discloses that dependency.

Recommendation: approve the Hale-side change. Submitted as a comment review because this authenticated account is also the PR author.

PR #565, opened 2026-09-10
docs: DNA has its own section of the Book (nine chapters, one real session)

rileyr:
Stacked on #564 (D7), because the chapters describe the whole of Track D.

`docs/src/dna/` replaces the single "Systems control" page with a section of nine chapters. The site's left nav is generated from `SUMMARY.md`, so this becomes a new **DNA** section on hale-lang.org/docs on the next sync, with no site work.

- **A governed application** — what DNA is, the three words (genome, expression, experience), what it is *not* (the kill test's finding, stated), and what it does mechanically every time.
- **Attaching it** — `init` output, what is generated and who owns it, the diff to your main, the Genome, the law read as four promises, the purpose Review, the seeded Journal.
- **Running it** — the host, the four commands, the membrane's three topics, what the organism leaves under `.hale/dna`.
- **A change, end to end** — one real session on the acceptance chat server with scripted models, every step's output and Journal rows. Not the CI assertions: the transcript.
- **Reviewing** — the full `hale dna review m1` render, why the three views are always together (the kill test in short form), what the verdict checks and in what order, iris.
- **Apply, restart, observe** — what approval does, the pin at apply time, the idempotent apply, the restart, the window, the crash path, and what this phase does not do (live state migration).
- **Autonomy and policy** — grants, the magnitude vector as a table, `dispose` as five rules with the evidence arithmetic, post-review autonomy, what a child cannot do, pressure and appendages, versioned Workflow definitions.
- **Models and credentials** — the three backends, selection, evidence never the prompt, the editor's grant, sealed credentials.
- **Reference** — the CLI, the Journal's vocabulary as a table, `status.json`, the files, the core by file.

Two small things the session showed, fixed here: the generated Genome's header comment still described Phase 1; `hale dna status` printed an empty `()` for a mutation that named no target.

Gates: docs snippet parse gate green, `mdbook build docs` clean, affected CLI suites green, `hale fmt --check .` clean.

The site guide page ("what is DNA, how do I use it", a top-level page beside Features and Tooling) is a separate PR in the site repository.


rileyr:
Closing per the Phase 3 plan: the Book section is rewritten once, for the final shape (the organization as the organism), in track F6. The branch stays for reference.

Issue #583, opened 2026-09-11
Hale DNA, Phase 4: model backends as a catalog, and the knowledge graph as a service (successor to #566)


rileyr:
Phase 3 (#566) landed the organization as its own program, the record as a git branch, the fleet as the expression, and the host in Hale. Two things it left where Phase 0 put them: model backends are three hardcoded OpenAI-shaped routers copied into `dna/org/main.hl`, and the knowledge graph is an in-memory locus nobody wires. This phase moves both to where they belong. Design agreed 2026-09-11; contracts sharpened the same day after a second review (the harness boundary, the tape, the ratification protocol, the learning scenario, the budget, dev's Postgres).

## Track M: model backends

**The seam stays.** `ModelBackend` (`identity`, `model_name`, `allows`, `capacity_bytes`, `complete`) is the interface every adapter implements, each publishing `ModelCalled` evidence. Routers keep returning data, never backends.

**Adapters.** `OpenAiChat` (today's `HostedModel`, named for what it speaks: OpenAI, Ollama, vLLM, OpenRouter, Anthropic's compatibility endpoint); `AnthropicMessages` (the native messages API; port the body builder and response parser from pond `agent/llm/anthropic.hl` onto the DNA's `std::http` path); `HarnessModel` (an installed CLI such as `claude` or `codex`, run per request); `FakeModel` (unchanged); `RecordedModel` (below). Credentials stay a sealed locus that never returns the bytes and gain a `scheme` (`bearer`, or `x-api-key` with `anthropic-version`) so no adapter touches the material.

**The catalog is source.** Hale holds no dynamic collection of loci and the core keeps the CQRS rule, so the registry is `dna/org/models.hl`, generated by `init`: backend constructor functions by name (`frontier()`, `fast()`, `desk()` …) and per-position router functions (`leader_models()`, `editor_models()`, `agent_models()`) composed from them. A new position as the org grows is one more function; switching provider is one file; `hale check` validates all of it and the law keeps concrete types at every boundary.

**The budget has one owner: the record.** The catalog declares the policy once, as data (`org_budget()` returns allowances per window and per backend name); it never constructs a counter, so naming the budget from several constructors cannot mint several allowances. The assembly owns the single counter, fed by the `cost_micros` of every `model.called` row and rehydrated from the journal at boot, so a restart loses nothing and a position created later is covered by construction. When a window is exhausted the assembly stops routing model-backed work and journals `budget.exhausted`; the Board still decides everything it decided before.

[...]

PR #586, opened 2026-09-11
dna: the harness works in an export; the genome is unreachable from it (GH #583 M3)

rileyr:
Part of #583 (Phase 4), Track M, M3: the harness backend, the export, `Confinement`. Stacked on #585 (M2).

## What changes

**`HarnessModel`.** An installed coding harness (`command: "claude"`, or `codex` with `output: "text"`) run per request with its own tools on, as a backend that `works_in_place`. For a source-editing request the editor makes an **export** of the Mutation's worktree (a plain directory with the same files, never `.git`, never `.hale`), runs the harness there with the whole objective as its prompt, and imports the export's diff back under the grant: a changed or new file is written through the same tools a chat model's edit goes through, a deleted one is removed, a change outside the grant is counted (`outside_grant`) and left behind. `files_changed` is derived from that diff, never from the harness's answer. Then the same fmt, check, retry (diagnostics in the next prompt) and assessment as the file-by-file flow. An answer-only request (a review) runs in an empty directory of its own. Evidence is one `model.called` row per call with the harness's own cost summary (`total_cost_usd`, `usage`), `tool_grant: harness @export`, and `confinement=<kind>` in `params`.

**The boundary, stated exactly.** The export decides what DNA accepts; it is not a sandbox. What DNA guarantees is that *the genome is unreachable from the harness process*: a `Confinement` (`kind`, `available`, `wrap(argv, cwd, mask)`) supplies it — `Bubblewrap` on Linux (the filesystem as the operator sees it: the harness's own home state, the toolchain, the network; the repository and the worktree replaced by empty tmpfs mounts; `--die-with-parent`) or `NoConfinement`, which is refused unless the org chart's `allow_unconfined` says otherwise. The evidence records which it was. On every platform the assembly also checks that neither the repository's head nor the worktree's moved during any attempt and fails the Mutation by the record (`mutation.failed`: `the genome moved during the attempt`) if they did: not a boundary, but a bypass becomes a loud refusal. Outside the genome the harness has exactly the operator's account, which is what running it by hand has. `effect harness_run` names the reach; `Repository.root()` names what is masked.

**Discovery.** `claude` (else `codex`) on `PATH` becomes `harness()`: the editor's and the agent's quick tier, and every model-backed slot when there is no key, so a laptop with a harness and no key still edits, reviews and decides. The probe runs it too.

**Seams.** `ModelBackend.works_in_place()`, `ModelRouter.in_place(req)`, `ModelRequest { workspace, mask }`, `SourceEditor.place(worktree, seed, repo)` (the assembly's call; `confine` still works for tests), `WorktreeTools.remove`. The assembly skips locating a file first when the editor's backend works in place.

## Tests

- `dna/tests/harness_test.hl`: a stand-in harness (a shell script playing `claude -p --output-format json`) edits the seed's main file in its export, leaves a file outside the grant, tries to read the genome by its real path, and prints the result object with a cost. One Dna over a real repository runs a Mutation through it: the candidate carries the edit, the stray file is left behind and counted, the evidence has `cost_micros: 12300`, `tool_grant: harness @export` and the confinement, the export is gone, and under Bubblewrap the harness read **zero bytes** of the genome (the mask is exercised where `bwrap` is present; reported as skipped otherwise). A harness that commits to the genome fails the attempt by the record. An unconfined harness is refused without the org chart's leave; a harness that is not on PATH is not permitted; customer data never goes to one. An answer-only call runs in a directory of its own and never touches the repository.
- `dna_models.rs`: a stand-in `claude` on PATH becomes `harness()`; with no key every slot is the harness, with a key the editor's quick tier only; the organization checks with it wired; `hale dna models` runs the harness and shows its cost.

Spec (the harness, the boundary, discovery), the Book (models: "The harness works in an export"; shaping; reference), CHANGELOG.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,13 @@ behavior.
 - `hale dna models`: the catalog probed — one line per backend with one small request to each that is permitted; the organization is not started and nothing is journaled.
 - `HostedModel` is `OpenAiChat` (`adapter: openai-chat`), named for what it speaks; `FakeModel { fail_after }` refuses after that many calls, for tests that fail an Attempt on cue.
 
+### DNA: the harness works in an export; the genome is unreachable from it (GH #583 M3)
+
+- `HarnessModel`: an installed coding harness (`claude`, or `codex` with `output: text`) as a backend that works in place. For a source-editing request the editor exports the Mutation's worktree (a plain directory, never `.git`, never `.hale`), runs the harness there with the whole objective and its own tools on, and imports the export's diff back under the grant: a changed or new file is written through the tools, a deleted one removed, a change outside the grant counted and left behind; `files_changed` is derived from the diff, never from the answer. Then the same fmt, check, retry and assessment as the file-by-file flow. An answer-only call runs in an empty directory of its own. Evidence: one `model.called` row with the harness's own cost summary, `tool_grant: harness @export`, `confinement=<kind>`.
+- `Confinement`: `Bubblewrap` (Linux; the repository and the worktree replaced by empty tmpfs mounts, everything else — the harness's home state, the toolchain, the network — as the operator sees it) or `NoConfinement`; an unavailable confinement refuses the harness unless the org chart's `allow_unconfined` says otherwise. The assembly checks that neither the repository's head nor the worktree's moved during any attempt and fails the Mutation by the record if they did. `effect harness_run` names the reach; `Repository.root()` names what is masked.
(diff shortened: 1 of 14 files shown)


rileyr:
Pushed two CI-only fixes on this branch: a `HarnessModel` whose confinement is unavailable but whose org chart allows running unconfined now runs the harness plainly (it still went through the wrapper, so with no `bwrap` the call failed with `exit -1: not_found` instead of doing what the evidence promised), and the fixture's own notice that the mask was not exercised goes to stderr, since `hale test` requires a passing test to be silent. CI has no `bwrap`; the fixture found both.

PR #584, opened 2026-09-11
dna: the model catalog is source, the budget has one owner (GH #583 M1)

rileyr:
Part of #583 (Phase 4), Track M, M1: the catalog, discovery, the budget, `hale dna models`.

## What changes

**The catalog is source.** `hale dna init` / `new` write `dna/org/models.hl`, owned by the project: a backend is a constructor function (`frontier()`, `fast()`, `desk()`), a position's router is a function composed from them (`leader_models()`, `editor_models()`, `agent_models()`), and the organization's main takes every router from it — no adapter is named inline any more. A new position is one more function; a provider switch is one file; `hale check` validates it and the law keeps the concrete types at every boundary.

**Discovery.** `init` looks at `ANTHROPIC_API_KEY` / `OPENAI_API_KEY` and at `PATH` (`ollama`, whose first listed model becomes the desk model; `claude` / `codex` reported), writes the catalog for what is there, and prints what each position was given:

```
created …/dna/org/models.hl
models  found   ANTHROPIC_API_KEY set, ollama on PATH (llama3)
models  frontier = claude-opus-5 · fast = claude-haiku-4-5-20251001 (ANTHROPIC_API_KEY) · desk = llama3 (ollama at 127.0.0.1:11434)
models  leader, editor, agent: deep = frontier, quick = fast, private = desk · budget 25.00 USD a day (`hale dna models` probes them)
```

With no key the hosted backends are not permitted and every Review waits for the Board, as before. `hale dna upgrade` writes a catalog for an organization from before it and says what to point at it; it never edits the project's main.

**The budget has one owner.** The catalog declares `org_budget()` (a `BudgetPolicy`: `window` day / week / none, `allowance_micros`) as data; the substrate owns the one `Budget`. Every `model.called` row is accounted at its time (the evidence now carries `at`), the counters are rehydrated from the record at birth (a restart loses nothing; a position added later is covered by construction), an exhausted window refuses intent with the reason, journals `budget.exhausted` once, tells the membrane, and leaves the Reviews to the Board, which needs no model. `ModelRouter` no longer has a `daily_cost_ceiling_micros` of its own: naming the budget from several constructors cannot mint several allowances.

**`hale dna models`** builds the catalog beside a one-line main in `.hale/dna/probe` and sends one small request to each backend:

```
catalog dna/org/models.hl
backend     slot      model                         answer
frontier    deep      deep-1                        ok  15789ns  50 micro-dollars  "ready"
[...]

Issue #402, opened 2026-08-05
Factory-returned loci still leak in two shapes: unbound temporaries, and factories with a call-valued return arm


rileyr:
Follow-up to #383, which fixed the general case: a locus returned by a proven-fresh free-fn factory is now owned by the binding that names it and dissolves at that scope's exit. Two narrower shapes still reach the program-lifetime path and leak. Both are characterized and reproducible; neither needs a design decision.

## 1. Unbound temporaries

```hale
let z = mat::add(mat::matmul(w, a), b);
```

The inner `matmul` result is never named, so the binding-scoped rule that fixed the general case cannot reach it — there is no binding to attach ownership to. Needs registration at expression level: a call to a known-fresh factory in argument position produces a value the enclosing statement owns, dissolvable at statement (or scope) end.

The ownership question is already settled by the v0.14 rule (a locus value cannot be assigned into a locus-typed field), so this is placement of the registration, not semantics.

## 2. Factories whose second return arm is a call

```hale
fn matmul(a: Matrix, b: Matrix) -> Matrix {
    let _check = check_matmul_shapes(...) or -1;
    if _check < 0 { return error_matrix(); }   // ← this arm
    let out = zeros(a.rows, b.cols);
    ...
    return out;
}
```

`compute_fresh_locus_factories` requires every return to be an `L { … }` literal or a single let-bound ident, so a `return <call>` arm disqualifies the whole fn — conservatively, which was the right default while the ownership rule was unsettled. Admitting a return that is itself a call to an already-qualifying factory would cover this, and the fixpoint already has the machinery to decide it.

Worth noting the interaction: `matmul` and friends are also the functions whose results are most often unbound temporaries, so fixing (2) alone moves less than it looks like it should. (1) is the higher-value half.

## Measurement

Reference workload (pond `math/matrix` + `ml/neural`, differential probe from #381): after #383's fix, values correct, no UAF, **8400 → 5320 bytes**. The remaining 5320 is these two shapes. Scaling check: the residual is still linear in call count (5 train_steps ≈ 21 KB, 40 ≈ 207 KB), so it matters for long-running loops.

## Reproducing

The probes used on #383 are the fastest way in — `probe3` (two train_steps, ASan) for the byte count and the #381 differential probe for correctness. **Validate against both**: two earlier attempts on #383 were leak-clean and numerically wrong, and that state looks like success under LeakSanitizer alone.

rileyr:
## Status after #403: half delivered, issue stays open

[...]

PR #587, opened 2026-09-11
dna: the tape, and the recorded fixture (GH #583 M4)

rileyr:
Part of #583 (Phase 4), Track M, M4: the tape and the recorded fixture. Stacked on #586 (M3). Closes Track M.

Two commits: the compiler fixes the fixture found (F.17, F.18), then M4 itself. **The chain #584 → #585 → #586 → this merges together**: the catalog #584 writes (`ModelRouter { quick: fast(), deep: frontier() }`) is only sound at runtime with the first commit here.

## Two compiler holes, found by running the organization for real

**F.17 — a literal owns a factory result written as its field.** `Router { quick: make("q") }` inside another factory registered `make`'s fresh result as a temporary of the enclosing frame (the GH #402 hook) and dissolved it at that frame's exit while the field, a fat pointer for an interface-typed slot, still pointed at it. The catalog's routers read garbage from their backends under a live organization: first a corrupted `dir` string, then a segfault in an interface dispatch, pinpointed with a `LOTUS_ASAN=1` build of the organization. A fresh-factory call written as a locus- or interface-typed field of a literal, explicit or default, is now owned by the literal, the same one-shot suppression a `let` RHS and an `=` into a locus slot already had. `tests/hale/factory_field_owner_test.hl` pins the shape; spec/semantics.md says it.

**F.18 — `check` refuses a call to a name nothing binds.** Under pressure the model wrote `models: fulfilment_models()` with no such function; `hale check` passed, the Board approved, the gateway applied, `hale build` refused at expression and the window rolled the organization back. Right machinery, wrong gate. The checker holds codegen's rule now (`call to X: no free fn, generic fn or fn-pointer binding with that name is in scope`, with a did-you-mean); zero new findings across the in-repo corpus. `check_unbound_callee.rs` pins it; spec/types.md says it.

## What changes

**`RecordedModel { dir, mode, inner }`** wraps any backend. The key is `sha256` over the fields that identify a request: role, the inner's name and model, the prompt and context digests, the data class, the grant normalized (`… @grant`: its path is where it ran, not what it was) and, for a backend that works in place, a digest of the workspace's starting tree. `record` forwards to the inner (whose evidence is the call's) and writes `<key>.json` with every keyed field in clear and the answer, plus for a workspace `<key>.patch`: what the backend changed there **before any import**, so a replay reproduces the edit and everything after it (the import, the grant, fmt, check) runs for real. `replay` answers from the directory (evidence `adapter: recorded`, `params: tape=<key> mode=replay`), applies the patch, and refuses a miss naming the request and, when an entry shares the prompt digest, the fields that differed. Replay is pure and keyless; record reaches whatever the inner reaches. What varies between runs must stay out of the key, and recording finds what does: the editor now strips the worktree's path from the diagnostics it feeds back on a retry.

**The fixture.** `dna/acceptance/trio`: a gateway (the root seed), an api and a worker under a plan of four instances on two nodes, with routes between them and claims across them (`require_publishes` on the gateway and the api, `forbid_reaches` gateway → worker avoiding the api). Its organization, every backend a `RecordedModel` over `AnthropicMessages` (`dna/acceptance/trio.fixture/catalog.hl`; mode from `HALE_DNA_TAPE`, the tape from `HALE_DNA_TAPE_DIR`), is driven by scripted asks end to end in `dna_recorded_fixture.rs`:

1. a change to the gateway, asked, reviewed by the Board, applied, and deployed to both nodes (`fleet.deploy` touching `gateway-0 gateway-1`, `instance.up` at the candidate on `edge-1` and `edge-2`);
2. pressure from the worker three times: the organization grows a supervisor position in its own main, an `organization` change the Board approves; the organization is restarted with it;
3. a change to one service that still checks but breaks the fleet's law (the gateway renames the subject the api is routed on): `mutation.deny` with `check=0 fleet=1`, never a Review.

Every model call is answered from the checked-in tape, keyless (`adapter: recorded`, no miss): six entries, recorded once on `claude-sonnet-5` for about ten cents; the keyless replay runs in 19 s. `HALE_DNA_TAPE=record cargo test --release -p hale-cli --test dna_recorded_fixture` with a key re-records it. This is the acceptance for every later change to the organization; it proves how the organization handles recorded outcomes, and model quality is only ever tested by a fresh run.

## Tests

- `dna/tests/recorded_model_test.hl`: an answer recorded and replayed into a fresh wrapper with the recorded cost and `recorded` evidence, the grant's path aside; a miss names the request and the nearest entry's differing field; a stand-in harness's edits recorded as a patch (before import) and replayed onto a fresh workspace with the stand-in made unrunnable; a workspace whose starting tree differs is another request.
- `crates/hale-cli/tests/dna_recorded_fixture.rs`: the scenario above, in replay mode by default.

Also: `fail_in_flight` no longer takes `mutation.worktree … removed` for an open worktree (a restart marked a retained Mutation failed); the growth objective names the catalog's router; `list_files` is shared by the editor and the tape.

Spec (the tape, the fixture; semantics and types for the compiler rules), the Book (models: "Recording and replay"; reference), CHANGELOG, dna/FRICTION.md F.17 and F.18.


rileyr:
Rebased: the compiler fix moved to the bottom of the chain as #592; this PR now carries M4 alone.

rileyr:
Pushed a fix for a CI-only race: a model call's evidence is journaled by the assembly's bus handler, so it lands after the mutation row the test waited on. The test judged a snapshot taken before the last attempt's rows arrived, which was luck locally. It now waits for the last attempt's evidence and re-reads the record.

PR #593, opened 2026-09-12
dna: the twelve findings of the 12 September shakeout

rileyr:
An independent shakeout of `main` (ab18c87) on 12 September found twelve defects: four P1 and eight P2. All twelve are fixed here, each with a test that fails without the fix.

The review's central criticism is the right one and worth repeating: every new seam of Phase 4 was tested against a stand-in written alongside it, and each stand-in was forgiving exactly where the real thing is strict. A harness stand-in that ignored its arguments could not notice that the prompt arrived as one argument per line. A store fixture that opened the store itself could not notice that no startup path ever did. A tape fixture that recorded a text-only answer first could not notice that the directory did not exist. The tests added here are contract tests: they assert what the real component receives, not what a convenient double accepts.

### P1

1. **Answer-only harness requests did not protect the genome.** The confinement's mask came from the request and only the editor's call set one, so a review, an agent's call, the editor's assessment and the catalog's probe ran with an empty mask while the evidence recorded `confinement=bubblewrap`. The genome is the model's own now (`genome` / `genome_env`, read at birth like a credential's source; the host exports `HALE_DNA_GENOME`); a request's mask adds what that call knows besides; a confinement with nothing to mask is refused rather than run, so the claim and the fact cannot diverge. Evidence records `masked=<n>`.
2. **A moved destination could apply an unreviewed combination.** `LocalGit.apply` fell back to a merge commit, and the staleness check looked only at the candidate's worktree. The substrate now checks the destination against the base the review was against, and the gateway fast-forwards to the exact candidate or fails.
4. **The Postgres knowledge service never opened its database.** `Pq.open()` is what connects and migrates and no startup path called it, so the service answered 200 with an empty graph and a well-formed digest while every query failed. It opens lazily on the first request, keeps retrying, reports `open` and the reason in its summary, and answers every other route with 503. Lazily, because a database that is down must not hold closed the surface an operator reads that on.
5. **Multiline prompts were split into process arguments.** Hale's argv is newline-separated and every real prompt has newlines, so a harness received one argument per line. The prompt arrives on stdin now, through a file this process owns outside the workspace so an import never sees it.

### P2

3. Unknown authority names satisfied stronger requirements by substring: `not-board` satisfied `board`. An unknown name satisfies only itself.
6. Ordinary project paths broke every control command. A Unix address holds 108 bytes and `.hale/dna/hale-dna.review.verdict.sock` spends 39; the organization binds relative, the client wrote absolute. Routes are relative on both sides now.
7. The first sync did not publish a local-only record, so a record made in one clone stayed there. A remote with no record is the local one's push, and the guide no longer tells the reader to push `refs/dna/*` by hand.
8. Postgres retrieval used `LIKE`, where `_` is a wildcard and an ordinary character in a project name: a practice bound to `org/orders_api` was retrieved for `org/ordersXapi`. Literal prefixes now.
9. Postgres rows were tab- and newline-separated with no escaping, so an ordinary paragraph came back cut at the first line with its provenance shifted into nothing. Rows are read as JSON the server builds.
10. The first recording in a fresh tape lost its patch while the entry still claimed one. The directory is made first, and an entry counts only once every file of it has landed.
11. Importing an export dropped additions inside new directories. Parents are made under the grant, and an in-grant write that fails is the attempt's failure rather than a partial candidate.
12. Signal projections double-counted across an interrupted watermark update. Signals count once per record row, keyed on its sequence.

### Tests

A harness contract fixture that records argument count, every argument and the exact stdin bytes; the knowledge service itself against a real Postgres and against a database that is down; a maintainer commit landing during a review; underscored paths, a paragraph with a tab, and a replay of the crash-window row through both stores; an export that adds a module in a new directory; an in-place edit recorded as the first entry of a tape that does not exist; a first sync with a second clone that acquires the record by itself; and an organization run from a path where the absolute socket address does not fit.

Also here: a retried apply replays instead of being refused as a moved genome (the destination check ran before the apply's own idempotency key was consulted), and the topology baseline records the knowledge fixture's second embedded program.


rileyr:
Second-round review addressed: all five findings fixed, each with a test that fails when the fix is reverted (checked, one by one).

**1 (P1), a successful fast-forward did not establish the expression.** Both reproductions are real and both are fixed, in two places rather than one. A destination carrying uncommitted work is refused *before* the apply's effect row exists, naming the files and leaving them alone, so the maintainer commits or stashes and approves again and the whole gate runs on the new base. The gateway then checks the destination once more immediately around the git call and reads the result back, so `merge --ff-only <ancestor>` reporting "already up to date" is a failed apply rather than a silent one. That second check is where the race lands, and a refusal there ends the Mutation, which is correct: the genome moved. A refusal now carries the repository's own reason instead of a bare `apply failed`.

The fixture covers a maintainer mid-edit whose work is still theirs afterwards, and an ancestor that git accepts and the gateway refuses. I did not take the further step of expressing from an immutable artifact of the reviewed candidate rather than the working directory. That is the right shape and it is a larger change than a repair; it is worth its own issue.

**2 (P2), the catalog probe had no genome.** Every verb the host runs is given `HALE_DNA_GENOME`, not only the organization under `run` and `dev`. The fixture now probes the catalog **as generated**, with its `Bubblewrap` confinement intact, instead of one with the confinement swapped out — which is exactly why this passed before. Reverting the one-line export reproduces your refusal verbatim.

**3 (P2), codex lost its write permission.** Fixed as you describe: `exec --skip-git-repo-check --sandbox workspace-write`, with the generated comment saying why. I could not verify against an installed codex, so this rests on the documented default you cited and is scoped to the working directory; DNA's own confinement still holds the genome out of reach.

**4 (P2), an established connection could die unnoticed.** The sticky flag is gone. The store is asked whether it still answers on each request, a dead session is re-established, and only a store that cannot be reached at all answers 503. The real-Postgres test terminates the service's own backend mid-run and expects the same package afterwards; with the sticky flag restored it fails with `"open": true, "watermark": 0, "error": "watermark: io peer closed mid-response"`, which is your reproduction.

**5 (P2), a failed removal reported success.** The unlink error travels like a write's and the path is checked afterwards, so the incomplete-import check can see a deletion that did not happen.

**Isolation note.** Taken: the real-Postgres service test creates a database of its own where the server grants one, and falls back to the shared DSN when it cannot.

Local: `hale fmt --check .` clean, `hale verify dna/core` clean, 23 DNA fixtures, 43 DNA integration tests, and the five pinned-fixture gates with no baseline movement. The Postgres tests ran against a disposable pgvector instance.

PR #594, opened 2026-09-12
dna: what a seed compiles, and what a store cannot answer

rileyr:
A third review round on merged main (`99d3e0f0`) confirmed the twelve shakeout findings and the five follow-ups are closed, and found four more: one P1 and three P2s. All four are fixed here, each with a test that fails when the fix is reverted.

**1 (P1), untracked source bypassed the dirty-genome check.** Correct, and the sharper part of the argument is the right one: `hale check` reads the directory, not the index, so an untracked `helper.hl` beside the reviewed `main.hl` is built by the very seed that expresses the candidate.

The first rule I wrote counted every untracked file git does not ignore. The suite rejected it within minutes: `hale build` leaves a binary beside every seed it builds, the organization's included, so an ordinary working project is untracked-dirty by construction and every apply refused. The rule that ships is narrower and follows how Hale actually compiles. A seed is one directory of `.hl` files, so untracked source **of the genome's seed** counts, along with `hale.toml`. Another program's source in a directory of its own, a build artifact, a log, anything under `.hale/` does not. The fixture asserts both halves: an untracked `helper.hl` is refused by name, and a seed artifact plus an unrelated `elsewhere/driver.hl` are not dirt.

**2 (P2), "stash and approve again" could not recover.** The verdict is spent; the apply is not. Approving again on an approved-but-unapplied candidate re-runs the whole gate over the same pinned candidate and records `mutation.apply_retried`. Nothing is relaxed: a candidate or base that moved in the meantime is refused exactly as before. I chose this over reopening the review because the human's verdict stands and it was the apply that was refused, and because every route already publishes `ReviewVerdict`, so the CLI, the membrane and the UI all recover the same way.

**3 (P2), query failures still became empty packages.** Right, and the diagnosis is exact: `SELECT 1` establishes the session, not the queries. The Postgres store records the errors of its reads as it already did its writes, and the service refuses with 503 and the reason on `/context`, `/structure`, `/signals` and `/idea/`. A failed record projection is a refusal too, rather than a package built from a graph that is stuck. Your other observation is fixed with it: the summary reported no error while every count returned zero, which reads as an empty graph. It now carries whatever the counts hit. The test injects your fault, renaming `knowledge_ideas` away and back.

**4 (P2), new empty files disappeared on import.** Absence is not emptiness. The harness fixture's stand-in now adds an empty `app/lib/.gitkeep` beside the new module, and the candidate is asserted to carry it.

On the deferred redesign: expressing from an immutable artifact of the reviewed candidate rather than the working directory remains the right shape, and finding 1 is a good argument for it. It is still a design change rather than a repair, and it is not in this PR.

Local: `hale fmt --check .` clean, `hale verify dna/core` clean, 23 DNA fixtures, 43 DNA integration tests, the five pinned-fixture gates with no baseline movement, and the Postgres tests against a disposable pgvector instance.


rileyr:
Fourth round addressed at `9deade9e`. Two of the four were holes in my own repairs from the round before, and both are the kind of hole a reviewer earns their keep finding.

**1 (P1), a second rejection could apply the candidate.** Yes. The retry keyed on the Review's `already settled` refusal, which is issued before the incoming verdict is examined, so it carried no admission at all. Readmission is the Review's now: at a settled approval it admits only the same approval in full — pinned digest, required authority, independence from the author, and the word `approve` — and publishes a settlement marked `reapproval`; every other verdict is refused with the reason. The assembly retries on that and nothing else. The fixture sends a rejection, a wrong digest and a weak authority at a settled, unapplied review; each applies nothing and is refused by name, and no `mutation.apply_retried` row exists until the same approval arrives.

**2 (P1), the source check missed real inputs.** Also yes, and your table is the argument. The narrowed rule was still a guess about what a build reads. It now asks the compiler: a new `hale inputs <seed>` prints every file a build of the seed reads, imports followed, resolved exactly as `check` and `build` resolve them, and any of those the candidate's tree does not hold is dirt. Each path is asked of git one at a time, so a quoted name is never parsed out of a status line, and an ignored file is listed because the compiler lists it. The fixture covers your four rows plus the artifact and the unrelated program, and drives the imported-library case through the whole gate.

**3 (P2), the retry route vanished after a restart.** A Review approved and never applied is rehydrated settled, so the second approval after a restart is admitted by the same locus under the same rules. The fixture builds a second organization over the same record after the first refused and applies through it, exactly once.

**P3, the empty record.** An absent watermark row reads as zero. The real-Postgres test now starts with an empty record over a fresh database and expects an empty package.

I reverted each of the four fixes in turn and reran; each test failed on its own assertion. Everything else as before: `hale fmt --check .`, `hale verify dna/core`, 23 fixtures, 43 integration tests, the five pinned-fixture gates with no baseline movement, Postgres against a disposable pgvector.

rileyr:
Fifth round addressed at `1e292ae4`.

**1 (P1), the candidate's dependency graph.** Right, and this one is structural rather than a slip: I inspected the graph the destination has today, and the graph that matters is the one the candidate brings. The inputs are now read in the candidate's worktree — `hale inputs` over the tree that will express — and any file of those kinds the destination holds untracked in one of the graph's directories, or that the graph names outright, is dirt. Your reproduction is the fixture's twelfth section: a base with no import, a candidate that adds one, an unfinished `lib/helper.hl` on the destination before preflight, refused by name; take it away and approve again, and it lands.

**3 (P2), manifests.** `hale inputs` now lists the `hale.toml` of every directory in the graph and every C source it declares under `[ffi]`, a declared source whether or not it exists yet, exactly as the native build adds them. Both of your cases are in the fixture: an untracked manifest beside an imported library, and an untracked C source a tracked manifest names.

**4 (P2), a failed inspection.** A graph that cannot be inspected is dirt now, carrying the inspector's exit code and its words. The fixture points a gateway at a toolchain that does not exist and expects a refusal, not a clean tree.

**2 (P2), the CLI after a reapproval.** The record answers a reapproval as it answers a first approval, with a second `review.settled` after the gate has run, so the waiting command hears it. I tested that at the journal, which is what the command keys on, rather than through the command itself; the fixture asserts the second settlement row and its wording.

Two things surfaced on the way. Three organizations in one fixture process share one bus, and a settlement names its review, not its organization, so with colliding mutation ids each assembly acted on the others' settlements. Each organization now mints ids from its own base. And a substrate defect: dissolving an organization whose rehydrated Review holds a keyed subscription, then registering another Review on that key, faults inside keyed dispatch on the next verdict. That is outside DNA and I will file it on its own; the fixture orders its sections so no subscriber registers after a dissolve.

Each of the four fails with its fix reverted. `hale fmt --check .`, `hale verify dna/core`, 23 fixtures, 43 integration tests, the five pinned-fixture gates with no baseline movement.

Issue #596, opened 2026-09-13
DNA Phase 5: the leader


rileyr:
**The picture.** A DNA organism is the nervous system of a software-enabled organization. Its product may be software or may not. Under one organism there are three kinds of thing it can change about itself: the **organism**, its own positions, laws and purpose; **appendages**, software it grows to do its own work; and **products**, software it ships to others. All three are children in the same tower, and they differ by policy, not mechanism: how wide the leader's grant is, who must sign, whether a change ships on approval or waits for a release. A person runs this organization by talking to one position, the leader. They ask for outcomes in their own words. The leader works out what an ask is, hands the work to whoever does that kind of work, and brings back a proposal with evidence. Nothing applies except through the gate, and nothing binding happens without the Board.

## Lineage: brained's Grand Architect

brained separated the **supervisor**, code that routes and delegates, from its **architect**, the prompted operator whose domain is structure. Every supervisor got an architect as standard equipment; the root's was the Grand Architect, paired with the Director, holding the system's own knowledge, proposing topology and knowledge together, never deciding. That separation already exists here and is why the mapping is clean:

| brained | DNA |
|---|---|
| Director | the Board |
| root supervisor (code) | the substrate: assembly, metabolism, gateway |
| Grand Architect | **the leader** — the organism's architect, not its manager |
| system knowledge bound to the architect | the design package bound to `org` (Track C) |
| goals down, concerns up, initiatives bridge; direction is the classification | `classify_binding` — built |
| concern caps and pressure | three concerns become a proposal — built |
| `take_initiative` selector | the pressure and concern path, `initiative: true` — built |
| `optimize` selector | **missing** — Track O below |
| membrane: absorb, scope, propose, diff, discuss, commit | a Mutation of `dna/org` is propose, diff and review by construction; scope is the plan's split-never-widen; absorb is the preprocessor |
| structure follows intent | no organism proposal before purpose, law and knowledge exist — the charter's first rule |
| contract invariance | a position's capabilities are effect contracts; the compiler checks that an internal restructuring leaves the parent's contract alone |
| standard equipment | a department is born with its architect position, bound to the package at its path |

Not copied: the database as the record, hard caps on thoughts, the operations and routing DSL, four fidelity levels, hats. DNA has budgets, grants, reviews and the work system for those.

[...]

PR #597, opened 2026-09-13
dna: the leader's charter, and the design as proposals (GH #596 L and C)

rileyr:
First landing of DNA Phase 5 (#596): Track L's charter and Track C, the design as proposals.

**The charter.** `init` writes `dna/org/charter.hl` beside `purpose.hl`, a function returning text so a change is a reviewed change to the organism. It opens with the sentence the rest of the phase turns on: the leader is the organism's architect, it proposes, the Board decides. Lineage is brained's Grand Architect, which is what the leader has been all along without a brief.

**The design, one Review per practice.** `init` seeds eight practices about how a DNA organization works, adapted from brained's structural domains, as proposals bound to `org`. A Review pins one digest and settles with one outcome, so each practice gets its own; the listing groups them under `design`, and `hale dna review design approve|reject` decides each pending one in turn as a convenience over the individual Reviews. Nothing is ratified by the toolchain.

**Supersession is the Board's decision.** Declining retires nothing, by the projection's existing rule. A proposal that supersedes an earlier digest retires it when ratified, as its own `knowledge.retired` row, and leaves it in force when declined. `upgrade` proposes each practice whose text changed since the latest proposal under its name, and nothing otherwise. Both stores retire by marking not accepted, so a retired idea leaves every package and stays readable.

**Test.** One fixture through a fresh organization: partial acceptance and replay through a fresh service, then earlier versions ratified in-process, an upgrade that supersedes both, a second upgrade that proposes nothing, one replacement ratified and one declined, and the package holding exactly what the Board decided. The seed grew from seven events to twenty-three; every fixture that counted them is updated.

Not in this PR: the leader reading the charter (the next landing, which re-records the trio tape), Track O, Track W, the layering checks.


rileyr:
**Second review pass — the remaining P2 (two pending replacements of one version), fixed here in `00bb90a5`; #598–#600 rebased over it.**

Two guards, so a receipt is never rewritten and the Board never sees a retirement target change under it:

- **The assembly ratifies a replacement only while its predecessor is active.** An approval of a proposal whose `supersedes` names a digest that is already retired is refused in the record — `knowledge.refused <digest>` with `supersedes`, `retired_by`, the review id and the decider — rather than ratified, and the membrane is told. The Board sees a fresh proposal against the active version at the next `upgrade`.
- **`upgrade` proposes one replacement at a time under a name.** While the latest proposal under a practice awaits the Board, a changed text waits too, and `upgrade` says so (`design  N practice(s) changed but wait: an earlier replacement is still before the Board`); a refused proposal is proposed again against what is active now. `DesignState` now carries `pending` and `refused` beside `active`.

Tests: `dna/tests/supersession_test.hl` drives your scenario through the assembly — C active, D and E both superseding C, D approved (C retired), E refused naming D, then E' against D ratified with D retired — and fails with the guard reverted. `dna_design.rs` gains a fourth round: a yet later toolchain over the third round's seven pending replacements proposes one (evolution, whose latest was decided), reports seven waiting, proposes nothing on a second run, and after the Board ratifies the pending principles the next `upgrade` supersedes exactly that digest.

Spec (`spec/dna.md` design bullet) and CHANGELOG updated in the same commit. Heads: #597 `00bb90a5`, #598 `aac7f3fb`, #599 `d20ec66e`, #600 `6c0418ff`. Locally every DNA integration binary (44/45 per branch), the native fixtures and `hale fmt --check` are green on each.

PR #600, opened 2026-09-13
dna: the optimize pass, and a person's job (GH #596 O and W)

rileyr:
Fourth landing of DNA Phase 5 (#596), stacked on #599: the optimize pass, and a person's job.

**The optimize pass (Track O).** On a cadence the org chart sets (`optimize_every_ms` on the substrate, ticked by the org program's loop; 0 is never), the substrate reads the record's structural signals and asks the leader to walk the machinery, not the work. The leader answers with one small proposal or none; the record keeps `org.reviewed` either way, with the signals it read. A proposal enters as an ask in the leader's name and takes the whole road. brained's `optimize` selector, in DNA's terms.

**A person's job (Track W).** A plan of kind `person` hands the Task on: `task.handed`, nothing mutated. The person reports it done with `hale dna task done <id> [--as <who>] [--note …]`, one row in their name. A Task that is not handed is refused.

**Found on the way.** Two organisms on one bus answered each other's plans; the plan exchange is now keyed by `org_id` like the optimize exchange. The fixture runs a clean organism and a proposing one side by side.

**Fixtures.** `dna/tests/optimize_test.hl`: the cadence, a clean pass recorded with its signals and nothing born, a proposal that becomes an ask and is itself planned. `crates/hale-cli/tests/dna_task_done.rs`: an ask planned a person's job through a scripted catalog, handed, reported done in the person's name, refused twice over.

**Tape.** Re-recorded: the org chart gained its tick and the tape's growth entry edits that file.

This closes the tracks of #596. The fresh-model runs that recorded the tape were the dogfood: a live leader read the brief, planned every ask, and reviewed with it.


rileyr:
**Stack review — the four blockers, fixed on their branches**

1. **Organism plan routed under the application's grant and the leader's authority (P1)** — #598, `8a9cc8f1`. The kind now decides whose change it is: kind `organism` is class `organization` whatever the leader wrote — expressed from the organism's seed, assessed under the organism's policy, the Board's to decide — and the `task.planned` row records `class_applied` beside the leader's class. Class `organization` for a child that is not the organism is a contradiction, refused before any editor is placed (`task.refused`, the Work fails, no Mutation). `dna/tests/plan_routing_test.hl` drives an organism plan with class `docs` through the real gate (candidate changes `org/main.hl`, application untouched, `review.requested` carries `seed: org` and `required_authority: board`) and the refused combination; it fails with the fix reverted. While re-recording, the tape showed the live leader had called the application's docs ask an `organism` change — harmless while the class was taken as written, wrong-routed now — because the plan prompt never said what the four kinds are. It does now, and the live leader answers `appendage` for both trio asks.

2. **Optimize pass bypassed budget admission (P1)** — #600, `5e8db7b2`. `optimize()` — the one entry the cadence and a direct call share — asks the budget first; on an exhausted window it journals `optimize.refused <org>` and waits for the next. The fixture runs an organism on an allowance one pass exhausts: the second pass is not taken, the leader is not walked, the exhaustion is announced once. Fails with the check reverted.

3. **Upgrade superseded the latest proposal instead of the active predecessor (P2)** — #597, `c489593f`. `design_in_record` tracks per practice the latest text and the *active* digest (last ratified, not retired); `upgrade` supersedes the active one and dedupes against the latest text. The design test's third round (`HALE_DNA_DESIGN_SUFFIX`) checks that a replacement names the ratified predecessor, not the pending one.

4. **`tick` compared seconds to milliseconds (P2)** — #600, `5e8db7b2`. The generated loop ticks with `std::time::monotonic_ns() / 1000000`; `tick` documents which clock it wants; the init test asserts the generated boundary.

The trio tape is re-recorded (on #598 for the prompt change, on #600 for the generated `main.hl`). Spec and CHANGELOG updated in each commit. #599 rebased over the new #598 with no changes of its own. Locally: all 45 DNA integration tests, the DNA native fixtures, the harness/corpus/topology gates and `hale fmt --check` are green on every branch.

Issue #893, opened 2026-09-20
runtime: a bindings transport is never torn down — its serve thread is never joined and its arena leaks


rileyr:
Split out of GH #871, which fixed eight of the nine fixtures it listed. `85-bindings-unix` is the ninth and has a different root.

## What is live at exit

On an ordinary build, chunk pool on, no sanitizer:

```
$ LOTUS_ARENA_RESIDENCY=1 ./85-bindings-unix/main
listener served its window; exiting
[arena_residency dump] 1 live arenas, sorted by bytes desc:
  [#2 arena=0x... label=__StdBusUnixListenTransport] chunks=0 bytes=0 (0.00 MiB) parent=(nil)
```

and under `LOTUS_NO_CHUNK_POOL=1` on an ASan build, 216 bytes direct (the `lotus_arena_t`) plus a 32-byte indirect. It is the only fixture still on `corpus_oracle.rs`'s `LEAKS_UNMASKED_BY_NO_CHUNK_POOL` list.

## Why

`emit_bindings_prelude` lowers a `bindings { T: unix(...) }` entry as a locus instantiation of `__StdBusUnixListenTransport` / `__StdBusUnixConnectTransport` in `fn main`'s prologue. It does that with `current_user_fn_ret` spoofed to `Some(Some(LocusRef(<transport>)))` — the m90 "this fn returns this locus" signal — because the transport's struct has to outlive `fn main`'s subregion and that flag is what routes it to the program-lifetime payload arena.

The same flag is also what suppresses the teardown. From `lower_locus_instantiation`:

> m90 (3f fix): when this instantiation will escape via fn return (`returns_this_locus` from above), suppress the eager dissolve + arena_destroy and DO NOT push onto the `deferred_dissolves` frame either — the fn-exit flush would otherwise dissolve it on the way out. **The locus leaks (heap allocation + uncleaned arena live until process exit)**; see the alloca branch above for the trade-off note.

So the binding prelude borrows storage from a path that documents its own leak, and inherits it.

## The part that is not just bytes

`__StdBusUnixListenTransport.dissolve()` is not empty:

```hale
dissolve() {
    // Interrupts the serve thread (closing flag + fd
    // shutdown), joins it, destroys the transport.
    std::bus::__transport_reclaim(self.handle);
}
```

[...]

Issue #871, opened 2026-09-20
runtime: nine corpus fixtures leak a locus arena at exit; the chunk pool hid it from LeakSanitizer


rileyr:
Found while building the `LOTUS_NO_CHUNK_POOL` knob for GH #816. Not a regression: these leaks are on `main` today and the chunk pool was hiding them from LeakSanitizer, the same way it was hiding the arena use-after-frees #816 is about.

## The masking mechanism, for leaks

A pooled chunk is a live `malloc` block reachable from `g_chunk_pool`, a `__thread` array, which LeakSanitizer takes as a root. LSan then scans that block's contents — and a recycled chunk still holds the dead arena's bytes, including any pointer to a `lotus_arena_t` that was never destroyed. The never-destroyed arena is therefore "reachable" and not reported. Free the chunks instead of pooling them and the same allocation is a direct leak.

So `LOTUS_NO_CHUNK_POOL=1` (GH #816) turns on leak visibility as well as use-after-free visibility, and nine example fixtures start reporting.

## Independent confirmation, no sanitizer involved

`LOTUS_ARENA_RESIDENCY=1` walks the runtime's own registry of live top-level arenas at exit. On an **ordinary build with the pool ON and nothing from #816 in play**:

```
$ LOTUS_ARENA_RESIDENCY=1 ./60-perspective-slot/main
...
[arena_residency dump] 1 live arenas, sorted by bytes desc:
  [#3 arena=0x62baa9179e40 label=RouterV1] chunks=0 bytes=0 (0.00 MiB) parent=(nil)
```

```
$ LOTUS_ARENA_RESIDENCY=1 ./87-temp-locus-receiver/main
...
[arena_residency dump] 9 live arenas, sorted by bytes desc:
  [#13 arena=... label=Rows]    chunks=1 bytes=65536 parent=(nil)
  [#12 arena=... label=Churner] chunks=1 bytes=65536 parent=(nil)
  [#10 arena=... label=Rows]    chunks=1 bytes=65536 parent=(nil)
  [#9  arena=... label=Churner] chunks=1 bytes=65536 parent=(nil)
  [#8  arena=... label=Queries] chunks=1 bytes=65536 parent=(nil)
  [#7  arena=... label=Rows]    chunks=1 bytes=65536 parent=(nil)
  [#6  arena=... label=Churner] chunks=1 bytes=65536 parent=(nil)
  [#3  arena=... label=Rows]    chunks=1 bytes=65536 parent=(nil)
[...]

Issue #836, opened 2026-09-20
codegen: a factory-built locus in a param field is never dissolved — F.17 excludes it from the frame but the owner never learns it owns it


rileyr:
`Router { quick: make(5), tag: 1 }` — a locus PARAM FIELD initialised from a factory call — never dissolves the child: at exit the router's own `dissolve()` prints, the child's never does, and its arena is not destroyed. The owner's F.29 teardown cascade reaches LITERAL-built children (`quick: Quick { }`), which are parent-owned at instantiation and carry the owner's `__locus_ref_owned_mask` bit; a factory-built child is excluded from the frame by the F.17 gate ("the frame does not reclaim it — the parent does") but the parent never learns it owns it, so nothing does. Identical for `make(5)` and `make(5) or raise` and unchanged by PR #835 (GH #793), which fixed the `let`/argument/temporary dispositions; the F.17 guard test passes with and without that fix because it only asserts "no double dissolve".

Ask: a factory call in a param-field position sets the owner's ownership bit for that field exactly as a literal does (the value's handle is transferred into the field), so the F.29 cascade — now recursive after PR #811 — reaches it; regression: a printing `dissolve()` on the factory-built child fires once at the owner's teardown, in both spellings, and `LOTUS_ASAN=1` + `detect_leaks=1` is clean from a METHOD frame (a `main`-frame test passes vacuously — see #835's note: main's arena teardown hides it). Suggested triage: P2 (a common composition shape leaks its child tree and any descriptor the factory opened). Found while fixing #793.

Issue #383, opened 2026-08-04
Factory-returned @form(vec) loci in method frames never dissolve — pre-existing leak unmasked by the #378 TLS fix (~3.5KB per train_step-shaped call)


rileyr:
Found while verifying #381's fix. The same two-seed shape that produced the zero-reads (a free-fn factory returning a `@form(vec)` locus, consumed inside another locus's method frame) has a second, previously **masked** defect: the factory-returned loci created inside a method frame are **never dissolved**. Deterministic; LeakSanitizer names the factory's `lotus_arena_create_labeled` and its `lotus_vec_push` buffers.

## Repro

Same seed setup as #381 (pond `math/matrix` + `ml/neural`), but with BOTH models in the safe allocation order — so it runs correctly on every compiler version and isolates the leak from the zero-reads bug:

```hale
// probe: two models, both safe order, one train_step each
let xs = mat::from_rows(2, 1, "1.0, 0.0");
let ys = mat::from_rows(1, 1, "1.0");
let m1 = nn::Model { name: "a", rng_state: 42 };
m1.add_dense(2, 3, "sigmoid");
m1.add_dense(3, 1, "sigmoid");
let l1 = m1.train_step(xs, ys, 0.5) or -1.0;
// …identical second block with m2…
```

Build with `LOTUS_ASAN=1`. At current main (`96fca12`): correct output, then at exit:

```
Direct leak of 6048 byte(s) in 28 object(s)
    lotus_arena_create_labeled ← __lib_matrix_matrix_zeros matrix.hl:208
Direct leak of 992 byte(s) in 28 object(s)
    lotus_vec_push ← __lib_matrix_matrix_zeros matrix.hl:212
```

The 28 objects are the Matrix temporaries `train_step` creates through `mat::zeros(...)` (activations/gradients) — factory-returned, let-bound in the method frame, dissolve never runs. ~3.5 KB per `train_step`: a 5000-epoch `fit()` loop leaks unboundedly.

## Why this is unmasked-by-#378, not introduced-by-#378 (ablation matrix)

| build | zero-reads (#381) | this leak |
|---|---|---|
| v0.13.0 release | **red** (5/5) | "clean" |
| 791e25e (post-#373/#374, pre-#378) | **red** (5/5) | "clean" |
| a6ec3a8 (#378) and main | green | **leaks** |
| main with #378 layer 1 only disabled | — | leaks |
| main with layer 2 only disabled | — | leaks |
[...]

PR #916, opened 2026-09-20
codegen: own a factory-built child in an interface-typed field (GH #895)

### LINKED ISSUE #895: codegen: a factory result stored into an interface-typed param field is never reclaimed
Residue of #871 (PR #894) and #836 (PR #852): a FACTORY result stored into an INTERFACE-typed param field still leaks. `field_init_is_fresh_factory` returns false for an `Interface` field by construction (its doc: "an interface-typed field … leaves the bit clear, which is the old leak rather than a double free"), so `Queries { j: make_churner() }` — where `j: Counter` is an interface and `make_churner()` returns a `Churner` — gets no ownership-mask bit and no `__owned_child_reclaim_<f>` write (the mechanism PR #894 added for contract-typed fields), and the child's arena is never destroyed. The machinery to close it now exists: the factory's DECLARED return type names the impl, so the instantiation that owns the field can write `__reclaim_<Impl>` exactly as a literal init does. No corpus fixture has the shape; regression: `Queries { j: make_churner() }` from a METHOD frame under `LOTUS_NO_CHUNK_POOL=1` ASan is clean and `LOTUS_ARENA_RESIDENCY=1` reports 0 live arenas; the `or raise` spelling; an accessor returning a shared handle is NOT reclaimed by the field's owner (control). Suggested triage: P2 (leak on a natural DI shape). Found while fixing #871.

rileyr:
Fixes #895.

**Stacks on #897** (`integration/s12`, batch J) — it carries PR #894's `__owned_child_reclaim_<f>` mechanism and PR #852's `field_init_is_fresh_factory`, which this fix is the last piece of. Base this on main only after #897 lands; the one commit here touches nothing else in that batch.

## The defect

`Queries { j: make_churner() }`, where `j: Counter` is an interface and `make_churner()` is declared `-> Churner`, left the `Churner` — and the `@form(vec)` child under it — alive for the life of the process.

It is the GH #836 transfer written against a CONTRACT-typed field, and it was the one shape in this family that nobody owned. Both halves of the decision were already in place and neither could fire:

* the F.17 gate reads an `Interface` field, so the enclosing frame stood back from the factory's result ("the frame does not reclaim it — the parent does");
* the GH #871 cascade tears down whatever `__owned_child_reclaim_<f>` names;
* between them sits the ownership-mask bit, and `field_init_is_fresh_factory` sets it by comparing the factory's declared locus with the **field's** — which a field typed by a contract does not have. Its own doc said so: "an interface-typed field … leaves the bit clear, which is the old leak rather than a double free."

Frame out, owner blind, child unreclaimed. Measured before the fix:

```
$ LOTUS_ARENA_RESIDENCY=1 ./hale run repro.hl
t=1
end
queries dissolved                      <- the churner's dissolve() never runs
[arena_residency dump] 2 live arenas, sorted by bytes desc:
  [#3 arena=0x… label=Rows] chunks=1 bytes=65536 (0.06 MiB) parent=(nil)
  [#2 arena=0x… label=Churner] chunks=0 bytes=0 (0.00 MiB) parent=(nil)
```

and from a locus METHOD frame under LeakSanitizer with `LOTUS_NO_CHUNK_POOL=1`, four iterations of two such fields: `SUMMARY: AddressSanitizer: 528576 byte(s) leaked in 40 allocation(s)`.

## The fix

The impl's NAME is what closes it, and it is known twice over at the instantiation: the factory **declares** it, and the locus → interface coercion at the field store records what actually reached the slot. When the two agree, the field takes the mask bit and the matching `__reclaim_<Impl>` goes into the reclaim slot — the same store a literal init makes, so the cascade runs the child's whole spine at the owner's timing: drain, dissolve body, capacity slots, arena, to the leaves.

Requiring the *agreement* is what keeps the bit and the reclaim pointer inseparable. A value that reached the field some other way — an already-erased interface handle, a coercion that did not happen — keeps the clear bit it always had, which is the old leak rather than a double free.

Two commits' worth of care in one:

* only a **proven-fresh** factory claims the field (the `fresh_locus_factories` fixpoint GH #383 / #402 / #836 already decide ownership with), in the bare and the diverging-`or` spellings, at the call site and as a param **default**;
* an **accessor** — `Queries { j: pick(x, y, 1) }`, which hands back a locus its own `let` still owns — leaves the bit clear. A missed bit is the old leak; an extra one is a double free;
* a factory returning an impl the param's default does *not* name (`Queries { j: make_tally() }` against `= Churner { }`) writes `__reclaim_Tally`, the #65 ctor-override shape for an interface slot.

`field_init_fresh_factory_impl` is the name half of `field_init_is_fresh_factory`'s yes/no question, written as its own predicate beside it.

### `perspective(P)` is deliberately unchanged

The F.17 gate does not cover a perspective field, so a factory's result there already takes the GH #402 frame temporary and is already reclaimed once. Claiming it for the owner as well would dissolve it twice. Verified both before and after this change: `Gateway { router: make_v2() }` reports zero live arenas and exactly one `dissolve()`.

## Tests, and that they bite

`crates/hale-codegen/tests/owned_child_arena_reclaim.rs` — four, each asserting the exact teardown transcript (a missing teardown is silence; a double one is the same line twice) and zero live arenas:

* both spellings — `make_churner()` and `make_churner_f(1) or raise`;
* the param DEFAULT site (`params { j: Counter = make_churner(); }`), the only way to reach the default-init arm;
* the impl-override variant;
* the accessor control, which passes before and after on purpose.

`crates/hale-codegen/tests/factory_field_ownership.rs` — the same defect in bytes, from a METHOD frame (where it is sanitizer-visible at all: measured in `main` the program reads clean, because main's arena is destroyed at exit), under `ASAN_OPTIONS=detect_leaks=1` and `LOTUS_NO_CHUNK_POOL=1`, through `harness::build_asan` so the instrumentation is checked rather than assumed.

**Proven to bite:** with the `instantiation.rs` hunk reverted and the tests kept, the three residency tests fail with `iface_factory: 4 arena(s) still live at exit`, `iface_factory_default: 2`, `iface_factory_override: 1`, and the ASan test fails with `528576 byte(s) leaked in 40 allocation(s)`. The accessor control passes either way. Restored, all eight in that file and all eight in the other are green.

## Run

`factory_field_ownership`, `owned_child_arena_reclaim`, `factory_returned_binding`, `fresh_temp_attribution`, `locus_field_cascade`, `corpus_oracle` (and `LOTUS_ASAN=1 … --ignored`, 88 fixtures, the one remaining quarantined leak is `85-bindings-unix` / GH #893), the whole of `-p hale-types`, both corpus gates (`topology_projection`, `effects_baseline_gate`), `corpus_check_build_agreement --ignored`, `hale test tests/hale` (35 passed), `hale fmt --check .`. The topology shape baseline gains the two new embedded programs and is regenerated in this commit.

Spec: `spec/semantics.md` F.17/F.29 now says both spellings of construction reach an `interface`-typed param, and that the child reclaimed is the impl the factory built. Book: `docs/src/services/lifecycle.md`.


rileyr:
Integration notes, for whoever batches this:

* Based on `origin/integration/s12` (#897), which carries the `__owned_child_reclaim_<f>` mechanism (#894) and `field_init_is_fresh_factory` (#852) this builds on. `git merge-tree` against `integration/s12` is clean.
* Against **#853** (`fix/or-substitute-field-owner-853`), which touches the same two field-init sites: every source file auto-merges — only `CHANGELOG.md` and `topology_shape_baseline.txt` conflict, both the usual "two entries at the top" / "two added rows" resolutions. I also merged the two locally and ran the combination: `cargo build --release` clean, and `factory_field_ownership`, `owned_child_arena_reclaim`, `factory_returned_binding`, `fresh_temp_attribution`, `locus_field_cascade` all green (36 tests). The two changes are disjoint by field type — #853 arms its rule for `LocusRef` fields, this one for `Interface` fields, and neither reads the other's state.
* Against **#883** (`fix/fresh-temp-per-arm-883`): only `CHANGELOG.md` conflicts; the baseline and both prose files auto-merge.
* `topology_shape_baseline.txt` is regenerated here for the two new embedded programs. If a sibling in the same batch also regenerates it, take both added rows and re-run `HALE_REGEN_TOPOLOGY_BASELINE=1` once at the end — the row for `owned_child_arena_reclaim.rs#8` is a hash of this branch's `FACTORY_TREE` text and will not change.

One shape this deliberately does **not** close, found while reproducing and out of scope here: `Queries { j: make_f(1) or make_churner() }` — an `or <substitute>` whose substitute is a CALL, into an *interface*-typed field. Measured on this branch, the ok branch's value still leaks (one live `Churner` arena; the err branch's substitute is reclaimed). #853 closes that spelling for `LocusRef` fields only — `or_substitute_transfers_into_field` is consulted from the `LocusRef` arm — so the contract-typed twin is left, and it wants the same "the impl's name, agreed by both branches" treatment once both of these land.

A near neighbour that is NOT affected, checked so nobody re-files it: `Queries { j: if flag { make_a() } else { make_b() } }` is clean (zero live arenas, one teardown). The F.17 gate at this site is guarded by `matches!(expr, Expr::Call { .. } | Expr::Or { .. })`, so an `if` / `match` initialiser never reaches `fresh_temp_decision_lands_on` here and keeps its ordinary GH #402 frame temporary.

